PR #4, opened 2024-07-23
feat: add FilterList to TreeDetails page

lfjnascimento:
### Visual reference
image

### DIFF
--- a/dashboard/src/components/FilterList/FilterList.tsx
+++ b/dashboard/src/components/FilterList/FilterList.tsx
@@ -5,16 +5,16 @@ import { useIntl } from 'react-intl';
 
 import { Button } from '../ui/button';
 
-interface IFilterList {
-  itens: string[];
-  onClickItem: (itemIdx: number) => void;
+export interface IFilterList {
+  items: string[];
+  onClickItem: (item: string, itemIdx: number) => void;
   onClickCleanAll: () => void;
   removeOnEmpty?: boolean;
 }
 
 interface IFilterItem extends IFilterButton {
   idx: number;
-  onClickItem: (idx: number) => void;
+  onClickItem: (item: string, itemIdx: number) => void;
 }
 
 export interface IFilterButton
@@ -57,8 +57,8 @@ const FilterItem = ({
   ...props
 }: IFilterItem): JSX.Element => {
   const onClickHandler = useCallback(
-    () => onClickItem(idx),
-    [onClickItem, idx],
+    () => onClickItem(text, idx),
+    [text, onClickItem, idx],
   );
 
   return (
@@ -72,7 +72,7 @@ const FilterItem = ({
 };
 
 const FilterList = ({
-  itens,
+  items,
   onClickItem,
   onClickCleanAll,
   removeOnEmpty = false,
@@ -81,7 +81,7 @@ const FilterList = ({
 
   const buttonList = useMemo(
     () =>
-      itens.map((item, idx) => (
+      items.map((item, idx) => (
         <FilterItem
           key={idx + item}
           text={item}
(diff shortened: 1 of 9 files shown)


dashboard/src/components/FilterList/FilterList.tsx
```diff
@@ -92,7 +92,7 @@ const FilterList = ({
     [itens, onClickItem],
   );
 
-  if (removeOnEmpty && !itens) {
+  if (removeOnEmpty && !itens.length) {
```
mari1912, line 95: nit: items

dashboard/src/routes/TreeDetails/TreeDetailsFilter.tsx
```diff
@@ -1,104 +1,152 @@
...
+  isSelected: boolean,
+): TFilter => {
+  const newFilter = { ...filter };
+  if (newFilter[filterField]) {
+    newFilter[filterField][value] = isSelected;
+  } else {
+    newFilter[filterField] = {
+      [value]: isSelected,
+    };
+  }
```
mari1912, line 79: your call, but what about using ternary operator?

dashboard/src/routes/TreeDetails/TreeDetailsFilter.tsx
```diff
@@ -118,6 +166,8 @@ const TreeDetailsFilter = ({
 
 export default TreeDetailsFilter;
 
+const treeUrl =
+  'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git';
```
mari1912, line 170: this should not be a const, it should change according to the tree you are in

dashboard/src/components/Tabs/Tabs.tsx
```diff
@@ -15,9 +15,14 @@ export interface ITabItem {
 export interface ITabsComponent {
   tabs: ITabItem[];
   defaultTab: ITabItem;
+  FilterListElement?: JSX.Element;
```
mari1912, line 18: nit: filterListElement

dashboard/src/components/Tabs/Tabs.tsx
```diff
@@ -48,6 +53,8 @@ const TabsComponent = ({ defaultTab, tabs }: ITabsComponent): JSX.Element => {
   return (
     <Tabs defaultValue={defaultTab.name} className="w-full">
       <TabsList className="w-full justify-start">{tabsTrigger}</TabsList>
+      <div className="py-6">{FilterListElement}</div>
```
mari1912, line 56: I'm not sure it is worth it for us to pass the whole element instead of a list with the filters and call the necessary ui here

mari1912 (commented):
It is working ok, but I noticed some things that could be improved:

- An error happens if i open the filters model, select one and close the modal without filtering it, when i open it again and select other filter it will add the both of them, even though the first one is not selected
- hover clean all state is too light, we cannot read the text
- about status filter: true or false doesn't make sense
- we should check what this refresh button is suppose to do, I don't think this should be there right now

dashboard/src/routes/TreeDetails/TreeDetailsFilter.tsx
lfjnascimento: I prefer to use ternary only in expressions

dashboard/src/components/Tabs/Tabs.tsx
lfjnascimento: In terms of processing, it's basically the same thing, we are passing a "reference" in both situations, but we gain more flexibility by passing an element here

mari1912 (approved):

PR #49, opened 2024-07-24
Fix: configure cors for local development

lfjnascimento:
- configure cors for local development
-  remove `.env` file and add `.env.example`
- update README


WilsonNet:
It is missing instructions on how  to setup the env variables to run the backend on bare metal

PR #51, opened 2024-07-24
feat/add breadcrumb

### LINKED ISSUE #14: [Tree Monitor] Breadcrumb component

WilsonNet:
This pr achieves the following:
-  make storybook work again**
-  add new figma color**
-  add new breadcrum component **

Closes #14

To test: run storybook

### DIFF
--- a/dashboard/src/components/Breadcrumb/Breadcrumb.tsx
+++ b/dashboard/src/components/Breadcrumb/Breadcrumb.tsx
@@ -1,0 +1,49 @@
+import { ComponentProps } from 'react';
+
+import {
+  Breadcrumb,
+  BreadcrumbList as BreadcrumbListComponent,
+  BreadcrumbItem,
+  BreadcrumbLink as BreadcrumbLinkComponent,
+  BreadcrumbPage,
+  BreadcrumbSeparator as BreadcrumbSeparatorComponent,
+  BreadcrumbEllipsis,
+} from '../ui/breadcrumb';
+
+type BreadcrumbListProps = Omit<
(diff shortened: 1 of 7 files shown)


dashboard/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+type BreadcrumbLinkProps = ComponentPropsWithoutRef<'a'> & {
+  asChild?: boolean;
+};
+const BreadcrumbLink = ({ ...props }: BreadcrumbLinkProps): JSX.Element => {
+  return (
+    <BreadcrumbLinkComponent {...props} className="hover:text-slate-500" />
+  );
+};
```
lfjnascimento, line 25: Nit: I think it would be clearer and more obvious what's going on if we did that, but it is your call

```suggestion
const BreadcrumbLink = ({ ...props }: typeof BreadcrumbLinkComponent): JSX.Element => {
  return (
    <BreadcrumbLinkComponent {...props} className="hover:text-slate-500" />
  );
};
```

lfjnascimento (approved):
Good work :)

dashboard/src/components/Breadcrumb/Breadcrumb.tsx
WilsonNet: improved the type to be
```
type BreadcrumbLinkProps = Omit<
  ComponentProps<typeof BreadcrumbLinkComponent>,
  'className'
>;
```

PR #53, opened 2024-07-24
feat: add buildDetails endpoint

lfjnascimento:
### Description
Add the `/api/build/<build_id>` endpoint which will be used in TreeDetails page

### How to test it

- run Django dev server with:
```sh
python manage.py runserver
```
- and make a request to `/api/build/<build_id>`, such as :
```sh
curl --location 'http://localhost:8000/api/build/0dayci:build:6681e0032e96c85bf5ad3a80'
```

### DIFF
--- a/backend/kernelCI_app/views/__init__.py
+++ b/backend/kernelCI_app/views/__init__.py
@@ -1,0 +1,20 @@
+import os
+import glob
+import importlib
+import sys
+
+
+current_dir = os.path.dirname(__file__)
+modules = glob.glob(os.path.join(current_dir, "*.py"))
+
+for module in modules:
+    module_name = os.path.basename(module)[:-3]
+    if module_name != "__init__":
+        imported_module = importlib.import_module(f'.{module_name}', package=__name__)
+
+        for attribute_name in dir(imported_module):
+            if not attribute_name.startswith('_'):
+                setattr(sys.modules[__name__], attribute_name, getattr(imported_module, attribute_name))
+
+
+del os, glob, importlib, sys, current_dir, modules, module_name, imported_module, attribute_name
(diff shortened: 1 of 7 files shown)


WilsonNet:
Please save the curl request in the project so we can have it documented, maybe a `requests` directory in the root, or inside the endpoint folder

also, I suggest us using `httpie` instead of `curl` for those save requested

backend/kernelCI_app/views.py
```diff
@@ -141,3 +141,28 @@ def get(self, request, commit_hash):
         summary = self.create_summary(records)
 
         return JsonResponse({"builds": records, "summary": summary}, safe=False)
+
+
+class BuildDetails(View):
```
WilsonNet, line 146: I think we should extract this to a different files otherwise this will become a clusterfuck

WilsonNet:
here is how the request wold be done with `httpie`

`http 'http://localhost:8000/api/build/0dayci:build:6681e0032e96c85bf5ad3a80'`

image

backend/kernelCI_app/views/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    module_name = os.path.basename(module)[:-3]
+    if module_name != "__init__":
+        imported_module = importlib.import_module(f'.{module_name}', package=__name__)
+
+        for attribute_name in dir(imported_module):
+            if not attribute_name.startswith('_'):
+                setattr(sys.modules[__name__], attribute_name, getattr(imported_module, attribute_name))
+
+
+del os, glob, importlib, sys, current_dir, modules, module_name, imported_module, attribute_name
```
WilsonNet, line 20: just out of curiosity, why are you deleting variables here since python is garbage collected (no necessity to remove this)

WilsonNet: oh got it, it is to not conflict with the api endpoint

WilsonNet (approved):
so easy to test now

image

PR #59, opened 2024-07-29
feat(settings): enhance security and deployment settings

WilsonNet:
This PR introduces several changes to improve the security and
deployment settings of the application. The `SECRET_KEY` is now fetched
from environment variables, and the `DEBUG` setting is set to `False` by
default, but can be overridden by an environment variable.

Additionally, several security-related settings have been added, such as
`SESSION_COOKIE_SECURE`, `CSRF_COOKIE_SECURE`, `SECURE_SSL_REDIRECT`,
and `SECURE_HSTS_SECONDS`.

The README.md file has also been updated with instructions on how to
generate a `DJANGO_SECRET_KEY` and check if the application is ready for
deployment.

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/5.0/ref/settings/
 """
 
 from pathlib import Path
+from utils.validation import isBooleanOrStringTrue
 import os
 import json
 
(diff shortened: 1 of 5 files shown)


backend/kernelCI/settings.py
```diff
@@ -35,14 +35,28 @@ def get_json_env_var(name, default):
...
+    "DJANGO_SECRET_KEY", "django-insecure--!70an0r@i00)oqf!3uq_)9dx2^%)"
+)
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = get_json_env_var('DEBUG', True)
+DEBUG = False
+
+ENV_DEBUG = get_json_env_var("DEBUG", False)
+
+if ENV_DEBUG is True or ENV_DEBUG == "True" or ENV_DEBUG == "true":
```
lfjnascimento, line 47: It would be nice to create a `isTruly` function to reuse this logic when we need it.

lfjnascimento (approved):
Nice!

PR #60, opened 2024-07-29
docs/database logic

WilsonNet:
This commit introduces a new documentation file that explains the logic
behind the database fields. It provides details about the Tree, Checkouts,
Builds, and Test tables. It also highlights important columns like `path`,
`start_time`, `status`, `misc`, and `environment_misc` in the Test table.

### DIFF
--- a/backend/docs/database-logic.md
+++ b/backend/docs/database-logic.md
@@ -1,0 +1,53 @@
+# Database Logic
+
+In this we translate the logic of the database fields and add details that we should know
+
+
+## Tree
+A tree has no respective table, but it is a collection of checkouts with the same `git_commit_hash` and `patchset_hash`, but pretty much nobody uses `patchset_hash` nowadays, so it will probably be `null` in most cases
+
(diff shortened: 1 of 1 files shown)


backend/docs/database-logic.md
```diff
@@ -0,0 +1,53 @@
...
+{
+  "arch": "arm64",
+  "compiler": "gcc-12",
+  "error_msg": "Invalid TESTCASE signal",
+  "error_code": "Test",
+  "kernel_type": "image"
+}
+```
+
+- `environtment_misc` is a field called environment misc which has a JSON and we can use for things about the environment like detecting the platform, this is not set in stone and can change,  so we should double check the validation
```
lfjnascimento, line 46: - too redundant

```suggestion
- `environtment_misc` a JSON field that we can use for things about the environment like detecting the platform, this is not set in stone and can change,  so we should double check the validation
```

lfjnascimento (approved):

PR #67, opened 2024-07-31
feat: update database-logic.md doc

### LINKED ISSUE #64: [Docs] Update data documentation considering Nikolai review

lfjnascimento:
- update database-logic.md doc

close #64

### DIFF
--- a/backend/docs/database-logic.md
+++ b/backend/docs/database-logic.md
@@ -1,39 +1,54 @@
 # Database Logic
 
-In this we translate the logic of the database fields and add details that we should know
+In this we translate the logic of the database fields and add details
+that we should know
 
+
+There are some resources that can be used as reference:
+- [Grafana Dashboard](https://kcidb.kernelci.org/d/home/home?orgId=1&refresh=30m)
+- [Submitter guide](https://docs.kernelci.org/kcidb/submitter_guide/)
+- [kcidb json schema](https://github.com/kernelci/kcidb-io/blob/main/kcidb_io/schema/v04_03.py)
 
 ## Tree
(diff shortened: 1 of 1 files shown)


backend/docs/database-logic.md
```diff
@@ -1,53 +1,58 @@
...
-  "platform": "hp-x360-14a-cb0001xx-zork"
-}
-```
\ No newline at end of file
+- `status` Column show a status that can be `PASS` `SKIP` or `ERROR`
+
+- `misc` The contents of the `misc` field is not standardized, and shouldn't
+ be relied upon to ever be used consistently by more than one
+ CI system ("origin"), or even by the originating CI system itself, over time.
+The most it could be used for is a proof-of-concept.
```
WilsonNet, line 52: I think we should still add the misc fields that we have found because it is the only thing that we have right now, we just need to add a not

> Like this
> note

lfjnascimento: done

WilsonNet:
I believe we should also add the definitions for Revision since it is closer to what we have been using for the time

backend/docs/database-logic.md
```diff
@@ -43,11 +58,24 @@ eg:
...
+>NOTE: The contents of the `misc` field is not standardized, and shouldn't
+>be relied upon to ever be used consistently by more than one
+>CI system ("origin"), or even by the originating CI system itself, over time.
+>The most it could be used for is a proof-of-concept.
+
+
+## Revision
+
+Revision has no respective table, but it is a collection of checkouts with
+the same `git_commit_hash` and `patchset_hash`.
```
WilsonNet, line 81: I think is important to add that `patchset_hash` is not commonly used anymore

WilsonNet (approved):
looks good

PR #73, opened 2024-08-02
feat/breadcrumb dynamic

WilsonNet:
- **feat(routes): add main route to tree**

- Imported `useParams` in `BuildAccordionContent.tsx` to get `treeId`
- Updated `navigateToBuildDetails` to include `treeId` in state
- Fixed navigation path in `TreeTable.tsx`
- Added breadcrumb navigation in `BuildDetails.tsx` using `treeId`
- Created `isString` utility function to check if value is a string


lfjnascimento (approved):
It is working

PR #90, opened 2024-08-07
refactor: drop hardcoded config and use revision instead of tree

### LINKED ISSUE #8: Drop trees configuration and show all trees in Trees Listing Page
Merge multiple checkouts by commit. And display them as a single tree.
Notice that checkouts are CI system dependent, so multiple checkouts can refer to the same commit. By merging them we show data from multiple CI systems in a single view. Each CI system is identified by the origin field.

lfjnascimento:
Close #8


WilsonNet (approved):
It is working here, maybe we need to limit this query and show patchset_hash on the table

lfjnascimento:
> It is working here, maybe we need to limit this query and show patchset_hash on the table

Patchset hash added to the table and I had already created a [ticket](https://github.com/orgs/kernelci/projects/28/views/3?pane=issue&itemId=73909542) to investigate the idea of ​​limiting the query, we will probably add pagination

Issue #69, opened 2024-08-01
Remove `hadcorded` Tree and `Commit/tag` from filter drawer


bdilly:
Probably needs to be done after #8

PR #99, opened 2024-08-13
refactor: remove hardcoded Tree and Commit from filter drawer

lfjnascimento:
also:
- rename SummarySectioncolumn to SummarySectionColumn
- move the i18n message 'treeTable.commit' to 'global.commit'

Close: #69


WilsonNet (approved):
looks good

PR #107, opened 2024-08-14
fix(#20): set the start index correctly when some filter is applied

### LINKED ISSUE #20: [Bug] [Builds Tab] Itens navigation breaks after filters change

mari1912:
- closes #20

### DIFF
--- a/dashboard/src/hooks/usePagination.tsx
+++ b/dashboard/src/hooks/usePagination.tsx
@@ -1,8 +1,12 @@
 import { useState, useEffect, useCallback } from 'react';
 
+import { TFilter, TableFilter } from '@/types/tree/TreeDetails';
+
 export const usePagination = (
   totalItems: number,
   itemsPerPage: number,
+  filters?: TFilter,
+  tableFilter?: TableFilter,
 ): {
   startIndex: number;
   endIndex: number;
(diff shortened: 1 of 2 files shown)


dashboard/src/hooks/usePagination.tsx
```diff
@@ -13,6 +13,7 @@ export const usePagination = (
   const [endIndex, setEndIndex] = useState(0);
 
   useEffect(() => {
+    setStartIndex(0); //restart the index whenever there is a change in filters
```
WilsonNet, line 16: this won't work if the itemsPorPage and totalItems have the same count, needs to add `diffFilter` and `filter` to the dependency array

WilsonNet: you will need `useSearch` for this one

mari1912: Can you give me an example when it's possible? I tried with http://localhost:5173/tree/0bdf955f408e17b64d9dc4de09d3a2fc4f233aca?tableFilter=all&currentTreeDetailsTab=treeDetails.builds&diffFilter=%7B%7D changing the itemsPerPage and it doesn't seem to be broken

WilsonNet: it will be a rare case if not never, but it is possible since are primitives in javascript and the `useEffect` wont detect a change if they are the same but coming from another source

WilsonNet (approved):
it's also missing `tableFilter` which  also changes the table

WilsonNet:
I just realized that there is a regression and tableFilter is not being used anymore when changing the tabel filter

Issue #78, opened 2024-08-05
Add navbar with origin selection


bdilly:
Before implementing talk to Gus to add more links
Origins:
"0dayci"
"broonie"
"kernelci"
"maestro"
"microsoft"
"redhat"
"syzbot"
"tuxsuite"

PR #116, opened 2024-08-15
feat: add origin selector

lfjnascimento:
- add Select component from shadcn
- add Select component which is a custom version of ui/Select
- add origin filter to tree endpoint
- add origin selector to top bar

image

close: #78

### DIFF
--- a/dashboard/src/components/Select/Select.tsx
+++ b/dashboard/src/components/Select/Select.tsx
@@ -12,7 +12,7 @@ const Select = (props: React.ComponentProps<typeof SelectUI>): JSX.Element => {
   return (
     <SelectUI {...propsWithoutChildren}>
       <SelectTrigger className="w-auto rounded-full border-2 border-dimGray px-6 py-4 text-base font-medium text-dimGray">
-        <SelectValue placeholder="Theme" />
+        <SelectValue placeholder="" />
       </SelectTrigger>
       <SelectContent>{children}</SelectContent>
     </SelectUI>
--- a/dashboard/src/components/TopBar/TopBar.tsx
+++ b/dashboard/src/components/TopBar/TopBar.tsx
@@ -1,12 +1,64 @@
 import { FormattedMessage } from 'react-intl';
 
+import { useSearch, useNavigate } from '@tanstack/react-router';
+import { useCallback, useEffect, useMemo } from 'react';
+
+import Select, { SelectItem } from '@/components/Select/Select';
+import { zOrigin, zOriginEnum } from '@/types/tree/Tree';
+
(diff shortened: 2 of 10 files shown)


dashboard/src/components/Select/Select.tsx
```diff
@@ -0,0 +1,23 @@
...
+  SelectValue,
+} from '@/components/ui/select';
+
+const Select = (props: React.ComponentProps<typeof SelectUI>): JSX.Element => {
+  const { children, ...propsWithoutChildren } = props;
+
+  return (
+    <SelectUI {...propsWithoutChildren}>
+      <SelectTrigger className="w-auto rounded-full border-2 border-dimGray px-6 py-4 text-base font-medium text-dimGray">
+        <SelectValue placeholder="Theme" />
```
mari1912, line 15: should we add a better placeholder and make it use i18n?

dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -1,12 +1,62 @@
...
+  useEffect(() => {
+    if (unsafeOrigin === undefined)
+      navigate({
+        search: previousSearch => ({ ...previousSearch, origin: origin }),
+      });
+  });
+
+  return (
+    <div className="flex items-center">
+      <span className="mr-4 text-base font-medium text-dimGray">Origin</span>
```
mari1912, line 44: use intl

dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -1,12 +1,62 @@
...
+  );
+};
+
 const TopBar = (): JSX.Element => {
   return (
     <div className="fixed top-0 z-10 mx-52 flex h-20 w-full bg-white pl-6 pr-12">
-      <div className="flex w-full flex-row items-center justify-between">
-        <span className="text-2xl">
+      <div className="flex flex-row items-center justify-between">
+        <span className="mr-14 text-2xl">
```
mari1912, line 56: just make sure it will be aligned with other content, because of the changes we are making

mari1912 (commented):
It's working nice. I just want to make sure that it is expected to be able to change the origin in any place and when we do it, go back to the revisions initial page, right?

dashboard/src/components/Select/Select.tsx
lfjnascimento: actually this shouldn't be here, i was using it for testing purposes. Nice catch

mari1912: can we remove it?

mari1912 (approved):

PR #185, opened 2024-08-26
fix(dashboard): enabling "Trees" link on side menu

### LINKED ISSUE #120: Clicking on Revisions/Trees on Side bar or Nav bar don't take us the the Trees page
It seems we didn't connect the links there yet.

lucasbracher:
Allows "Trees" link on side menu, redirecting it to '/?origin=<origin>'

function useNavigateTo receives one parameter (path) and adds GET "origin" parameter to URL.

Closes #120


mari1912 (approved):
nice job!

WilsonNet (changes requested):
please use tanstack router for routing usage

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -65,6 +48,43 @@ const SideMenu = (): JSX.Element => {
   const notSelectedItemClassName =
     'w-full flex pl-5 py-4 cursor-pointer text-white';
 
+  const navigate = useNavigate();
+
+  const useNavigateTo = (path: string): (() => void) =>
+    useCallback(() => {
+      const queryParams = new URLSearchParams(location.search);
```
WilsonNet, line 55: tanstack router is type safe so please use it to get the search parameters

https://tanstack.com/router/latest/docs/framework/react/guide/search-params

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -65,6 +46,45 @@ const SideMenu = (): JSX.Element => {
   const notSelectedItemClassName =
     'w-full flex pl-5 py-4 cursor-pointer text-white';
 
+  const useNavigateTo = (path: string): (() => void) => {
+    const router = useRouter();
+    const searchParams = useSearch({ strict: false });
+    const org = searchParams.origin;
```
WilsonNet, line 52: using `{ stict: false }` means that it is unsafe and it needs to be validated, otherwise it can break the entire app, you can do something like 

const { origin: unsafeOrigin } = useSearch({ strict: false })
and validate it using zod

`const origin = zOrigin.parse(unsafeOrigin)` 

`zOrigin` is defined in the types folder

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -65,6 +46,45 @@ const SideMenu = (): JSX.Element => {
...
+
+    let finalPath = path;
+    if (!finalPath.endsWith('/')) {
+      finalPath = finalPath + '/';
+    }
+
+    return () => {
+      const newPath = `${finalPath}?origin=${org}`;
```
WilsonNet, line 60: please use big variable names, org is probably used more as an abbreviation to organization than to origin

WilsonNet (approved):
nice

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -4,8 +4,12 @@ import { ImTree, ImImages } from 'react-icons/im';
 
 import { FormattedMessage } from 'react-intl';
 
+import { useRouter, useSearch } from '@tanstack/react-router';
+
 import { MessagesKey } from '@/locales/messages';
 
+import { zOrigin } from '../../types/tree/Tree';
```
WilsonNet, line 11: Please avoid `../../` we still need to add the eslint rule, use `@/types`

PR #196, opened 2024-08-27
fix: route hierarchy

### LINKED ISSUE #165: bug: tree information disapear after clicking in trees
if you select a tree, go to test details or build details page and click in `Tree Details` in breadcrumb, the tree information will not be shown in the header

### LINKED ISSUE #191: Navigation from Tree Details Page to Test Details Page passes wrong params (e.g. origin)

lfjnascimento:
Using just `/<route>/index.tsx` in TanStack router is not enough to create a route,
that's why we were with just one route (root route) and every page was its child,
this commit fix that and create the needed router hierarchy

also:

- refactor: remove the need for `~` in the route files name
- fix: search params not being passed on navigation

Close #165 
Close #191


dashboard/src/routeTree.gen.ts
```diff
@@ -11,38 +11,62 @@
...
+  path: '/',
+  getParentRoute: () => TreeTreeIdRouteRoute,
+} as any)
+
+const TreeTreeIdTestRouteRoute = TreeTreeIdTestRouteImport.update({
+  path: '/test',
+  getParentRoute: () => TreeTreeIdRouteRoute,
+} as any)
+
+const TreeTreeIdTestTestIdRouteRoute = TreeTreeIdTestTestIdRouteImport.update({
```
mari1912, line 56: nit: is there any way we can improve those variables names to avoid this much repetition?

mari1912 (approved):
works nice

dashboard/src/routeTree.gen.ts
lfjnascimento: This is a generated file, so we can't, but we also don't need to, as we won't be manipulating this code.

PR #202, opened 2024-08-28
feat(dashboard): loading message for tree listing and builds tab

### LINKED ISSUE #172: Create loading for Tree Listing Page and Builds Tab

lucasbracher:
Closes #172

### DIFF
--- a/dashboard/src/components/Table/TreeTable.tsx
+++ b/dashboard/src/components/Table/TreeTable.tsx
@@ -13,9 +13,9 @@ import {
   possibleTestsTableFilter,
 } from '@/types/tree/TreeDetails';
 
-import { TableRow, TableCell } from '../ui/table';
+import { TreeTableBody, zOrigin } from '@/types/tree/Tree';
 
-import { TreeTableBody, zOrigin } from '../../types/tree/Tree';
+import { TableRow, TableCell } from '@/components/ui/table';
 
 import BaseTable from './BaseTable';
 import { sanitizeTableValue } from './tableUtils';
--- a/dashboard/src/components/TreeListingPage/TreeListingPage.tsx
+++ b/dashboard/src/components/TreeListingPage/TreeListingPage.tsx
@@ -2,9 +2,13 @@ import { useSearch } from '@tanstack/react-router';
 
 import { useMemo } from 'react';
 
+import { FormattedMessage } from 'react-intl';
+
 import { zOrigin } from '@/types/tree/Tree';
 
 import { usePagination } from '@/hooks/usePagination';
+
+import { Skeleton } from '@/components/Skeleton';
 
 import TreeTable from '../Table/TreeTable';
 import { TreeTableBody } from '../../types/tree/Tree';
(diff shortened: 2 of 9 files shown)


dashboard/src/components/Table/TreeTable.tsx
```diff
@@ -13,9 +13,9 @@ import {
   possibleTestsTableFilter,
 } from '@/types/tree/TreeDetails';
 
-import { TableRow, TableCell } from '../ui/table';
+import { TreeTableBody, zOrigin } from '@/types/tree/Tree';
 
-import { TreeTableBody, zOrigin } from '../../types/tree/Tree';
+import { TableRow, TableCell } from '../ui/table';
```
WilsonNet, line 18: since you are changing here, can you update for `@` import?
the reason we are allowing `..` in this project is because we didn't have the plugin for @ imports on vite

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -79,6 +83,13 @@ const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
     return <div>Error: {error.message}</div>;
   }
 
+  if (isLoading)
+    return (
+      <Skeleton className="grid h-[400px] place-items-center">
```
WilsonNet, line 88: since all the Skeletons are using the same classes, you could add this to a component

the way we are doing this is that components from the `ui` folder are straight from `shadcn` 
we don't wan't to touch those, we reimport in another folder and add the classes

something like
```typescript
import type { ReactNode } from 'react'
import { Skeleton as SkeletonComponent} from '@/components/ui/skeleton'; //to avoid name conflict we rename the import

type SkeletonProps = {
  children: ReactNode
}
const Skeleton = ({ children }: SkeletonProps) =>  {
<SkeletonComponent className="grid h-[400px] place-items-center"> 
  {children}
</SkeletonComponent>
}

export default Skeleton // default exports are terrible but we are using here because every project use them for components, feel free to not use it
```

and now to use you just import Skeleton from this file and wrap without the classes

dashboard/src/components/ui/skeletonCenter.tsx
```diff
@@ -0,0 +1,16 @@
+import type { ReactNode } from 'react';
+import { Skeleton as SkeletonComponent } from '@/components/ui/skeleton'; //to avoid name conflict we rename the import
+
+type SkeletonProps = {
+  children: ReactNode;
+};
+
+const Skeleton = ({ children }: SkeletonProps) => {
```
WilsonNet, line 8: great job here, two things

the `ui` folder is used just for components that have code generated from `shadcn`, move this one level up to the components folder

also, react components file use PascalCase name, it shuold be `SkeletonCenter`, and it is a common lint rule to have the file name the same as the default export, so the file could be named `Skeleton.tsx`

take a look at the `components` folder and see if you see a pattern there

WilsonNet (approved):
nice job

PR #206, opened 2024-08-29
feat(dashboard): empty state for tree listing

### LINKED ISSUE #173: Create empty state for Tree Listing Page

lucasbracher:
Closes #173

### DIFF
--- a/dashboard/src/components/Skeleton/Skeleton.tsx
+++ b/dashboard/src/components/Skeleton/Skeleton.tsx
@@ -1,14 +1,18 @@
-import type { ReactNode } from 'react';
+import type { ReactNode, ComponentProps } from 'react';
 
 import { Skeleton as SkeletonComponent } from '@/components/ui/skeleton'; //to avoid name conflict we rename the import
 
-type SkeletonProps = {
+import { cn } from '@/lib/utils';
+
+type SkeletonProps = ComponentProps<typeof SkeletonComponent> & {
   children: ReactNode;
 };
 
-const Skeleton = ({ children }: SkeletonProps): JSX.Element => {
+const Skeleton = ({ children, className }: SkeletonProps): JSX.Element => {
   return (
-    <SkeletonComponent className="grid h-[400px] place-items-center">
+    <SkeletonComponent
+      className={cn('grid h-[400px] place-items-center', className)}
+    >
       {children}
     </SkeletonComponent>
   );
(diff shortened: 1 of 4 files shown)


dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -114,6 +114,10 @@ const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
         />
       </div>
     </div>
+  ) : (
+    <div className="grid h-[400px] place-items-center rounded-md bg-slate-100 dark:bg-slate-800">
```
mari1912, line 118: lets avoid using random colors, if necessary, add some of them to the tailwind.config file, but let's going to use only colors that are there

dashboard/src/components/Skeleton/Skeleton.tsx
```diff
@@ -1,14 +1,16 @@
...
+type SkeletonProps = ComponentProps<typeof SkeletonComponent> & {
   children: ReactNode;
 };
 
-const Skeleton = ({ children }: SkeletonProps): JSX.Element => {
+const Skeleton = ({ children, className }: SkeletonProps): JSX.Element => {
   return (
-    <SkeletonComponent className="grid h-[400px] place-items-center">
+    <SkeletonComponent
+      className={`grid h-[400px] place-items-center ${className}`}
```
mari1912, line 12: we are using a library `classNames` that helps us to concat different classNames without needing to use `$something`. here is an example, but you can search for more in the code

```javascript
className={classNames('flex flex-row gap-2 pb-1', hasBorder)}
```

dashboard/src/index.css
```diff
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+.remove-animate-pulse.animate-pulse {
+  animation: none !important;
+}
```
mari1912, line 7: why are you adding this?

lucasbracher: Because `animate-pulse` is in the original component, so I created this class that cancels out the previous one.

dashboard/src/components/Skeleton/Skeleton.tsx
WilsonNet: don't use classnames, use our `cn` library

mari1912 (approved):

PR #231, opened 2024-09-02
feat(dashboard): accordion chevron turning when clicked

### LINKED ISSUE #208: change the > in the accordion to an arrow down when opened
after open the accordion, change the icon in the trigger

lucasbracher:
Implements component ChevronRightAnimate, that detects click. 
Piggybacking: putting Skeleton css class inside its folder.

Closes #208

### DIFF
--- a/dashboard/src/components/Skeleton/Skeleton.tsx
+++ b/dashboard/src/components/Skeleton/Skeleton.tsx
@@ -11,7 +11,10 @@ type SkeletonProps = ComponentProps<typeof SkeletonComponent> & {
 const Skeleton = ({ children, className }: SkeletonProps): JSX.Element => {
   return (
     <SkeletonComponent
-      className={cn('grid h-[400px] place-items-center', className)}
+      className={cn(
+        'animation-none grid h-[400px] place-items-center',
+        className,
+      )}
     >
       {children}
     </SkeletonComponent>
--- a/dashboard/src/components/Accordion/Accordion.tsx
+++ b/dashboard/src/components/Accordion/Accordion.tsx
@@ -1,6 +1,6 @@
 import { ReactElement, useCallback, useMemo } from 'react';
 
-import { MdCheck, MdChevronRight, MdClose } from 'react-icons/md';
+import { MdCheck, MdClose, MdChevronRight } from 'react-icons/md';
 
 import { FormattedMessage } from 'react-intl';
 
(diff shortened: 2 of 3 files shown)


dashboard/src/components/ChevronRightAnimate/ChevronRightAnimate.tsx
```diff
@@ -0,0 +1,21 @@
+import { useState } from 'react';
+import { MdChevronRight } from 'react-icons/md';
+import './styles.css';
+
+const ChevronRightAnimate = (): JSX.Element => {
+  const [isRotated, setIsRotated] = useState(false);
```
WilsonNet, line 6: we don't need javascript for this, this can be done in CSS

dashboard/src/components/ChevronRightAnimate/styles.css
```diff
@@ -0,0 +1,7 @@
+.chevron {
```
WilsonNet, line 1: we are using tailwind here, css is not scoped, every css here would be global

dashboard/src/components/Skeleton/styles.css
```diff
@@ -0,0 +1,5 @@
+.remove-animate-pulse.animate-pulse {
```
WilsonNet, line 1: please restrain from use css files

dashboard/src/components/Skeleton/styles.css
```diff
@@ -0,0 +1,5 @@
...
+  animation: none !important;
```
WilsonNet, line 2: never use `!important` unless you have less than 30 minutes for the deadline of a task, or there is no other way around

dashboard/src/components/Skeleton/Skeleton.tsx
```diff
@@ -11,7 +11,10 @@ type SkeletonProps = ComponentProps<typeof SkeletonComponent> & {
 const Skeleton = ({ children, className }: SkeletonProps): JSX.Element => {
   return (
     <SkeletonComponent
-      className={cn('grid h-[400px] place-items-center', className)}
+      className={cn(
+        'animation: none grid h-[400px] place-items-center',
```
WilsonNet, line 15: this sohuld be `animation-none`

https://tailwindcss.com/docs/animation

dashboard/src/components/ChevronRightAnimate/ChevronRightAnimate.tsx
```diff
@@ -0,0 +1,9 @@
+import { MdChevronRight } from 'react-icons/md';
+
+const ChevronRightAnimate = (): JSX.Element => {
```
WilsonNet, line 3: Because of the `group`,  this component only makes sense inside the `dashboard/src/components/Accordion/Accordion.tsx`.

This won't be animated anywhere else in the code. 

So, just move there, no need export it or create these new files.

WilsonNet (approved):

PR #236, opened 2024-09-04
feat(commit-navigation): add boots and tests

WilsonNet:
# Description
This PR adds the commit navigation graph to boots and tests.
- Modify the endpoint to add the status count for Boots and Tests
- Add a single component for commit graph that verifies the URL 
It also adds:
- QuerySwitcher component to avoid the need of checking react query status everytime
- Improve the responsiviness in boots and query tabs

Closes #151

PR #237, opened 2024-09-04
feat: add duration filters for test and boots

lfjnascimento:
- create `getTargetFilter` function that sanitizes the filters and
select only the ones that matter for a give enpoint. This helps the
backend that will not have to deal with a bunch of useless filters
and also improve the query keys on the frontend that has become more
granular watch only the necessary filters which avoid unnecessary
requests

- feat: make TreeTestView accept more filters besides status
- feat: add duration filters for test and boots


dashboard/src/pages/TreeDetails/TreeDetailsFilter.tsx
```diff
@@ -39,14 +40,18 @@ interface ITreeDetailsFilter {
...
+  'treeDetails.valid': 'buildStatus',
+  'treeDetails.duration_[gte]': 'buildDurationMin',
+  'treeDetails.duration_[lte]': 'buildDurationMax',
+  'boot.status': 'bootStatus',
+  'boot.duration_[gte]': 'bootDurationMin',
+  'boot.duration_[lte]': 'bootDurationMax',
+  'test.status': 'testStatus',
+  'test.duration_[gte]': 'testDurationMin',
+  'test.duration_[lte]': 'testDurationMax',
 } as const satisfies Record<string, TFilterKeys>;
```
WilsonNet, line 55: the string here in the Record< seems a little weird, I don't know exatcly where this comes from but this could be a little more specific

backend/kernelCI_app/views/treeTestsView.py
```diff
@@ -44,19 +44,19 @@ def __paramsFromBootAndFilters(self, commit_hash, path_param, filter_params, par
...
-            path_filter += "AND t.status IN (" + placeholders + ")"
-            for value in value_list:
+            value_is_list = isinstance(value, list)
+            if value_is_list:
+                placeholder = ",".join(['%s' for i in value]) if value_is_list else '%s'
+                path_filter += f" AND t.{f['field']} IN ({placeholder})"
+                params.extend(value)
+            else:
+                comparison_op = filter_params.get_comparison_op(f, 'raw')
+                path_filter += f" AND t.{f['field']} {comparison_op} %s"
```
WilsonNet, line 59: Please add a TODO  to swap those string interpolation with parametrized queries since this doesn't get sanitized and it is a vulnerability

loadez: I would say that we should sanitize the input instead of adding a TODO

WilsonNet: I agree, it is fairly easy, just use a dictionary as a parameter and use `%(key)s` structure like I did in my PR, instead of an `f-string`

https://github.com/kernelci/dashboard/pull/236/files

lfjnascimento: The thing is, I also need to pass the parameters as a list here and we cant do booth, I look for alternatives

lfjnascimento: I cant use a dict and a list as parameters at the same time, i will look for an alternative...

lfjnascimento: I can check the ['field'] to make sure it is valid

loadez (approved):

PR #246, opened 2024-09-05
feat(dashboard): `maestro` is default origin

### LINKED ISSUE #239: Make maestro the default origin
As per the default origin 0-day is the default page, but it has only 2 trees show. OTOH, maestro has a bunch of tree and a few active engagement with upstream maintainers to look at the data, so we should make it the default one for now.

lucasbracher:
Makes `maestro` appear as default origin when it's not at the query string at the URL.

Closes #239


dashboard/src/types/tree/Tree.tsx
```diff
@@ -77,4 +77,4 @@ const origins = [
 ] as const;
 
 export const zOriginEnum = z.enum(origins);
-export const zOrigin = zOriginEnum.catch(origins[0]);
+export const zOrigin = zOriginEnum.catch(origins[2]);
```
lfjnascimento, line 80: it would be better to use the literal `maestro` here, because if we add more items to the list the index will be affected, another option is to create a `const DEFAULT_ORIGIN = 'maestro'`

lfjnascimento (approved):

dashboard/src/types/tree/Tree.tsx
WilsonNet: yeah, typescript will typeguard to have only the right strings to work, so you don't need to worry typing an origin that doesn't exist here

Issue #233, opened 2024-09-02
Tree Details Page [execution list]: add hover in status column in the table header


joaobertacchi:
Add a (?) question mark right to the Status label in the execution list for Boots and Tests tabs.

Message for the Builds List table:
```
Success - builds completed successfully
Failed - builds failed
Inconclusive - builds with unknown status, including ongoing builds that didn't finish yet.
```

Message for the Boots List table:
```
Success - boots with PASS status
Failed - boots with FAIL status
Inconclusive - boot concluded with inconclusive results such as infrastructure errors.

Inconclusive groups boots with ERROR, MISS, SKIP, DONE and unknown statuses defined by KCIDB.
```

Message for the Tests List table:
```
Success - tests with PASS status
Failed - tests with FAIL status
Inconclusive - test concluded with inconclusive results such as infrastructure errors.

Inconclusive groups tests with ERROR, MISS, SKIP, DONE and unknown statuses defined by KCIDB.
```

tales-aparecida:
FYI Tests might also have unknown (null) status

padovan:
For builds:

```
Success - builds completed successfully
Failed - builds failed
Inconclusive - builds with unknown status, including ongoing builds that didn't finish yet.
```

For boots
``` 
Success - boots with PASS status
Failed - boots with FAIL status
Inconclusive - boot test concluded with inconclusive results such as infrastructure errors.

Inconclusive groups boots with ERROR, MISS, SKIP and DONE statuses defined by KCIDB.
```

For tests I'd add do the same as I did for boots.

PR #248, opened 2024-09-06
feat: show tooltip when hover on Status column

lfjnascimento:
- add HeaderWithInfo component 
- add tooltip on Status header in `Build`, `Boot`, `Test` tab

Close: #233 

https://github.com/user-attachments/assets/85763faf-d926-4f0e-b4fa-cc68f96ef9ad


WilsonNet:
Please add in the tree listing even though it is not in the ticket

WilsonNet (approved):
Looks good
But I think you could add to the tree listing

PR #254, opened 2024-09-06
feat: group tree list status

loadez:
Group status from the tree list page

closes: 232


dashboard/src/components/Status/Status.tsx
```diff
@@ -73,6 +74,50 @@ export const TestStatus = ({
...
 };
 
+export const GroupedTestStatus = ({
+  pass,
+  error,
+  miss,
+  fail,
+  done,
+  skip,
+  forceNumber = true,
```
WilsonNet, line 84: you can remove forced here since we are not using

WilsonNet (approved):

PR #256, opened 2024-09-06
feat(dashboard): "Under Development" message

lucasbracher:
- shows "Under Development" message when "Devices" and "Labs" labels
   are clicked
 - labels change when clicked

Closes https://github.com/kernelci/dashboard/issues/223

### DIFF
--- a/dashboard/src/components/SideMenu/SideMenu.tsx
+++ b/dashboard/src/components/SideMenu/SideMenu.tsx
@@ -1,11 +1,11 @@
+import { useMemo } from 'react';
+
 import { MdOutlineMonitorHeart } from 'react-icons/md';
 
 import { ImTree, ImImages } from 'react-icons/im';
 import { HiOutlineDocumentSearch } from 'react-icons/hi';
 
-import { useRouter, useSearch } from '@tanstack/react-router';
-
-import { useMemo } from 'react';
+import { useRouter, useSearch, useLocation } from '@tanstack/react-router';
 
 import { MessagesKey } from '@/locales/messages';
 
@@ -26,7 +26,7 @@ import SendFeedback from './SendFeedback';
 import NavLink from './NavLink';
 
 type RouteMenuItems = {
-  onClick: () => void;
+  navigateTo: string;
   idIntl: MessagesKey;
   icon: JSX.Element;
   selected: boolean;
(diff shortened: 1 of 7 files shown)


dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -20,13 +22,14 @@ import {
...
 type MenuItems = {
+  id: number;
   onClick: () => void;
   idIntl: MessagesKey;
   icon: JSX.Element;
   selected: boolean;
 };
 
-const emptyFunc = (): void => {};
+//const emptyFunc = (): void => {};
```
lfjnascimento, line 32: leftover

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -68,19 +72,22 @@ const SideMenu = (): JSX.Element => {
 
   const items: MenuItems[] = [
     {
+      id: 0,
```
lfjnascimento, line 75: we can just use the item index for this purpose

WilsonNet: we should use the path url for this

dashboard/src/routes/underDevelopment.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { z } from 'zod';
+
+import { FormattedMessage } from 'react-intl';
+
+import { zOrigin } from '@/types/tree/Tree';
+
+export const HomeSearchSchema = z.object({
+  origin: zOrigin,
+});
+
```
WilsonNet, line 12: the zOrigin is already in the root path, we don't need it here

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -49,6 +54,7 @@ const linkItems: LinkMenuItems[] = [
 const emptyFunc = (): void => {};
 
 const SideMenu = (): JSX.Element => {
+  const [selectedItemId, setSelectedItemId] = useState<number>(0);
```
WilsonNet, line 57: we can extract the selected item just by getting the path from the router and seeing if equals the path that the link brings us

WilsonNet: no need for useState here

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -126,10 +129,15 @@ const SideMenu = (): JSX.Element => {
...
+              item.id === selectedItemId
+                ? selectedItemClassName
+                : notSelectedItemClassName
             }
             key={item.idIntl}
-            onClick={item.onClick}
+            onClick={() => {
+              setSelectedItemId(item.id); // Atualizando o item selecionado
+              item.onClick(); // Navegando ao clicar
+            }}
```
WilsonNet, line 140: Remove the portuguese comments also

dashboard/src/components/SideMenu/SideMenu.tsx
```diff
@@ -46,47 +46,65 @@ const linkItems: LinkMenuItems[] = [
...
-  const useNavigateTo = (path: string): (() => void) => {
-    const router = useRouter();
-    const { origin: unsafeOrigin } = useSearch({ strict: false });
+  const { pathname } = useLocation();
+
+  const isCurrentPath = pathname === item.navigateTo;
+
+  const router = useRouter();
+
+  const { origin: unsafeOrigin } = useSearch({ strict: false });
```
lfjnascimento, line 65: A pattern that helps us not to break the hook rules is to aggregate all hooks at the beginning of the component

PR #278, opened 2024-09-10
fix: wrong usage of misc platform field

### LINKED ISSUE #261: Add platform data to Test Details Page (from misc field, works only for maestro)

lfjnascimento:
- misc fields are json strings and must be parsed before being used as an object

| Before | After |
|--------|--------|
| image | image |

Close #261


dashboard/src/pages/TestDetails/TestDetails.tsx
```diff
@@ -53,6 +53,11 @@ const LogExcerpt = ({ test }: TTestDetailsDefaultProps): JSX.Element => {
 
 const TestDetailsSection = ({ test }: { test: TTestDetails }): JSX.Element => {
   const intl = useIntl();
+  const misc = test.environment_misc ?? test.misc;
+  const platform: string = misc
+    ? JSON.parse(misc).platform
```
lfjnascimento, line 58: would be better to move the parsing to the data fetch function

WilsonNet (approved):
nice

PR #281, opened 2024-09-11
feat: improve tests boot query peformance

loadez:
Check if the information present on the endpoint /tree/<commit_hash>/full is sufficient for the tests and boots tabs.

http://localhost:8000/api/tree/4f8fa5a023dce383394ddfbfdd908964226b051a/full?origin=maestro&git_url=https://git.kernel.org/pub/scm/linux/kernel/git/broonie/sound.git&git_branch=for-next

http://localhost:8000/api/tree/3e7256c41c6572b4e5b6eb024090f8b0730d7304/full?origin=maestro&git_url=https://android.googlesource.com/kernel/common&git_branch=android-mainline

### DIFF
--- a/backend/kernelCI_app/urls.py
+++ b/backend/kernelCI_app/urls.py
@@ -11,6 +11,8 @@ urlpatterns = [
         "tree/<str:commit_hash>/tests/", views.TreeTestsView.as_view(), name="treeTests"
     ),
     path("build/<str:build_id>/status-count", views.BuildStatusCountView.as_view(), name="buildStatusCount"),
+    path("tree/<str:commit_hash>/full",
+         views.TreeDetailsSlow.as_view(), name="TreeDetailsSlow"),
     path("build/<str:build_id>", views.BuildDetails.as_view(), name="buildDetails"),
     path("build/<str:build_id>/tests", views.BuildTests.as_view(), name="buildTests"),
     path("tests/test/<str:test_id>", views.TestDetails.as_view(), name="testDetails"),
--- a/backend/kernelCI_app/views/treeDetailsSlowView.py
+++ b/backend/kernelCI_app/views/treeDetailsSlowView.py
@@ -1,14 +1,50 @@
-from django.http import JsonResponse
+from django.http import JsonResponse, HttpResponseBadRequest
 from django.views import View
-from kernelCI_app.utils import extract_error_message, extract_platform
+from kernelCI_app.utils import (FilterParams, extract_error_message,
+                                extract_platform, InvalidComparisonOP, getErrorResponseBody, toIntOrDefault)
 from django.db import connection
 
 
 class TreeDetailsSlow(View):
+    def __processFilters(self, request):
+        filterTestStatus = ["FAIL", "MISS", "PASS", "DONE", "ERROR", "SKIP"]
+        filterTestDurationMin, filterTestDurationMax = None, None
+        filterBootStatus = ["FAIL", "MISS", "PASS", "DONE", "ERROR", "SKIP"]
+        filterBootDurationMin, filterBootDurationMax = None, None
+
+        try:
+            filter_params = FilterParams(request)
+            for f in filter_params.filters:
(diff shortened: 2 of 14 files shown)


backend/kernelCI_app/views/treeDetailsSlowView.py
```diff
@@ -0,0 +1,202 @@
+from django.http import JsonResponse
+from django.views import View
+from kernelCI_app.utils import extract_error_message, extract_platform
+from django.db import connection
+
+
+class TreeDetailsSlow(View):
```
WilsonNet, line 7: please add a httpie request so we can easily test it

backend/kernelCI_app/views/treeTestsView.py
```diff
@@ -1,46 +1,16 @@
...
 from django.http import HttpResponseBadRequest, JsonResponse
 from django.views import View
-from kernelCI_app.utils import FilterParams, InvalidComparisonOP, getErrorResponseBody
+from kernelCI_app.utils import (FilterParams, InvalidComparisonOP,
+                                extract_error_message, extract_platform, getErrorResponseBody)
 from kernelCI_app.models import Checkouts
-import json
 
 
 class TreeTestsView(View):
```
WilsonNet, line 9: will this endpoint still be necessary? we can delete the unnecessary endpoints

backend/kernelCI_app/views/treeDetailsSlowView.py
```diff
@@ -0,0 +1,202 @@
...
+    def get(self, request, commit_hash: str | None):
```
WilsonNet, line 8: we are missing the filters, we want to get them in this PR or will it be a regression, we can do in a separated PR but targeting this PR and move this to draft

backend/kernelCI_app/views/treeDetailsSlowView.py
```diff
@@ -0,0 +1,202 @@
...
+                            checkouts.git_repository_url = %(git_url_param)s AND
+                            checkouts.git_repository_branch = %(git_branch_param)s AND
+                            checkouts.origin = %(origin_param)s
+                    ) AS tree_head
+                LEFT JOIN builds
+                    ON tree_head.checkout_id = builds.checkout_id
+                WHERE
+                    builds.origin = %(origin_param)s
+            ) AS builds_filter
+        LEFT JOIN tests
```
WilsonNet, line 62: these joins can be inner joins because we are looking for tests, we don't have business with checkouts that does not have builds and builds that does not have tests

dashboard/src/components/Table/TestsTable.tsx
```diff
@@ -25,21 +23,68 @@ import TableStatusFilter from './TableStatusFilter';
...
   const navigate = useNavigate({ from: '/tree/$treeId' });
-  const { data, isLoading } = useTreeTest(
-    treeId,
-    testPath,
-    treeInfo.gitBranch ?? '',
-    treeInfo.gitUrl ?? '',
-    origin,
-  );
+
+  const data = useMemo((): TPathTests[] => {
```
WilsonNet, line 35: I think we can have a loading state if we use a web worker here

loadez: Is it really necessary considering the time frame that we have?

backend/kernelCI_app/urls.py
```diff
@@ -11,6 +11,8 @@
         "tree/<str:commit_hash>/tests/", views.TreeTestsView.as_view(), name="treeTests"
     ),
     path("build/<str:build_id>/status-count", views.BuildStatusCountView.as_view(), name="buildStatusCount"),
+    path("tree/<str:commit_hash>/full",
+         views.TreeDetailsSlow.as_view(), name="TreeDetailsSlow"),
```
WilsonNet, line 15: Slow is not a good name because it may be fast one day, use the functionality

WilsonNet (approved):
looks nice

PR #296, opened 2024-09-13
feat: add filters to megaquery

loadez:


WilsonNet (approved):
looks good, haven't put too much hard thinking in the logic yet, we will need to do that by monday

PR #308, opened 2024-09-16
feat(dashboard): adding tz offset to dates

### LINKED ISSUE #305: Datetime: format all datetimes using timezone info

lucasbracher:
Adds timezone offset to dates at main page and other places:

| Before | After |
|--------|--------|
| Screenshot from 2024-09-16 11-21-15 | image | 

Closes #305

### DIFF
--- a/dashboard/src/components/Accordion/Accordion.tsx
+++ b/dashboard/src/components/Accordion/Accordion.tsx
@@ -28,6 +28,8 @@ import {
   CollapsibleTrigger,
 } from '@/components/ui/collapsible';
 
+import { TooltipDateTime } from '@/components/TooltipDateTime';
+
 import AccordionBuildContent from './BuildAccordionContent';
 
 export interface IAccordion {
@@ -165,7 +167,18 @@ const AccordionBuildsTrigger = ({
     <>
       <TableCell>{triggerInfo.config}</TableCell>
       <TableCell>{triggerInfo.compiler}</TableCell>
-      <TableCell>{triggerInfo.date}</TableCell>
+      <TableCell>
+        {triggerInfo.date ? (
+          <TooltipDateTime
+            dateTime={triggerInfo.date}
+            lineBreak={true}
+            showLabelTime={true}
+            showLabelTZ={true}
+          />
+        ) : (
+          '-'
+        )}
+      </TableCell>
       <TableCell>
         <ColoredCircle
           className="max-w-6"
@@ -255,7 +268,14 @@ const TestTableRow = ({ test, onClick }: ITestTableRow): JSX.Element => {
     >
       <TableCell>{test.path}</TableCell>
       <TableCell>{test.status}</TableCell>
-      <TableCell>{test.start_time ?? '-'}</TableCell>
+      <TableCell>
+        <TooltipDateTime
+          dateTime={test.start_time}
+          lineBreak={true}
+          showLabelTime={true}
+          showLabelTZ={true}
+        />
+      </TableCell>
       <TableCell>{test.duration ?? '-'}</TableCell>
       <TableCell>
         <ChevronRightAnimate />
--- a/dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
+++ b/dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
@@ -1,0 +1,56 @@
+import { format, isValid } from 'date-fns';
+
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
+import { getDateOffset } from '@/utils/utils';
+
+type TooltipDateTimeProps = {
+  dateTime: string;
+  dateFormat?: string;
+  timeFormat?: string;
+  lineBreak?: boolean;
+  showLabelTime?: boolean;
+  showLabelTZ?: boolean;
+  showTooltip?: boolean;
+};
+
+const TooltipDateTime = ({
+  dateTime,
+  dateFormat,
+  timeFormat,
+  lineBreak = true,
+  showLabelTime,
+  showLabelTZ = false,
+  showTooltip = true,
+}: TooltipDateTimeProps): JSX.Element => {
+  const dateObj = new Date(dateTime);
+  if (!isValid(dateObj)) return <div>-</div>;
+
+  const date = dateFormat
+    ? format(dateObj, dateFormat)
+    : dateObj.toLocaleDateString();
+  const time = timeFormat
+    ? format(dateObj, timeFormat)
+    : dateObj.toLocaleTimeString();
+  const tz = getDateOffset(dateObj);
+
+  return (
+    <Tooltip>
+      <TooltipTrigger>
+        <div>
+          {date} {showLabelTime ? time : ''} {showLabelTZ ? tz : ''}
+        </div>
+      </TooltipTrigger>
+      {showTooltip && (
+        <TooltipContent>
+          <div className="text-center">
+            {date}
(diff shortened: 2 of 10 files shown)


dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,62 @@
...
+  dateFormat,
+  timeFormat,
+  lineBreak,
+  showTime,
+  showTZ,
+}: TooltipDateTimeProps): JSX.Element => {
+  let dateObj: Date;
+
+  if (!dateTime) {
+    dateObj = new Date();
```
WilsonNet, line 28: this is a violation of the rules of react, it will make the component impure

https://react.dev/reference/rules

https://react.dev/reference/react/PureComponent

https://react.dev/learn/keeping-components-pure


you need to initiate this date inside a set state    `useState(date ?? new Date())`

WilsonNet: also this is mislieading, this component should always receive a datetime, it is not this components responsability to create a datetime

imagine if it is `undefined` in the database, it will just create a misleading datetime here

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,62 @@
+import { format } from 'date-fns';
+
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
+import { getDateOffset } from '@/utils/utils';
+
+type TooltipDateTimeProps = {
+  dateTime?: Date | string;
```
WilsonNet, line 7: this is not optional

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
lfjnascimento: I don't understand the logic here. If you don't receive a dateTime, will you use today as the date?

lfjnascimento: IMO we have two better approaches in a situation that we need to do something like:
`const [currentDateTime, ] = useState<Date>(date ?? new Date())`

1. `useMemo` with an empty deps array:
`const currentDateTime = useMemo(()=> new Date(), [])`

2. `useRef` with initial value
`const currentDateTime = useRef(new Date())`

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,64 @@
...
+}: TooltipDateTimeProps): JSX.Element => {
+  const dateObj = new Date(dateTime);
+
+  let date = dateFormat
+    ? format(dateObj, dateFormat)
+    : dateObj.toLocaleDateString();
+  let time = timeFormat
+    ? format(dateObj, timeFormat)
+    : dateObj.toLocaleTimeString();
+  let tz = getDateOffset(dateObj);
```
WilsonNet, line 35: you are not redefining `tz` here so it should be const

tip: always use `const` for variable, if there is a moment that you need to redefine, change it to let when that moment arrives

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,64 @@
...
+        ''
+      ) : (
+        <TooltipContent>
+          <div className="text-center">
+            {date}
+            {lineBreak ? <br /> : ' '}
+            {time} {timeFormat && dateFormat ? '' : tz}
+          </div>
+        </TooltipContent>
+      )}
```
WilsonNet, line 59: In react it is common to use short circuit with the `and` operator (`&&`), it is also possible to use with `or` but that is not common

`showTootltip && (component)` 

https://react.dev/learn/conditional-rendering

dashboard/src/components/Accordion/Accordion.tsx
```diff
@@ -165,7 +167,14 @@ const AccordionBuildsTrigger = ({
     <>
       <TableCell>{triggerInfo.config}</TableCell>
       <TableCell>{triggerInfo.compiler}</TableCell>
-      <TableCell>{triggerInfo.date}</TableCell>
+      <TableCell>
+        <TooltipDateTime
+          dateTime={triggerInfo.date || ''}
```
WilsonNet, line 172: it shouldn't pass an empty string if there is no date
it should just not render the component
otherwise, it will get an Invalid Date

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,61 @@
+import { format } from 'date-fns';
+
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
+import { getDateOffset } from '@/utils/utils';
+
+type TooltipDateTimeProps = {
+  dateTime: string;
+  forceNoDate?: boolean;
```
WilsonNet, line 8: you defined this prop but didn't use in a single place, please YAGNI it

dashboard/src/pages/TreeDetails/Tabs/Build/BuildTab.tsx
```diff
@@ -156,7 +156,7 @@ const BuildTab = ({ treeDetailsData }: BuildTab): JSX.Element => {
       ) : (
         '-'
       ),
-      date: row.date?.split('T')[0],
+      date: row.date ?? '-',
```
WilsonNet, line 159: same issue here, you are paassing an invalid date that is tricking the typing because it is using a string instead of a valid timestamp  `new Date('-')' will return `Invalid Date`

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,55 @@
...
+const TooltipDateTime = ({
+  dateTime,
+  dateFormat,
+  timeFormat,
+  lineBreak,
+  showLabelTime,
+  showLabelTZ,
+  showTooltip,
+}: TooltipDateTimeProps): JSX.Element => {
+  const dateObj = new Date(dateTime);
```
WilsonNet, line 25: not all strings are valid datetime, so please validate the string and if it is not a valid datetime, return something like `<div> - </div>`

WilsonNet: if you do this you can make it accept `dateTime?:string` again since there will be an internal validation

WilsonNet (approved):
nice job

dashboard/src/components/TooltipDateTime/TooltipDateTime.tsx
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const TooltipDateTime = ({
+  dateTime,
+  dateFormat,
+  timeFormat,
+  lineBreak,
+  showLabelTime,
+  showLabelTZ,
+  showTooltip,
```
lfjnascimento, line 23: The majority of the use will be with `lineBreak`, `showLabelTZ` and `showTooltip` so we can let this values as `true` by default

```suggestion
  lineBreak,
  showLabelTime,
  showLabelTZ,
  showTooltip,
```

```suggestion
  lineBreak = true,
  showLabelTime,
  showLabelTZ = true,
  showTooltip = true,
```
and then just use 
```
<TooltipDateTime dateTime={test.startTime} />
```

lfjnascimento (approved):

PR #310, opened 2024-09-16
fix: test path was being used as test id

lfjnascimento:
Fix: The test tab was using` test.path` to navigate to TestDetails when it should have been using `test.id`

| Before | After |
|--------|--------|
| image | image |


WilsonNet (approved):
looks good

loadez (approved):

Issue #301, opened 2024-09-16
Add query level caching


joaobertacchi:
Add cache to queries that all the data needed for the filtering is available from the first query.

PR #311, opened 2024-09-16
feat: add query level cache logic

loadez:
Creates the basic logic for caching database queries, currently it will only be applied to the slowest query.

Test by going to a tree details page and filtering the tests or boots tabs, the first load should be normal but any other filtering by status should be faster.

Closes: #301

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -196,7 +196,7 @@ MIGRATION_MODULES = DisableMigrations()
 
 CORS_ALLOW_ALL_ORIGINS = False
 
-CACHE_TIMEOUT = int(get_json_env_var("CACHE_TIMEOUT", "600"))
+CACHE_TIMEOUT = int(get_json_env_var("CACHE_TIMEOUT", "180"))
 
 if DEBUG:
     CORS_ALLOWED_ORIGIN_REGEXES = [
--- a/backend/kernelCI_app/cache.py
+++ b/backend/kernelCI_app/cache.py
@@ -1,0 +1,83 @@
+from django.core.cache import cache
+from django.db import connection
+from threading import Thread
+from django.conf import settings
+
+NOTIFY_CHANNEL = "teste"
+timeout = settings.CACHE_TIMEOUT
+
+__commit_lookup = {}
+__build_lookup = {}
+__test_lookup = {}
+
+
+def __createCacheParamsHash(params: dict):
+    params_list = list(params.items())
+    params_list.sort(key=lambda x: x[0])
+    params_string = ",".join([str(i[1]) for i in params_list])
+    return hash(params_string)
+
+
+def setQueryCache(key, params, rows, commit_hash=None, build_id=None, test_id=None,
+                  timeout=timeout):
+    params_hash = __createCacheParamsHash(params)
+    hash_key = "%s-%s" % (key, params_hash)
+
+    __addToLookup(hash_key, commit_hash, __commit_lookup)
+    __addToLookup(hash_key, build_id, __build_lookup)
+    __addToLookup(hash_key, test_id, __test_lookup)
+
+    return cache.set(hash_key, rows, timeout)
+
+
+def getQueryCache(key, params: dict):
+    params_hash = __createCacheParamsHash(params)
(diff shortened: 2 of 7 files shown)


backend/kernelCI_app/utils.py
```diff
@@ -1,10 +1,29 @@
 import json
 from typing import Union
 from django.utils import timezone
+from django.core.cache import cache
 from datetime import timedelta
 import re
 
 DEFAULT_QUERY_TIME_INTERVAL = {'days': 7}
+DEFAULT_QUERY_CACHE = 180
```
lfjnascimento, line 9: this should be an env var right?

lfjnascimento (approved):
Big win, working beautifully!

backend/kernelCI_app/utils.py
WilsonNet: I think in debug this should be 0 also

backend/kernelCI_app/cache.py
```diff
@@ -0,0 +1,83 @@
...
+    __addToLookup(hash_key, commit_hash, __commit_lookup)
+    __addToLookup(hash_key, build_id, __build_lookup)
+    __addToLookup(hash_key, test_id, __test_lookup)
+
+    return cache.set(hash_key, rows, timeout)
+
+
+def getQueryCache(key, params: dict):
+    params_hash = __createCacheParamsHash(params)
+    return cache.get("%s-%s" % (key, hash(params_hash)))
```
WilsonNet, line 35: you are already hashing the params in `__createCacheParamsHash`

backend/kernelCI_app/cache.py
```diff
@@ -0,0 +1,83 @@
...
+    params_list = list(params.items())
+    params_list.sort(key=lambda x: x[0])
+    params_string = ",".join([i[1] for i in params_list])
+    return hash(params_string)
+
+
+def setQueryCache(key, params, rows, commit_hash=None, build_id=None, test_id=None,
+                  timeout=DEFAULT_QUERY_CACHE):
+    params_hash = __createCacheParamsHash(params)
+    hash_key = "%s-%s" % (key, hash(params_hash))
```
WilsonNet, line 24: ditto about params already being hashed in createCacheParamsHash

backend/kernelCI_app/cache.py
```diff
@@ -0,0 +1,83 @@
...
+
+    __addToLookup(hash_key, commit_hash, __commit_lookup)
+    __addToLookup(hash_key, build_id, __build_lookup)
+    __addToLookup(hash_key, test_id, __test_lookup)
```
WilsonNet, line 28: this seems like a modular utility for caching function but this section is making it specific to the tests/boots, at least add a todo here

backend/kernelCI_app/cache.py
```diff
@@ -0,0 +1,83 @@
...
+        with connection.cursor() as c:
+            c.execute("LISTEN %s" % NOTIFY_CHANNEL)
+            for n in c.connection.notifies():
+                if n.channel == NOTIFY_CHANNEL:
+                    __processUpdate(n.payload)
+    except Exception as ex:
+        print(ex)
+
+
+def __addToLookup(key, value, lookup):
```
WilsonNet, line 78: what is the purpose for this lookup, it is being added but never read

WilsonNet, line 78: these names are a little confusing

maybe this could be `cacheKey` and instead of `value` it could be `specificPropertyKey` 
because the value is being used as the key

WilsonNet (approved):
looks good

backend/kernelCI/settings.py
```diff
@@ -206,7 +206,7 @@ def __getitem__(self, item):
     CSRF_COOKIE_SECURE = False
     SECURE_SSL_REDIRECT = False
     SECURE_HSTS_SECONDS = 3600
-    CACHE_TIMEOUT = 0
+    CACHE_TIMEOUT = 60
```
WilsonNet, line 209: I think we want debug without cache as the default behaior, otherwise it get's annoying to develop since every change will take one minute to reflect

WilsonNet (approved):

PR #319, opened 2024-09-18
feat(dashboard): lighter red for 0 errors

### LINKED ISSUE #316: Use lighter color for zero'ed circle statuses

lucasbracher:
Improves visibility of zeroed errors.

| Before | After |
|--------|--------|
| image | image |

Closes #316

### DIFF
--- a/dashboard/src/components/ColoredCircle/ColoredCircle.tsx
+++ b/dashboard/src/components/ColoredCircle/ColoredCircle.tsx
@@ -47,6 +47,12 @@ const ColoredCircle = ({
           'inline-flex h-6 w-fit min-w-6 items-center justify-center rounded-full px-2 text-sm text-black',
           className,
           backgroundClassName,
+          {
+            ['bg-opacity-20']:
+              quantity == 0 &&
+              (tooltipText == 'global.failed' ||
+                tooltipText == 'global.inconclusive'),
+          },
         )}
       >
         {quantity}


dashboard/src/components/ColoredCircle/ColoredCircle.tsx
```diff
@@ -40,13 +40,15 @@ const ColoredCircle = ({
   className,
   tooltipText,
 }: IColoredCircle): JSX.Element => {
+  const adHocClearerRed = quantity == 0 ? 'bg-opacity-40' : '';
```
WilsonNet, line 43: `isDimmed` owuld probably be a better name

dashboard/src/components/ColoredCircle/ColoredCircle.tsx
```diff
@@ -40,13 +40,19 @@ const ColoredCircle = ({
...
+      ? 'bg-opacity-20'
+      : '';
   return (
     <TooltipWrapper tooltipText={tooltipText}>
       <div
         className={cn(
           'inline-flex h-6 w-fit min-w-6 items-center justify-center rounded-full px-2 text-sm text-black',
           className,
           backgroundClassName,
+          clearerCircles,
```
WilsonNet, line 55: `cn` is a wrapper of the clsx library, you can do the contitional in another way 

```
{ ['bg-opacity-20']: quantity == 0 &&
    (tooltipText == 'global.failed' || tooltipText == 'global.inconclusive') }
```

but this way is also fine 

https://github.com/JedWatson/classnames

dashboard/src/components/ColoredCircle/ColoredCircle.tsx
```diff
@@ -40,13 +40,19 @@ const ColoredCircle = ({
   className,
   tooltipText,
 }: IColoredCircle): JSX.Element => {
+  const clearerCircles =
+    quantity == 0 &&
+    (tooltipText == 'global.failed' || tooltipText == 'global.inconclusive')
+      ? 'bg-opacity-20'
+      : '';
```
lfjnascimento, line 47: Maybe we can put this behavior in every circle not just for `failed` or `inconclusive` IMO that will improve consistency

lfjnascimento: If we can't, create a `fadeOnEmpty` prop for the component, as `tooltipText` is not a suitable value for this.

lucasbracher: I think the same, but ticket's title is clear.

lucasbracher:
Got back to 40% opacity because %20 was harder in order to distinct gray and lighest red:

On 20% opacity:

image

On 40% opacity:

image

lfjnascimento (approved):

Issue #302, opened 2024-09-16
Make filters affect all elements in all the tabs


joaobertacchi:
Missing #264 #378

WilsonNet:
this still need the graph fillters to be working properly

PR #325, opened 2024-09-19
Feat/add filters tests

WilsonNet:
# Add filter to tests and boots in tree details

- Changed filter lists to sets for better performance and clarity.
- Updated variable names for better readability.
- Added additional filters for configs, compiler, and architecture.
- Refactored the filtering logic in the `TreeDetailsSlow` view.
- Updated the frontend to handle new filter parameters.
- Removed unnecessary useSearch hook in BootsTab and TestsTab.

## How to test
Go to any tree that has tests eg: (broonie-sounds), and mess with the filters, see if it affects the boots/tests tab 

It is really hard to know if it is right, so, check the test details to see which filters you can use, because not all builds have tests

Coses #302, #326

### DIFF
--- a/backend/kernelCI_app/utils.py
+++ b/backend/kernelCI_app/utils.py
@@ -4,7 +4,7 @@ from django.utils import timezone
 from datetime import timedelta
 import re
 
-DEFAULT_QUERY_TIME_INTERVAL = {'days': 7}
+DEFAULT_QUERY_TIME_INTERVAL = {"days": 7}
 
 
 def toIntOrDefault(value, default):
@@ -25,7 +25,6 @@ def extract_platform(misc_environment: Union[str, dict, None]):
     platform = parsedEnvMisc.get("platform")
     if platform:
         return platform
-    print("unknown platform in misc_environment", misc_environment)
     return "unknown"
 
 
@@ -41,7 +40,6 @@ def extract_error_message(misc: Union[str, dict, None]):
     error_message = parsedEnv.get("error_msg")
     if error_message:
         return error_message
-    print("unknown error_msg in misc", misc)
     return "unknown error"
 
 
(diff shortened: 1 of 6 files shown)


backend/kernelCI_app/utils.py
```diff
@@ -55,50 +55,54 @@ def getErrorResponseBody(reason: str):
...
+        "lte": ["lte", "<="],
     }
 
     def __init__(self, request):
         self.filters = []
         self.create_filters_from_req(request)
 
     def create_filters_from_req(self, request):
+        print(request.GET.keys())
+        print(request.GET)
```
lfjnascimento, line 93: leftover

backend/kernelCI_app/views/treeDetailsSlowView.py
```diff
@@ -1,49 +1,82 @@
...
+                    filterBootStatus.add(value)
                 elif field == "boot.duration":
                     if operation == "lte":
                         filterBootDurationMax = toIntOrDefault(value, None)
                     else:
                         filterBootDurationMin = toIntOrDefault(value, None)
                 if field == "test.status":
-                    filterTestStatus = value
+                    filterTestStatus.add(value)
+                if field == "config_name":
```
lfjnascimento, line 39: Have you considered using a `switch/case`? I'm not sure, but maybe a `switch/case` would also help us not need the `# noqa: max-complexity=13`

backend/kernelCI_app/utils.py
```diff
@@ -41,7 +41,7 @@ def extract_error_message(misc: Union[str, dict, None]):
     error_message = parsedEnv.get("error_msg")
     if error_message:
         return error_message
-    print("unknown error_msg in misc", misc)
+    # print("unknown error_msg in misc", misc)
```
loadez, line 44: remove unnecessary comment

backend/kernelCI_app/utils.py
```diff
@@ -25,7 +25,7 @@ def extract_platform(misc_environment: Union[str, dict, None]):
     platform = parsedEnvMisc.get("platform")
     if platform:
         return platform
-    print("unknown platform in misc_environment", misc_environment)
+    # print("unknown platform in misc_environment", misc_environment)
```
loadez, line 28: remove unnecessary comment

backend/kernelCI_app/views/treeDetailsSlowView.py
loadez: Heads up: switch/case is match in python.

The way to reduce complexity is to divide this function in smaller ones

backend/kernelCI_app/views/treeDetailsSlowView.py
```diff
@@ -1,49 +1,82 @@
...
+            filterBootStatus,
+            filterBootDurationMin,
+            filterBootDurationMax,
+            filterTreeDetailsConfigs,
+            filterTreeDetailsCompiler,
+            filterArchitecture,
+        )
 
-    def get(self, request, commit_hash: str | None):
+    def get(self, request, commit_hash: str | None):  # noqa: max-complexity=13
```
loadez, line 64: ditto

backend/kernelCI_app/views/treeDetailsSlowView.py
WilsonNet: it wasn't needed just splitted into functions

lucasbracher (approved):

loadez (approved):
LGTM

PR #327, opened 2024-09-20
feat: add items per page selector in TableInfo

lfjnascimento:
| Before | After |
|--------|--------|
| image | image |

Close: #175


loadez (approved):
Everything seems alright. 

I would like to have the pagination and items per page in the url but that would cause problems when navigating to new pages. So LGTM.

lfjnascimento:
> Everything seems alright.
> 
> I would like to have the pagination and items per page in the url but that would cause problems when navigating to new pages. So LGTM.

I understand your consideration, the problem is that our URL is getting huge, we have to be more restrictive with what we put in the URL.

WilsonNet (approved):
looks good

PR #340, opened 2024-09-25
feat/lazy lazy load

### LINKED ISSUE #341: Add a fast path and a lazy load to the trees query

WilsonNet:
- **feat(urls): add fast tree view endpoint**
- **feat: add fast path response handling in Tree API**

## Test

Run local and open the Maestro branch

Closes #341

### DIFF
--- a/backend/kernelCI_app/urls.py
+++ b/backend/kernelCI_app/urls.py
@@ -24,6 +24,10 @@ urlpatterns = [
          viewCache(views.TreeView),
          name="tree"
          ),
+    path("tree-fast/",
+         viewCache(views.TreeViewFast),
+         name="tree-fast"
+         ),
     path("tree/tests/",
          viewCache(views.groupedTests),
          name="treeGroupedTests"
--- a/backend/kernelCI_app/views/treeViewFast.py
+++ b/backend/kernelCI_app/views/treeViewFast.py
@@ -1,0 +1,66 @@
+from django.http import JsonResponse
+from django.views import View
+from kernelCI_app.models import Checkouts
+
+
+DEFAULT_ORIGIN = "maestro"
+
+
+class TreeViewFast(View):
+    def get(self, request):
+        origin = request.GET.get("origin", DEFAULT_ORIGIN)
+
+        checkouts = Checkouts.objects.raw(
+            """
+            WITH ordered_checkouts AS (
+                SELECT
+                    id,
+                    tree_name,
+                    git_repository_branch,
+                    git_repository_url,
+                    git_commit_hash,
+                    git_commit_name,
+                    patchset_hash,
+                    start_time,
+                    ROW_NUMBER() OVER
+                    (PARTITION BY git_repository_url, git_repository_branch ORDER BY start_time DESC)
+                    as time_order
+                FROM
+                    checkouts
+                WHERE
+                    origin = %s
+            )
+            SELECT
+                id,
+                tree_name,
+                git_repository_branch,
+                git_repository_url,
+                git_commit_hash,
+                git_commit_name,
+                patchset_hash,
+                start_time
+            FROM
+                ordered_checkouts
+            WHERE
+                time_order = 1
+            ORDER BY
+                tree_name ASC;
+            """,
+            [origin],
+        )
+
+        response_data = [
+            {
+                "id": checkout.id,
+                "tree_name": checkout.tree_name,
+                "git_repository_branch": checkout.git_repository_branch,
+                "git_repository_url": checkout.git_repository_url,
+                "git_commit_hash": checkout.git_commit_hash,
+                "git_commit_name": checkout.git_commit_name,
+                "patchset_hash": checkout.patchset_hash,
+                "start_time": checkout.start_time,
+            }
+            for checkout in checkouts
+        ]
+
+        return JsonResponse(response_data, safe=False)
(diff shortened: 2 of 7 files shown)


backend/kernelCI_app/urls.py
```diff
@@ -24,6 +24,10 @@ def viewCache(view):
          viewCache(views.TreeView),
          name="tree"
          ),
+    path("tree-fast/",
```
lucasbracher, line 27: I'd rather use the same endpoint and pass an optional boolean GET argument from `.env`'s frontend.

backend/kernelCI_app/views/treeViewFast.py
```diff
@@ -0,0 +1,66 @@
...
+                ordered_checkouts
+            WHERE
+                time_order = 1
+            ORDER BY
+                tree_name ASC;
+            """,
+            [origin],
+        )
+
+        response_data = [
```
lucasbracher, line 52: I think it's better to create a serializer here, like `TreeSerializer`.

WilsonNet: The slow query will have to be reviewed, we can use that after this

WilsonNet: Added a TODO comment

dashboard/src/api/Tree.tsx
```diff
@@ -9,9 +11,31 @@ const fetchTreeCheckoutData = async (origin: string): Promise<Tree[]> => {
   return res.data;
 };
 
-export const useTreeTable = (origin: string): UseQueryResult<Tree[]> => {
+export const useTreeTable = ({
+  enabled,
+}: {
+  enabled: boolean;
+}): UseQueryResult<Tree[]> => {
+  const { origin } = useSearch({ from: '/tree' });
```
lfjnascimento, line 19: IMO it would be better to receive the origin as an argument rather than to obtain it itself, that way its more clear what values is being used in the query

dashboard/src/api/Tree.tsx
```diff
@@ -9,9 +11,31 @@ const fetchTreeCheckoutData = async (origin: string): Promise<Tree[]> => {
...
+
+const fetchTreeFastCheckoutData = async (
+  origin: string,
+): Promise<TreeFastPathResponse> => {
+  const res = await http.get('/api/tree-fast/', { params: { origin } });
+  return res.data;
+};
+
+export const useTreeTableFast = (): UseQueryResult<TreeFastPathResponse> => {
+  const { origin } = useSearch({ from: '/tree' });
```
lfjnascimento, line 36: same about origin

dashboard/src/api/Tree.tsx
WilsonNet: this query is specific to for tree, there are plenty of places to see what is being passed to the query including the network tab

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -1,81 +1,107 @@
...
-  const { origin: unsafeOrigin } = useSearch({ strict: false });
-  const origin = zOrigin.parse(unsafeOrigin);
   const [itemsPerPage, setItemsPerPage] = useState(ItemsPerPageValues[0]);
 
-  const { data, error, isLoading } = useTreeTable(origin);
+  const { data: fastData, status } = useTreeTableFast();
+
+  const { data, error, isLoading } = useTreeTable({
+    enabled: status === 'success' && !!fastData,
+  });
```
lfjnascimento, line 36: Have you considered the possibility of doing it in parallel?

dashboard/src/api/Tree.tsx
WilsonNet: also, the hook is already an abstraction over the fetch request

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -1,81 +1,107 @@
...
+  return 'build_status' in data;
+}
+
 const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
-  const { origin: unsafeOrigin } = useSearch({ strict: false });
-  const origin = zOrigin.parse(unsafeOrigin);
   const [itemsPerPage, setItemsPerPage] = useState(ItemsPerPageValues[0]);
 
-  const { data, error, isLoading } = useTreeTable(origin);
+  const { data: fastData, status } = useTreeTableFast();
```
lfjnascimento, line 32: NIT: a bit confusing not use a "fastStatus" also
```suggestion
  const { data: fastData, status: fastStatus } = useTreeTableFast();
```

lfjnascimento:
what if we had a hook that handles the two api calls by it self so we can use it as a single source of data? `const {data, error, isLoading  } = useLazyTreeTable();`

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
WilsonNet: I think it is better if is in series because we want the fast path ASAP, if the fast path is taking time to load, in parallel will stress an already stressed database

WilsonNet:
I agree that we should combine the hooks, added as a todo so we can move forward, already too much time was spent in this task

lfjnascimento (approved):

PR #374, opened 2024-10-10
docs: add a note in the onboarding about running the backend

### LINKED ISSUE #373: Add a note in Onboarding about run backend
It is possible to have authentication issues escaping special characters. Then a note should be added with tips to solve them.

anajalvarenga:
Add a note in the onboarding about running the backend locally
- closes #373

### DIFF
--- a/docs/Onboarding.md
+++ b/docs/Onboarding.md
@@ -60,6 +60,9 @@ export DB_DEFAULT="{
 poetry run python3 manage.py runserver
 ```
 
+> Note:
+> It is possible to have authentication issues when escaping special characters. In some cases, it is necessary to add more than one backslash, while in others, no addition is needed. To assist with this, when `DEBUG` is set to `True`, the default database info will be printed in the terminal, allowing you to determine if the characters got escaped as intended.
+
 Definition of Done: You have the KernelCI Dashboard backend running locally.
 
 
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -131,6 +131,9 @@ DATABASES = {
     )
 }
 
+if DEBUG:
+    print("DEBUG: DEFAULT DATABASE:", DATABASES)
+
 
 CACHES = {
     "default": {


docs/Onboarding.md
```diff
@@ -60,6 +60,9 @@ export DB_DEFAULT="{
 poetry run python3 manage.py runserver
 ```
 
+> Note:
+> It is possible to have authentication issues escaping special characters. In some cases, it is necessary to add more than one backslash, while in others, no addition is needed. One way to see how your computer is handling the escapes is to add a print statement for the environment variable `DB_DEFAULT` in the `manage.py` file with `print(os.getenv('DB_DEFAULT'))`.
```
WilsonNet, line 64: I think a more realistic approach would be to print the `DATABASES` variable  in

``` 
DATABASES = {
    "default": get_json_env_var(
        "DB_DEFAULT",
        {
            "ENGINE": "django.db.backends.postgresql",
            "NAME": "kernelci",
            "USER": "kernelci",
            "PASSWORD": "kernelci-db-password",
            "HOST": "127.0.0.1",
            "OPTIONS": {
                "connect_timeout": 5,
            },
        },
    )
}
```

Since we are having much trouble with escaped characters, it is a good idea to print `if DEBUG` is equal to True

WilsonNet: Also, leave that print as a bug for extra precaution, I don't think it has any danger of printing passwords in production, but better safe than sorry

docs/Onboarding.md
```diff
@@ -60,6 +60,9 @@ export DB_DEFAULT="{
 poetry run python3 manage.py runserver
 ```
 
+> Note:
+> It is possible to have authentication issues escaping special characters. In some cases, it is necessary to add more than one backslash, while in others, no addition is needed. To assist with this, when `DEBUG` is set to `True`, the default database will be printed in the terminal, allowing you to determine the best approach.
```
WilsonNet, line 64:
```suggestion
> It is possible to have authentication issues escaping special characters. In some cases, it is necessary to add more than one backslash, while in others, no addition is needed. To assist with this, when `DEBUG` is set to `True`, the default database info will be printed in the terminal, allowing you to determine if the characters got escaped as intended.
```

MarceloRobert, line 64: nit: I think it's best to say "It is possible to have authentication issues _when_ escaping special characters" or "... issues _by_ escaping..."

WilsonNet: `when` is probably the best word here, `by` means that you got a bug because you escaped

WilsonNet (approved):
looks good

mari1912 (approved):

MarceloRobert (commented):
Currently the DB_DEFAULT is being printed if there is a "wrong password" error, but if it is an "invalid \escape" error it won't print. Technically it _does_ show that error when trying to run the backend and it _is_ tackled by the onboarding note, so I don't think there is a real need for change but maybe the onboarding note could be a bit more specific.

anajalvarenga:
> Currently the DB_DEFAULT is being printed if there is a "wrong password" error, but if it is an "invalid \escape" error it won't print. Technically it _does_ show that error when trying to run the backend and it _is_ tackled by the onboarding note, so I don't think there is a real need for change but maybe the onboarding note could be a bit more specific.

Currently DB_DEFAULT is always being printed in debug mode. So, if you have some issues, you can check your terminal and see how your machine is dealing with \escape

Issue #88, opened 2024-08-05
make it possible to change the query limit by time in tree listing page


lfjnascimento:
As discussed in the [Slack thread](https://kernelci.slack.com/archives/C067UJC404A/p1722533269459619?thread_ts=1722355781.145189&cid=C067UJC404A), we should limit queries by time, 7 days by default, but user should be able to change it.

PR #412, opened 2024-10-21
Feat: tree list query limit by time

Francisco2002:
Add an input to allow the user to provide a time parameter in the tree listing page.

How to test:
1. go to http://localhost:5173/tree and interact with the input.
2. go to http://localhost:5173/tree?origin=maestro&time=... and try to pass a non-positive value in time param.

#88


dashboard/src/components/ui/form.tsx
MarceloRobert, line 1: Is this file being used?

dashboard/src/components/ui/label.tsx
MarceloRobert, line 1: ditto about file being used

dashboard/src/hooks/use-toast.ts
MarceloRobert, line 1: Use camel case for the file name (useToast instead of use-toast) and use .tsx

dashboard/src/locales/messages/index.ts
```diff
@@ -134,6 +136,7 @@ export const messages = {
     'global.total': 'Total',
     'global.tree': 'Tree',
     'global.underDevelopment': 'Under Development',
+    'global.unit': 'Days',
```
MarceloRobert, line 139: I would call this "global.days"

dashboard/src/pages/treeConstants.ts
```diff
@@ -0,0 +1,2 @@
+export const DEBOUNCE_INTERVAL = 666;
```
WilsonNet, line 1: I don't think there is any specific reason for this to be a global costat, maybe it could be the default debounce interval for the text input if the user didn't specify any

dashboard/src/hooks/use-toast.ts
```diff
@@ -0,0 +1,200 @@
...
+  toast: ({ ...props }: Toast) => ToastStateProps;
+  dismiss: (toastId?: string) => void;
+  toasts: ToasterToast[];
+};
+
+function useToast(): UseToastType {
+  const [state, setState] = React.useState<State>(memoryState);
+
+  React.useEffect(() => {
+    listeners.push(setState);
```
WilsonNet, line 184: wow this is so hacky, I know it is from shadcn, lets keep this as is since it is working

dashboard/src/components/ui/label.tsx
WilsonNet: we are keepig shad cn components file as they come to identify they are from there, but we can change later

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -135,8 +206,41 @@ const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
...
+  const InputTime = ({
+    field,
+    fieldError,
+  }: {
+    field: ControllerRenderProps<z.infer<typeof InputTimeSchema>, 'time'>;
+    fieldError: FieldError | undefined;
+  }): JSX.Element => {
+    const className = useMemo(
+      () =>
+        `border-${fieldError ? 'red' : 'gray'} mx-[10px] flex w-[100px] flex-1 rounded border`,
```
WilsonNet, line 218: `border-${fieldError ? 'red' : 'gray'}` 

will bug the tailwind compiler, remember that it interprets the test value not the javascript 

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

WilsonNet: use the `cn` utils for it

MarceloRobert (approved):
Looks good to me

dashboard/src/components/DebounceInput/DebounceInput.tsx
```diff
@@ -3,24 +3,26 @@ import { ChangeEvent, useEffect, useRef, useState } from 'react';
...
   startingValue?: string;
 }
 
+const DEBOUNCE_INTERVAL = 666;
+
 const DebounceInput = ({
-  debouncedInterval,
   debouncedSideEffect,
   startingValue = '',
+  onChange,
```
WilsonNet, line 16: I think this is a leftover, you are not using in the TreeListingPage and it feels weird since we have the debouncedSideEffect

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -35,6 +60,52 @@ const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
     enabled: fastStatus === 'success' && !!fastData,
   });
 
+  function validateString(val: string): boolean {
+    const convertedNumber = parseInt(val);
+    return !Number.isNaN(convertedNumber) && convertedNumber > 0;
+  }
+
```
WilsonNet, line 67: since this is a pure function you can leave outside of the component so we don't have to worry about memoization or keeping an stable reference

dashboard/src/components/ui/form.tsx
```diff
@@ -0,0 +1,176 @@
+import * as React from "react"
```
WilsonNet, line 1: I think you never used this file, you can delete it

dashboard/src/components/ui/label.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react"
```
WilsonNet, line 1: since you are not using form, you can delete the label component also

WilsonNet (approved):
it is working really nice, just some few pointers to go

backend/kernelCI_app/views/treeView.py
```diff
@@ -11,6 +11,8 @@
 class TreeView(View):
     def get(self, request):
         origin = request.GET.get('origin', DEFAULT_ORIGIN)
+        time = int(request.GET.get('time', '7'))
+        timedata = {"days": time}
```
lfjnascimento, line 15: I think we should receive it in a more standard way such as a unix timestamp

backend/requests/tree-listing-fast.sh
```diff
@@ -1,3 +1,6 @@
 # If you want to see headers, add -p H to the http call:
 # http -p H 'http://localhost:8000/api/tree/' origin==maestro
 http 'http://localhost:8000/api/tree-fast/' origin==maestro
+
+# If you want to provide another limit to query:
+# http 'http://localhost:8000/api/tree-fast/' origin==maestro value==8
```
lfjnascimento, line 6: is this `value` right? shouldn't it be `time`?

PR #413, opened 2024-10-21
feat: show hardware used in tests in tree details header

### LINKED ISSUE #384: Tree Details Page: show hardware used for testing in the page header

anajalvarenga:
Adding Hardware Used table in Tree details header

image

The table data shouldn't change by adding filters
- closes #384

### DIFF
--- a/dashboard/src/pages/TreeDetails/Tabs/TestCards.tsx
+++ b/dashboard/src/pages/TreeDetails/Tabs/TestCards.tsx
@@ -1,5 +1,5 @@
 import { FormattedMessage } from 'react-intl';
-import { memo } from 'react';
+import { memo, useMemo } from 'react';
 
 import { Link } from '@tanstack/react-router';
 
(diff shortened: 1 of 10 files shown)


Francisco2002:
very nice!!

In the page, the height of loading content is to big comparing with the card height. Make then the same height

Screenshot from 2024-10-23 09-26-41
Screenshot from 2024-10-23 09-26-55

anajalvarenga:
> very nice!!
> 
> In the page, the height of loading content is to big comparing with the card height. Make then the same height
> 
> Screenshot from 2024-10-23 09-26-41 Screenshot from 2024-10-23 09-26-55

Since I don't have the data when the loading is showing, I can't fit it to the same size as the card, but I reduce its height

WilsonNet:
@anajalvarenga just put an average height from something that has 2 lines (there is a prop in the query switcher for that)

WilsonNet (approved):
great job

dashboard/src/pages/TreeDetails/Tabs/TestCards.tsx
```diff
@@ -294,3 +296,36 @@ const IssuesList = ({ issues, title }: IIssuesList): JSX.Element => {
...
+            return (
+              <Badge
+                key={hardware}
+                variant="outline"
+                className="text-sm font-normal"
+              >
+                {hardware}
+              </Badge>
+            );
+          })}
```
lfjnascimento, line 325: put this block inside the `useMemo`

lfjnascimento (approved):

PR #419, opened 2024-10-22
fix: TestsTable was not counting null tests

### LINKED ISSUE #418: [Bug] TestTable with with wrong counting
The `Test Table`has a wrong count, we can see in the image `total_test = 743` but if we add the values ​​it gives `740`
Image

lfjnascimento:
Close #418


WilsonNet (approved):
looks good

PR #458, opened 2024-10-28
feat: rename Devices side menu entry to Hardware

lfjnascimento:
| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/dashboard/src/locales/messages/index.ts
+++ b/dashboard/src/locales/messages/index.ts
@@ -158,7 +158,7 @@ export const messages = {
     'logSheet.fileSize': 'File Size',
     'logSheet.indexOf': 'Index of {link}',
     'logSheet.title': 'Logs Viewer',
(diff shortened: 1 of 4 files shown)


WilsonNet (approved):
nice one

dashboard/src/locales/messages/index.ts
```diff
@@ -158,7 +158,7 @@ export const messages = {
     'logSheet.fileSize': 'File Size',
     'logSheet.indexOf': 'Index of {link}',
     'logSheet.title': 'Logs Viewer',
-    'routes.deviceMonitor': 'Devices',
+    'routes.hardware': 'Hardware',
```
MarceloRobert, line 161: I think that keeping the "Monitor" suffix would be better to keep consistency, it identifies it as being a route, just like "treeMonitor"

lfjnascimento: done

PR #475, opened 2024-10-29
Fix count of tests in TreeListing Page

### LINKED ISSUE #469: BUG Tree listing is counting boots as tests
Image

Image

Francisco2002:
| Old | New |
| ----- | ------ |
| Screenshot from 2024-10-29 11-36-57 | Screenshot from 2024-10-29 11-37-58 |

Close #469

### DIFF
--- a/backend/kernelCI_app/views/treeView.py
+++ b/backend/kernelCI_app/views/treeView.py
@@ -53,17 +53,17 @@ class TreeView(View):
                 COUNT(DISTINCT CASE WHEN builds.valid IS NULL AND builds.id IS NOT NULL THEN builds.id END)
                     AS null_builds,
                 COUNT(DISTINCT builds.id) AS total_builds,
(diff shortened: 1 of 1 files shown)


WilsonNet (approved):
oh that makes total sense

backend/kernelCI_app/views/treeView.py
```diff
@@ -53,17 +53,17 @@ def get(self, request):
                 COUNT(DISTINCT CASE WHEN builds.valid IS NULL AND builds.id IS NOT NULL THEN builds.id END)
                     AS null_builds,
                 COUNT(DISTINCT builds.id) AS total_builds,
-                COUNT(CASE WHEN (tests.path <> 'boot' OR tests.path NOT LIKE 'boot.%%')
```
lfjnascimento, line 56: We're not sure if there is a boot whose path is just `boot`, so I think we should keep `tests.path <> 'boot'` and use an `AND` instead of `OR`

lfjnascimento (approved):
LGTM

PR #492, opened 2024-11-01
feat/hardware listing

WilsonNet:
Add endpoints for the hardware listing page:

This PR adds the endpoint for the hardware listing page, it adds both a fast and slow view that are accessed via query parameters.


backend/kernelCI_app/views/hardwareView.py
```diff
@@ -0,0 +1,70 @@
...
+        ).values("environment_compatible", "status", "start_time", "build__valid")
+
+        hardware = {}
+        for test in tests:
+            for compatible in test["environment_compatible"]:
+                if hardware.get(compatible) is None:
+                    hardware[compatible] = {
+                        "hardwareName": compatible,
+                        "statusCount": defaultdict(int),
+                        "buildCount": {"valid": 0, "invalid": 0, "null": 0},
```
WilsonNet, line 22: buildStatusCount

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -0,0 +1,70 @@
...
+                    hardware[compatible]["statusCount"]["NULL"] += 1
+                else:
+                    hardware[compatible]["statusCount"][test["status"]] += 1
+
+                if test["build__valid"] is None:
+                    hardware[compatible]["buildCount"]["null"] += 1
+                elif test["build__valid"] is True:
+                    hardware[compatible]["buildCount"]["valid"] += 1
+                else:
+                    hardware[compatible]["buildCount"]["invalid"] += 1
```
lfjnascimento, line 34: just a nit but you can create a reusable map in `utils` for build status

```python
build_status_map = {True: 'valid', False: 'invalid', None: 'null'}

# and then
build_status = build_status_map[test["build__valid"]] # or using a function get_build_status(...)
hardware[compatible]["buildCount"][build_status] += 1
```

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -0,0 +1,70 @@
...
+
+        result = {"hardware": list(hardware.values())}
+
+        return result
+
+    def __getFastResult(self, start_date):
+        hardwares = set()
+        tests = Tests.objects.filter(
+            start_time__gte=start_date, environment_compatible__isnull=False
+        ).values("environment_compatible", "start_time")
```
lfjnascimento, line 44: not using `start_time` here

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def get(self, request):
+        mode = request.GET.get("mode", "fast")
+        try:
+            daysInterval = int(request.GET.get("daysInterval", 3))
+        except ValueError:
+            return JsonResponse({"error": "Invalid daysInterval"}, status=400)
+
+        if daysInterval < 1:
+            return JsonResponse({"error": "Invalid daysInterval"}, status=400)
```
lfjnascimento, line 61: we are using this logic in 3 different queries, let's create a function to handle this

lfjnascimento (approved):
LGTM

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -0,0 +1,70 @@
...
+            for compatible in test["environment_compatible"]:
+                hardwares.add(compatible)
+
+        result = {"hardware": list(hardwares)}
+        return result
+
+    def get(self, request):
+        mode = request.GET.get("mode", "fast")
+        try:
+            daysInterval = int(request.GET.get("daysInterval", 3))
```
MarceloRobert, line 56: I would either move this logic to a separate function as @lfjnascimento said or use the same variable name as it is in treeView and treeViewFast (interval_days for the variable and intervalInDays for the url parameter)

MarceloRobert: Also, use a const for the default daysInterval instead of a magic number

Francisco2002 (approved):
Nice!!!!

PR #546, opened 2024-11-13
Fix counts datatables

### LINKED ISSUE #466: Make search affect counts on datatables
"I see I can search for a given test on the bottom, but could I have this to the entire tab? So it would also update the statistics on the top?"

Image

Also, when changing between All/Success/Failed/Inconclusive, the search field should be applied automatically (no need to type again)

Francisco2002:
- make filters affect counts in datatables header

Close #466

### DIFF
--- a/dashboard/src/components/BootsTable/BootsTable.tsx
+++ b/dashboard/src/components/BootsTable/BootsTable.tsx
@@ -13,7 +13,7 @@ import {
   useReactTable,
 } from '@tanstack/react-table';
 
-import { memo, useCallback, useMemo, useState } from 'react';
+import { memo, useCallback, useEffect, useMemo, useState } from 'react';
 import type { LinkProps } from '@tanstack/react-router';
 
 import { MdChevronRight } from 'react-icons/md';
@@ -64,6 +64,8 @@ const columns: ColumnDef<TestByCommitHash>[] = [
   },
   {
     accessorKey: 'status',
+    filterFn: (row, columnId, filterValue) =>
+      getStatusGroup(row.getValue(columnId)) === filterValue,
     header: ({ column }): JSX.Element =>
       TableHeader({
         column: column,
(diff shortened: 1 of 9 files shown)


dashboard/src/components/BootsTable/BootsTable.tsx
```diff
@@ -285,7 +292,11 @@ export function BootsTable({
       <BaseTable headerComponents={tableHeaders}>
         <TableBody>{tableRows}</TableBody>
       </BaseTable>
-      <PaginationInfo table={table} data={data} intlLabel="treeDetails.boots" />
+      <PaginationInfo
+        table={table}
+        data={table.getPrePaginationRowModel().rows.map(row => row.original)}
```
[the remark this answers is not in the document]
Francisco2002: `rawData.tests` will pass all the data from the pagination component, but I need the filtered data only. But I can get the table props of the component to get this information, so, data props is unnecessary and I remove it.

dashboard/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -297,14 +305,18 @@ export function BuildsTable({
...
+                : ''
+            }
             navigationLogsActions={navigationLogsActions}
           />
         </Sheet>
       </BaseTable>
       <PaginationInfo
         table={table}
-        data={data}
+        data={table.getPrePaginationRowModel().rows.map(row => row.original)}
```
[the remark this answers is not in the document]
Francisco2002: ditto about unnecessary prop

dashboard/src/components/Table/PaginationInfo.tsx
```diff
@@ -33,28 +35,26 @@ interface IPaginationInfo {
...
   intlLabel,
 }: IPaginationInfo): JSX.Element {
   const buttonsClassName = 'text-blue font-bold';
 
+  const filteredRowModel = table.getFilteredRowModel();
+  const countData = useMemo(
+    () => filteredRowModel.rows.length,
+    [filteredRowModel],
+  );
+
```
MarceloRobert, line 52: you can skip filteredRowModel and get the countData _length_ directly, so you can also use this in the span below at line 67

Francisco2002: if I understood correctly, your suggestion is to do something like this

Francisco2002: If I understood correctly, your suggestion is to do something like this

`const countData = useMemo(
    () => table.getFilteredRowModel().rows.length,
    [table],
  );`

This way don't update the page count correctly

MarceloRobert: As we discussed, it makes sense to do how you're doing, it is also done that way in other places. Also, my mistake, I confused countData with the old data.length, the way you're doing is right.

MarceloRobert (approved):
Great work

PR #549, opened 2024-11-14
Show full text in issues

### LINKED ISSUE #547: [Issues]: Show full text on mouseover
Some issues have ellipsis because they are too big, show the fulltext on a tooltip

Francisco2002:
Add tooltip in issues list to show full text.

Close #547

### DIFF
--- a/dashboard/src/components/ListingItem/ListingItem.tsx
+++ b/dashboard/src/components/ListingItem/ListingItem.tsx
@@ -1,8 +1,11 @@
 import classNames from 'classnames';
 
-import { useCallback } from 'react';
+import { useCallback, Fragment, memo } from 'react';
+
+import type { PropsWithChildren } from 'react';
 
 import ColoredCircle from '../ColoredCircle/ColoredCircle';
+import { Tooltip, TooltipContent, TooltipTrigger } from '../Tooltip';
 
 export interface IListingItem {
   leftIcon?: React.ReactNode;
@@ -14,6 +17,7 @@ export interface IListingItem {
   hasBottomBorder?: boolean;
   showNumber?: boolean;
   onClick?: (item: string) => void;
+  tooltip?: string;
 }
 
 export enum ItemType {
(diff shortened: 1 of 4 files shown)


dashboard/src/components/ListingItem/ListingItem.tsx
```diff
@@ -90,20 +93,34 @@ const ListingItem = ({
...
+            <span className="truncate text-sm text-black">{text}</span>
+          </WrapperComponent>
+        </TooltipTriggerComponent>
+
+        <TooltipContent>
+          <span className="text-sm text-black">{text}</span>
+        </TooltipContent>
+      </div>
+    </Tooltip>
   );
```
lfjnascimento, line 124: Nit: the less html/components we render the better, so i would only use the `Tooltip`, `TooltipTrigger`  and `TooltipContent` components when they are really needed.

dashboard/src/components/ListingItem/ListingItem.tsx
```diff
@@ -14,6 +15,7 @@ export interface IListingItem {
   hasBottomBorder?: boolean;
   showNumber?: boolean;
   onClick?: (item: string) => void;
+  tooltip?: boolean;
```
lfjnascimento, line 18: It will be more flexible if the tooltip is the text itself, that way we can pass whatever we want to show

lfjnascimento (approved):
LGTM

dashboard/src/components/ListingItem/ListingItem.tsx
```diff
@@ -89,21 +94,45 @@ const ListingItem = ({
     <></>
   );
 
+  const TooltipComponent = ({ children }: PropsWithChildren): JSX.Element => {
+    return (
+      <Tooltip>
+        <div className="flex">
+          <TooltipTrigger className="overflow-hidden">
```
WilsonNet, line 101: I think we probably want an `asChild` here so it doesn't render as an html button

WilsonNet (approved):
looks good, please try the `asChild` before merging

PR #553, opened 2024-11-14
Feat/hd filter trees

lfjnascimento:


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -101,24 +118,26 @@ def generate_test_dict():
     }
 
 
+@method_decorator(csrf_exempt, name='dispatch')
```
Francisco2002, line 121: What is this line for?

WilsonNet, line 121: add a TODO to remove this, we need to figure out a way to create post without making vulnerabilities, even though they are harmless in this particular example

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -157,7 +171,49 @@ def sanitize_records(self, records):
...
+            "build__checkout__tree_name",
+            "build__checkout__git_repository_branch",
+            "build__checkout__git_repository_url",
+        ]
+
+        trees_query_set = Tests.objects.filter(
+            environment_compatible__contains=[hardware_id],
+            origin=origin,
+            start_time__gte=start_datetime,
+            start_time__lte=end_datetime,
```
WilsonNet, line 187: this filter should happen in the checkout table only
in case another batch of tests will be released for those trees
that way will mimic better the behavior in the tree monitoring

MarceloRobert (approved):
LGTM

WilsonNet (approved):
nice job, wait for `[feat/implement-timelimit](https://github.com/kernelci/dashboard/tree/feat/implement-timelimit)` to be merged

PR #557, opened 2024-11-18
feature: make hardware listing timestamp based

WilsonNet:
# Description
Due to our limitation with URLs we can't pass all the data there, so we went for a timestamp based approach so in ten years we can display something similar that the user saw today

## How to test
Just go to hardware listing, enter any hardware details page, it should be sending the timestamps to the backend


dashboard/src/utils/date.ts
```diff
@@ -0,0 +1,9 @@
+export const dateObjectToTimestampInSeconds = (date: Date): number => {
+  const MILLISECONDS_IN_ONE_SECOND = 1000;
+  return Math.floor(date.getTime() / MILLISECONDS_IN_ONE_SECOND);
+};
+
+export const daysToSeconds = (days: number): number => {
+  const SECONDS_IN_ONE_DAY = 86400;
+  return days * SECONDS_IN_ONE_DAY;
+};
```
lfjnascimento, line 9: declare the constants in the top level scope

lfjnascimento (approved):
We should try to use the snake case format in python files.
pre-approving, good job!

PR #562, opened 2024-11-21
fix: missing required tableFilter bug

### LINKED ISSUE #554: Test details: remove query parameter requirement

lfjnascimento:
- move `tableFilter` from `/tree/$treeId` route to` /tree/$treeId/` page since it is not used by all `/tree/$treeId` children

- make `tableFilter` be an optional value by using zTableFilterInfoValidator which will correctly parse undefined to the default value for table filters

Close #554


MarceloRobert (approved):
Cool, there has to be a catch for the parameters of the object and for the object itself, nice work

murilx (approved):

PR #567, opened 2024-11-21
Feat/hardware by trees checkout

WilsonNet:
# Description
Make the hardware listing accounts for heads of the trees only
This also fix the logic to account for builds
## How to test 
Go to the hardware listing page

Enter in any of the hardware the details and see how the count is more precise

**Note: It is not exactly precise, if this does not get solved in time another ticket will be created**


lfjnascimento:
front-end is making two requests, one `slow` and another `fast` but we are not using modes for this endpoint

lfjnascimento (approved):
LGTM

Francisco2002 (approved):

PR #568, opened 2024-11-21
fix: alter log-downloader error status code to 400

### LINKED ISSUE #559: [bug]: log-downloader endpoint should return 400 instead of 500

murilx:
Close #559

## How to test
- Run the backend (after exporting the necessary environment variables) with `poetry run python3 manage.py runserver`
- Send and GET request to `api/log-downloader` and check if the status code is 400 with `curl 'localhost:8000/api/log-downloader/?log_download_url=example.com' -o - -I`


backend/kernelCI_app/views/logDownloaderView.py
```diff
@@ -58,6 +58,6 @@ def get(self, request):
 
         error_message = parsed_data.get("error")
         if error_message:
-            return JsonResponse({"error": error_message}, status=500)
+            return JsonResponse({"error": error_message}, status=400)
```
lfjnascimento, line 61: would be a good ideia to use `getErrorResponseBody`

lfjnascimento (approved):

backend/kernelCI_app/views/logDownloaderView.py
murilx: Done

WilsonNet (approved):
lgtm

PR #570, opened 2024-11-22
fix: random fixes

lfjnascimento:
- fix: some files in src/api was following the component naming convention
- fix: make ITableHeader use generics
- fix: TableHeader component was being used as function 
- fix: remove unnecessary memoization from TreeTable
- fix: move memo(InputTime) to module scope
- fix: memoize tree details tab array
- fix: remove unused accordionData type


MarceloRobert (commented):
Don't forget the `tooltipId` of some of the table headers

Francisco2002 (approved):

PR #576, opened 2024-11-22
feat(IssuesList): add a counter for unknown issues

### LINKED ISSUE #544: [Issues]: Show unknown count in issues cards
Create an unknown count that is the number of failed tests that have no linked issues

Discussion link: https://discord.com/channels/1245820301053530313/1301896040349433957/1306244286316019814

murilx:
Add a counter containing the number of failed tests with no linked issues. To do it, add an optional parameter to `IssuesList` that accept the number of total failed tests/builds/boots. If the parameter is not present the behaviour of the Issues List is equal as it was before, otherwise it adds a new line to the list containing the number of unknown issues.

Close #544

### DIFF
--- a/backend/kernelCI_app/views/hardwareDetailsView.py
+++ b/backend/kernelCI_app/views/hardwareDetailsView.py
@@ -123,11 +123,19 @@ def generate_test_dict():
         "failReasons": defaultdict(int),
         "configs": defaultdict(lambda: defaultdict(int)),
         "issues": {},
+        "failedWithUnknownIssues": 0,
     }
 
 
 def is_record_in_tree_head(record, tree):
     return record["build__checkout__git_commit_hash"] == tree["headGitCommitHash"]
+
+
+def update_unknown_issues(tests_or_boots, status, builds, is_build_processed, build_is_valid):
+    if status == "FAIL":
+        tests_or_boots["failedWithUnknownIssues"] += 1
+    if not is_build_processed and build_is_valid is False:
+        builds["failedWithUnknownIssues"] += 1
 
 
 # disable django csrf protection https://docs.djangoproject.com/en/5.0/ref/csrf/
@@ -181,7 +189,7 @@ class HardwareDetails(View):
 
     def sanitize_records(self, records, selected_trees, filters):
         processed_builds = set()
-        builds = {"items": [], "issues": {}}
+        builds = {"items": [], "issues": {}, "failedWithUnknownIssues": 0}
 
         filters_map = self.get_filters(filters)
 
@@ -223,7 +231,8 @@ class HardwareDetails(View):
                     tests_or_boots["archSummary"][archKey] = archSummary
                 archSummary["status"][status] += 1
 
-            if build_id not in processed_builds:
+            is_build_processed = build_id in processed_builds
+            if not is_build_processed:
                 processed_builds.add(build_id)
                 build = get_build(r, current_tree["index"])
 
@@ -235,6 +244,10 @@ class HardwareDetails(View):
                 update_issues(builds["issues"], currentIssue)
                 if not jump_test:
                     update_issues(tests_or_boots["issues"], currentIssue)
+            else:
+                update_unknown_issues(
+                    tests_or_boots, r["status"], builds, is_build_processed, r["build__valid"]
+                )
 
         builds["summary"] = create_details_build_summary(builds["items"])
         properties2List(builds, ["issues"])
--- a/dashboard/src/components/Cards/IssuesList.tsx
+++ b/dashboard/src/components/Cards/IssuesList.tsx
@@ -2,6 +2,8 @@ import { memo } from 'react';
 
 import { Link } from '@tanstack/react-router';
 
+import { useIntl } from 'react-intl';
+
 import type { IBaseCard } from '@/components/Cards/BaseCard';
 import BaseCard from '@/components/Cards/BaseCard';
 import { DumbListingContent } from '@/components/ListingContent/ListingContent';
@@ -13,11 +15,20 @@ import type { TIssue } from '@/types/general';
 
 interface IIssuesList {
   issues: TIssue[];
+  failedWithUnknownIssues?: number;
   title: IBaseCard['title'];
 }
 
-const IssuesList = ({ issues, title }: IIssuesList): JSX.Element => {
-  const hasIssue = issues.length > 0;
+const IssuesList = ({
+  issues,
+  failedWithUnknownIssues,
+  title,
+}: IIssuesList): JSX.Element => {
+  const intl = useIntl();
+  failedWithUnknownIssues = failedWithUnknownIssues
+    ? failedWithUnknownIssues
+    : undefined;
+  const hasIssue = issues.length > 0 || failedWithUnknownIssues;
 
   const titleElement = (
     <span>
@@ -26,7 +37,7 @@ const IssuesList = ({ issues, title }: IIssuesList): JSX.Element => {
         <ColoredCircle
           className="ml-2 font-normal"
           backgroundClassName={ItemType.Error}
-          quantity={issues.length}
+          quantity={issues.length + (failedWithUnknownIssues ? 1 : 0)}
         />
       )}
     </span>
@@ -49,6 +60,12 @@ const IssuesList = ({ issues, title }: IIssuesList): JSX.Element => {
           </WrapperLink>
         );
       })}
+      {failedWithUnknownIssues && (
+        <ListingItem
+          unknown={failedWithUnknownIssues}
+          text={intl.formatMessage({ id: 'global.unknown' })}
+        />
+      )}
     </DumbListingContent>
   );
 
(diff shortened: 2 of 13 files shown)


dashboard/src/components/Cards/IssuesList.tsx
```diff
@@ -49,6 +56,13 @@ const IssuesList = ({ issues, title }: IIssuesList): JSX.Element => {
           </WrapperLink>
         );
       })}
+      {totalFails !== undefined && (
+        <ListingItem
+          unknown={totalFails - knownLinkedIssues}
+          // TODO: Change this to alter the message according to the selected language
+          text={'Unknown'}
```
Francisco2002, line 63:
```suggestion
          text={intl.formatMessage({ id: 'global.unknown' })}
```

dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
```diff
@@ -106,6 +108,16 @@ const TestsTab = ({ reqFilter }: TestsTabProps): JSX.Element => {
     );
   }
 
+  const failedCount = groupStatus({
+    doneCount: data.testStatusSummary.DONE ?? 0,
+    errorCount: data.testStatusSummary.ERROR ?? 0,
+    failCount: data.testStatusSummary.FAIL ?? 0,
```
Francisco2002, line 114: You can get the faildCount directly of `data.testStatusSummary.FAIL` without use the `groupStatus` function

dashboard/src/components/Cards/IssuesList.tsx
murilx: Done!

dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
murilx: That's true. After your comment I read the groupStatus function more carefully and figured I didn't need this in this case to get the fail count! However the new version of the code don't use this any more.

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -218,17 +226,20 @@ def sanitize_records(self, records, selected_trees, filters):
                 archSummary["status"][status] += 1
 
             if build_id not in processed_builds:
-                processed_builds.add(build_id)
                 build = get_build(r, current_tree["index"])
+                processed_builds.add(build_id)
```
lfjnascimento, line 230: why change this line? usually when opening a PR we want to keep only significant changes and minimize noise

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -123,13 +123,21 @@ def generate_test_dict():
...
 
 
 def is_record_in_tree_head(record, tree):
     return record["build__checkout__git_commit_hash"] == tree["headGitCommitHash"]
 
 
+def update_unknown_issues(tests_or_boots, builds, build_is_valid, status):
+    if status == "FAIL":
+        tests_or_boots["failedWithUnkownIssues"] += 1
+    if build_is_valid is not None and not build_is_valid:
```
lfjnascimento, line 137: would be simpler to do `if build_is_valid == False`

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -218,17 +226,20 @@ def sanitize_records(self, records, selected_trees, filters):
                 archSummary["status"][status] += 1
 
             if build_id not in processed_builds:
-                processed_builds.add(build_id)
                 build = get_build(r, current_tree["index"])
+                processed_builds.add(build_id)
 
                 if len(filters_map['build']) == 0 or self.pass_in_filters(build, filters_map['build']):
                     builds["items"].append(build)
+                    build_is_valid = build["valid"]
```
lfjnascimento, line 234: You are using this variable outside of this if scope, which may lead to a `NameError`, IMO we don't need to create this var and just use build["valid"] since you are using it only once

murilx: From my understanding build["valid"] could also be outside of scope (since it's created inside an if) and even if not (I could move it to be outside of the if), build is not always added to the builds["items"] list, that's why I'm defining this variable inside the if statement that appends the build to build["items"] (if the build is not part of the final list, then I believe it shouldn't count as a failed build with unknown issue even if the others constraints apply). Is my reasoning correct? If so, I think that defining `build_is_valid = None` before both if statements would work

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -175,7 +183,7 @@ def pass_in_filters(self, data, filters):
 
     def sanitize_records(self, records, selected_trees, filters):
         processed_builds = set()
-        builds = {"items": [], "issues": {}}
+        builds = {"items": [], "issues": {}, "failedWithUnkownIssues": 0}
```
lfjnascimento, line 186: `failedWithUnkownIssues` is related to `tests` and `boots` not builds

backend/kernelCI_app/views/hardwareDetailsView.py
lfjnascimento: you are right, it could be `record["build__valid"]`

backend/kernelCI_app/views/hardwareDetailsView.py
murilx: Done

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -123,13 +123,21 @@ def generate_test_dict():
...
 
 def is_record_in_tree_head(record, tree):
     return record["build__checkout__git_commit_hash"] == tree["headGitCommitHash"]
 
 
+def update_unknown_issues(tests_or_boots, builds, build_is_valid, status):
+    if status == "FAIL":
+        tests_or_boots["failedWithUnkownIssues"] += 1
+    if build_is_valid is not None and not build_is_valid:
+        builds["failedWithUnkownIssues"] += 1
```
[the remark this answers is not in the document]
murilx: Removed every piece of code that would implement this counter for builds, now it's only for tests and boots

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -123,13 +123,19 @@ def generate_test_dict():
...
     }
 
 
 def is_record_in_tree_head(record, tree):
     return record["build__checkout__git_commit_hash"] == tree["headGitCommitHash"]
 
 
+def update_unknown_issues(tests_or_boots, status):
+    if status == "FAIL":
+        tests_or_boots["failedWithUnknownIssues"] += 1
```
Francisco2002, line 136: the key `failedWithUnknownIssues` isn't the same that you put in generate test dict `(failedWithUnkownIssues)` - line 126

murilx: Typo. Already changed it in the generate_test_dict function.

Francisco2002 (approved):

dashboard/src/components/Cards/IssuesList.tsx
```diff
@@ -13,26 +15,32 @@ import type { TIssue } from '@/types/general';
...
           backgroundClassName={ItemType.Error}
-          quantity={issues.length}
+          quantity={issues.length + (failedWithUnknownIssues ? 1 : 0)}
         />
       )}
     </span>
   );
 
-  const contentElement = !hasIssue ? (
+  const contentElement = !failedWithUnknownIssues ? (
```
lfjnascimento, line 43: dont change this just make` const hasIssue = issues.length > 0 || failedWithUnknownIssues > 0;`

dashboard/src/components/Cards/IssuesList.tsx
```diff
@@ -13,26 +15,32 @@ import type { TIssue } from '@/types/general';
...
+  title,
+}: IIssuesList): JSX.Element => {
+  const intl = useIntl();
   const hasIssue = issues.length > 0;
 
   const titleElement = (
     <span>
       {title}
-      {hasIssue && (
+      {(hasIssue || failedWithUnknownIssues) && (
```
lfjnascimento, line 33: just change the definition of `hasIssue` to consider `failedWithUnknownIssues`

dashboard/src/components/Cards/IssuesList.tsx
murilx: In this case, `const hasIssue = issues.length > 0` could also just be `const hasIssue = issues.length`, is this change acceptable or does it just make the code less readable?

murilx: After more thought I don't think I could do just `issues,length` because that could show a random 0 instead of my component

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -217,7 +225,8 @@ def sanitize_records(self, records, selected_trees, filters):
                     tests_or_boots["archSummary"][archKey] = archSummary
                 archSummary["status"][status] += 1
 
-            if build_id not in processed_builds:
+            build_is_not_processed = build_id not in processed_builds
```
lfjnascimento, line 228: nit: It is good practice to keep flag names in positive format, such as `is_build_precesses`

murilx: Done

dashboard/src/components/Cards/IssuesList.tsx
murilx: Done

lfjnascimento (approved):
LGTM

Issue #582, opened 2024-11-25
Hardware Monitor: Filter boots and tests by path


MarceloRobert:
Similar to #524 , add backend filtering by path in HardwareDetails

PR #602, opened 2024-11-25
Feat: HD path filter and patches

### LINKED ISSUE #578: Fix (path filter): frontend filtering, path input sync, commit graph
In Hardware Details and Tree Details page,

- [x] when the path filter is removed from the filter list, the path input should be cleared
- [x] the global filter of the table (frontend) shouldn't be applied automatically if the diffFilter has been removed
- [x] the commit graph should consider boot records when applying the path filters

MarceloRobert:
- Splits path filter between boots and tests, now one doesn't affect the other.
- Adds filtering to hardwareDetails page;
- Syncs filter listing and input field, after removing the path filter from the filter list the input field will also clear;
- Fixes boot commit history graph not using the path filter.

Also refactors the sorting exhaustive-deps useMemo dependency just so eslint can complain if a useMemo dependency is missing

## How to test

Go to a tree details page with boot and/or test data, scroll down and use the searchbox alongside the path column. Check the filtering of the table and the commit history graph. Compare this with the previous version currently on staging.
Also go to a hardware details page with boot and/or test data and see the same behavior as in tree details (except for the commit history graph, which isn't implemented yet).

Closes #578  and #582

### DIFF
--- a/backend/kernelCI_app/views/hardwareDetailsView.py
+++ b/backend/kernelCI_app/views/hardwareDetailsView.py
@@ -168,7 +168,13 @@ class HardwareDetails(View):
             field = currentFilter.get("field")
             value = currentFilter.get("value")
 
-            if data[field] not in value:
+            if (
+                field == "path"
+                and value[0] not in data[field]
+            ):
+                return False
+
+            if (data[field] not in value):
                 return False
 
         return True
--- a/dashboard/src/components/BootsTable/BootsTable.tsx
+++ b/dashboard/src/components/BootsTable/BootsTable.tsx
@@ -330,7 +330,13 @@ export function BootsTable({
     return groupHeaders.map(header => {
       const headerComponent = header.isPlaceholder
         ? null
-        : flexRender(header.column.columnDef.header, header.getContext());
+        : // the header must change the icon when sorting changes,
+          // but just the column dependency won't trigger the rerender
+          // so we pass an unused sorting prop here to force the useMemo dependency
+          flexRender(header.column.columnDef.header, {
+            ...header.getContext(),
+            sorting,
+          });
       return (
         <TableHead key={header.id} className="border-b px-2 font-bold">
           {header.id === 'path' ? (
@@ -350,9 +356,7 @@ export function BootsTable({
         </TableHead>
       );
     });
-    // TODO: remove exhaustive-deps and change memo  (all tables)
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [groupHeaders, sorting]);
+  }, [groupHeaders, intl, onSearchChange, sorting]);
 
   const modelRows = table.getRowModel().rows;
 
--- a/dashboard/src/components/BootsTable/BootsTable.tsx
+++ b/dashboard/src/components/BootsTable/BootsTable.tsx
@@ -118,7 +118,8 @@ interface IBootsTable {
   filter: TestsTableFilter;
   getRowLink: (testId: TestHistory['id']) => LinkProps;
   onClickFilter: (newFilter: TestsTableFilter) => void;
-  updatePathFilter: (pathFilter: string) => void;
+  updatePathFilter?: (pathFilter: string) => void;
+  currentPathFilter?: string;
 }
 
 const TableCellComponent = ({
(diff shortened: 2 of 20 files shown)


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -168,8 +168,12 @@ def pass_in_filters(self, data, filters):
...
             value = currentFilter.get("value")
 
-            if data[field] not in value:
-                return False
+            if (field == "path"):
+                if (value[0] not in data[field]):
+                    return False
+            else:
+                if data[field] not in value:
+                    return False
```
MarceloRobert, line 176: this if else tree could be cleaner

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -255,14 +258,17 @@ export function TestsTable({
     [filterCount, intl, filter],
   );
 
-  // TODO: there should be a filtering for the frontend before the backend AND that filtering should consider the individual tests inside each test "batch" (the data from individualTestsTables), not only the rows of the external table
+  // TODO: there should be a filtering for the frontend before the backend
+  // that frontend filtering should consider the individual tests inside each test batch (the data from individualTestsTables),
+  // not only the rows of the external table (same for BootsTable)
```
WilsonNet, line 263: We are not using this approach anymore, we will just preserve the data, add a spinner until the fresh data is retrieved

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -255,14 +258,17 @@ export function TestsTable({
...
   const onSearchChange = useCallback(
     (e: React.ChangeEvent<HTMLInputElement>) => {
       if (e.target.value !== undefined && updatePathFilter) {
         updatePathFilter(e.target.value);
       }
-      // TODO: remove this frontend filter when the hardwareDetails backend filtering gets in place
-      table.setGlobalFilter(String(e.target.value));
+      if (updatePathFilter === undefined) {
+        table.setGlobalFilter(String(e.target.value));
+      }
```
WilsonNet, line 271: do we need to set the global filter? since the filter is being added in the backend

MarceloRobert: In this case it is being used for tables such as within the buildDetails page, which doesn't have backend filtering for the page so the frontend filtering is being used

dashboard/src/components/BootsTable/BootsTable.tsx
```diff
@@ -333,6 +333,8 @@ export function BootsTable({
         : flexRender(header.column.columnDef.header, header.getContext());
       return (
         <TableHead key={header.id} className="border-b px-2 font-bold">
+          {/* TODO: remove this hidden div, this is just to force memoization reset when sorting changes (all tables) */}
+          <div className="hidden">{sorting.toString()}</div>
```
lfjnascimento, line 343: Make TableHead explicitly dependent on sorting

dashboard/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -295,14 +295,14 @@ export function BuildsTable({
     return groupHeaders.map(header => {
       return (
         <TableHead key={header.id} className="border-b px-0 font-bold">
+          {/* TODO: remove this hidden div, this is just to force memoization reset when sorting changes (all tables) */}
+          <div className="hidden">{sorting.toString()}</div>
```
Francisco2002, line 299: apply the same logic of BootsTable

lfjnascimento (approved):

dashboard/src/components/Table/TableHeader.tsx
```diff
@@ -58,7 +59,10 @@ export function TableHeader({
   intlKey,
   intlDefaultMessage,
   tooltipId,
+  sorting,
 }: ITableHeader): JSX.Element {
+  console.error(sorting); // <- use the sorting the function below instead of column.getIsSorted
```
Francisco2002, line 64: remove this

Francisco2002 (approved):
nice!!

PR #609, opened 2024-11-26
Refactor: intl cleaning

MarceloRobert:
Renaming some intl keys and removing unused ones.

A lot of keys were being used by hardware monitor with the key from tree monitor, while others had the key for specific tabs or pages but were being used in unrelated components.


WilsonNet (approved):
looks good just left a question but it is up to you to decide

dashboard/src/types/tree/TreeDetails.tsx
```diff
@@ -96,18 +96,16 @@ export type TTreeTestsFullData = {
...
+  'global.builds' | 'global.boots' | 'global.tests'
 >;
 
 const possibleTabs = [
-  'treeDetails.builds',
-  'treeDetails.boots',
-  'treeDetails.tests',
+  'global.builds',
+  'global.boots',
+  'global.tests',
```
lfjnascimento, line 105: this is being used as a value and not as an i18key so we can just use like this:
```suggestion
  'builds',
  'boots',
  'tests',
```

MarceloRobert: In this case the PossibleTabs type is an Extract of MessagesKey, so this variable needs to be an intl key unless the typing is also changed

lfjnascimento (approved):

PR #613, opened 2024-11-27
feat: add new routes for tests and builds

### LINKED ISSUE #555: Test details: remove hardware/tree requirement to enter the page
Currently, you always need to access /tree/\<treeId\> or /hardware/\<hardwareId\> to acess /test/\<testId\>, it would be interesting to be able to go to a test details page directly, without passing through trees or hardwares

murilx:
Add two new routes /test/$testId and /build/$buildId, removing the need of accessing tests or builds through /hardware or /tree. Also, redirects the old routes to the new ones.

Closes #555

### DIFF
--- a/dashboard/src/components/BuildDetails/BuildDetails.tsx
+++ b/dashboard/src/components/BuildDetails/BuildDetails.tsx
@@ -36,7 +36,7 @@ const BlueFolderIcon = (): JSX.Element => (
 );
 
 interface BuildDetailsProps {
-  breadcrumb: JSX.Element;
+  breadcrumb?: JSX.Element;
   buildId?: string;
   onClickFilter: (filter: TestsTableFilter) => void;
   tableFilter: TableFilter;
--- a/dashboard/src/components/LinkWithIcon/LinkWithIcon.tsx
+++ b/dashboard/src/components/LinkWithIcon/LinkWithIcon.tsx
@@ -9,6 +9,7 @@ export interface ILinkWithIcon {
   linkText?: string | ReactElement;
   link?: string;
   icon?: ReactElement;
+  linkComponent?: ReactElement;
   onClick?: () => void;
 }
 
@@ -17,6 +18,7 @@ const LinkWithIcon = ({
   linkText,
   icon,
   link,
+  linkComponent,
   onClick,
 }: ILinkWithIcon): JSX.Element => {
   const WrapperLink = link ? 'a' : 'div';
@@ -25,16 +27,18 @@ const LinkWithIcon = ({
       <span className="font-bold">
         <FormattedMessage id={title} />
       </span>
-      <WrapperLink
-        className="flex flex-row items-center gap-1"
-        href={link}
-        target="_blank"
-        rel="noreferrer"
-        onClick={onClick}
-      >
-        {linkText}
-        {icon}
-      </WrapperLink>
+      {linkComponent ?? (
+        <WrapperLink
+          className="flex flex-row items-center gap-1"
+          href={link}
+          target="_blank"
+          rel="noreferrer"
+          onClick={onClick}
+        >
+          {linkText}
+          {icon}
+        </WrapperLink>
+      )}
     </div>
   );
 };
--- a/dashboard/src/components/TestDetails/TestDetails.tsx
+++ b/dashboard/src/components/TestDetails/TestDetails.tsx
@@ -7,12 +7,18 @@ import { useMemo } from 'react';
 import { PiComputerTowerThin } from 'react-icons/pi';
 import { MdFolderOpen } from 'react-icons/md';
 
+import { Link, useRouterState, useSearch } from '@tanstack/react-router';
+
+import { FiLink } from 'react-icons/fi';
+
 import { truncateBigText } from '@/lib/string';
 import type { TTestDetails } from '@/types/tree/TestDetails';
 import { Sheet, SheetTrigger } from '@/components/Sheet';
 import { useTestDetails, useTestIssues } from '@/api/testDetails';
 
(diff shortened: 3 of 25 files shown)


dashboard/src/components/BuildDetails/BuildDetails.tsx
```diff
@@ -193,7 +193,7 @@ const BuildDetails = ({
   return (
     <ErrorBoundary FallbackComponent={UnexpectedError}>
       <Sheet>
-        {breadcrumb}
+        {breadcrumb ?? null}
```
WilsonNet, line 196: this is the same as just `{breadcrumb}`

dashboard/src/routes/tree/$treeId/build/$buildId/index.tsx
```diff
@@ -1,15 +1,21 @@
...
   tableFilter: zTableFilterInfoValidator,
 });
 
 export const Route = createFileRoute('/tree/$treeId/build/$buildId/')({
-  component: () => <TreeBuildDetails />,
   validateSearch: buildDetailsSearchSchema,
+  loader: async ({ params, location }) => {
+    throw redirect({
+      to: '/build/$buildId',
+      search: location.search,
```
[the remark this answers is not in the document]
murilx: Do you mean doing something like `search: previousSearch => { previousSearch }`as we do in other parts of the code?

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -22,22 +24,25 @@ import IssueSection from '../Issue/IssueSection';
...
-  const buildDetailsLink =
-    `${window.location.origin}` +
-    `/${context}/${test.git_commit_hash}/build/${test.build_id}` +
-    `${window.location.search}`;
+  let buildDetailsLink = '';
+  if (historyState.from !== undefined) {
+    buildDetailsLink =
+      `${window.location.origin}` +
+      `/${historyState.from}/${historyState.id}/build/${test.build_id}` +
+      `${window.location.search}`;
```
WilsonNet, line 39: use tnastack router for the search params

murilx: Since I need the search params string in this case (the way the component is built it requires a link string to be passed to it) isn't location.search acceptable? As far as I searched I didn't find a way for tanstack router to return me the search params string. If location.search is not acceptable, a solution would be to implement a `stringify` function to get the search params string from the object returned by `useSearch` or use a library like [query-string](https://github.com/sindresorhus/query-string) that seems to do the job

dashboard/src/components/BuildDetails/BuildDetails.tsx
murilx: Done

dashboard/src/components/TestDetails/TestDetails.tsx
WilsonNet: I'm seeing the root cause, we had unnecessary `window.location`  here, we don't need that, build details and test details are internal urls, they can be accesses with `/` 
and aside from that we can just use regular tanstack links with the callback search params

dashboard/src/components/LinkWithIcon/LinkWithIcon.tsx
```diff
@@ -25,16 +27,18 @@ const LinkWithIcon = ({
...
-        className="flex flex-row items-center gap-1"
-        href={link}
-        target="_blank"
-        rel="noreferrer"
-        onClick={onClick}
-      >
-        {linkText}
-        {icon}
-      </WrapperLink>
+      {linkComponent ?? (
```
Francisco2002, line 30: Why did you use this props?

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -22,22 +28,42 @@ import IssueSection from '../Issue/IssueSection';
...
+        search={searchParams}
+        className="flex flex-row items-center gap-1"
+        target="_blank"
+        rel="noreferrer"
+      >
+        {truncateBigText(test.build_id)}
+        <FiLink className="text-blue" />
+      </Link>
+    );
+  }, [historyState, test.build_id, searchParams]);
```
Francisco2002, line 66: Why you don't use `LinkWithIcon` component and pass `linkText=truncateBigText(test.build_id)` and `icon=<FiLink className="text-blue" />`?

murilx: As stated by Wilson in a previous comment, it would be better to use tanstack router to get the search params. LinkWithIcon does not accept the object returned by useSearch as a prop (since LinkWithIcon is basically an 'a' HTML element) so the solution was to use tanstack router `Link` and modify the LinkWithIcon component to render this Link in case it's passed to it instead of 'a' (or 'div')

murilx: I was originally using LinkWithIcon but that would mean I'd have to build the link using things like `location.search` for example

MarceloRobert (commented):
Seems to be working fine, but the header for the entire page becomes "Unknown" if you access a build or tests details page without information about tree or hardware, maybe showing Build/Test would be better?

dashboard/src/components/LinkWithIcon/LinkWithIcon.tsx
murilx: I'm using this prop so that I'm able to pass a custom Link component (in this case a tanstack router `Link` component) to the LinkWithIcon component, since it only render "simple" links (as an 'a' HTML Element)

MarceloRobert (approved):
Nice work

Francisco2002 (approved):

PR #615, opened 2024-11-27
Feat: async filters

### LINKED ISSUE #585: Improve search by path: ui is a bit laggy
- Updates page only when data is available
- Show a spinner to indicate that a query is running

MarceloRobert:
Adds a loading circle and async loading for treeDetails and hardwareDetails filtering.

## How to test
Go to treeDetails page or hardwareDetails page and add any filter, the page won't show the "Loading" screen anymore, instead it will show a loading circle in the filterListing and will only update the page after the data has been fetched.
This works not only for the test path filtering but also for any other filter, although with the path filter it's more noticeable.

Closes #585

### DIFF
--- a/dashboard/src/components/FilterList/FilterList.tsx
+++ b/dashboard/src/components/FilterList/FilterList.tsx
@@ -3,9 +3,12 @@ import { IoClose } from 'react-icons/io5';
 import classNames from 'classnames';
 import { useIntl } from 'react-intl';
 
+import { LoadingCircle } from '@/components/ui/loading-circle';
+
 import { Button } from '../ui/button';
 
 export interface IFilterList {
+  isLoading: boolean;
   items: string[];
   onClickItem: (item: string, itemIdx: number) => void;
   onClickCleanAll: () => void;
@@ -72,6 +75,7 @@ const FilterItem = ({
 };
 
 const FilterList = ({
+  isLoading,
   items,
   onClickItem,
   onClickCleanAll,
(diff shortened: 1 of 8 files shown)


dashboard/src/components/FilterList/FilterList.tsx
```diff
@@ -97,7 +101,8 @@ const FilterList = ({
   }
 
   return (
-    <div className="flex flex-wrap gap-4">
+    <div className="flex flex-wrap items-center gap-4">
+      {isLoading && <LoadingCircle />}
```
lfjnascimento, line 105: Let's put the `LoadingCircle` at the end of the list, that way the items doesn't move when the loading appears or disappears.

lfjnascimento (approved):
Pre-approving

dashboard/src/components/FilterList/FilterList.tsx
WilsonNet: another solution is isLoading to be a prop for `LoadingCircle`  and use the visibility property https://tailwindcss.com/docs/visibility

WilsonNet (approved):
also pre-approving

Francisco2002 (approved):

dashboard/src/components/FilterList/FilterList.tsx
MarceloRobert: I'll put it at the end of the list because making it invisible would create too big of a padding before the filters

PR #625, opened 2024-11-28
fix: issues not being filtered

### LINKED ISSUE #545: [Issues]: The Issues card is not being affected by filters

WilsonNet:
# Description
It was believed that filter were not affecting Issues, but after further inspection we understood that it was just a misunderstanding of the fact that we have both global and per tab filters
- Add icons and tooltip for filters
- Fix possible bug that processed tests were still being counted
- Fix possible bag that all issues were being counted

## How to test
Go to details page and open filter tab

image

Closes #545


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -190,8 +203,24 @@ def pass_in_filters(self, data, filters):
...
+            return True
+
+        test_filter_is_empty = len(filters_map[table_test]) == 0
+        test_pass_in_filters = self.pass_in_filters(record, filters_map[table_test])
+        return not test_filter_is_empty and not test_pass_in_filters
+
     def sanitize_records(self, records, selected_trees, filters):
-        processed_builds = set()
+        processed_builds: Set[str] = set()
+        processed_test: Set[str] = set()
```
lfjnascimento, line 223: There is no need need to mark the processed tests since each record is a unique test

lfjnascimento: nit: we can rename the records variable to make this fact more explicit

lfjnascimento:
In the second commit you say `Now build issues is only updated when is relevant...` but I didn't see any change in the issue logic

lfjnascimento:
Nice work adding the Global filters and per-tab icons and captions !

WilsonNet:
`In the second commit you say Now build issues is only updated when is relevant... but I didn't see any change in the issue logic`

There is an `if` checking now to see if the build passed the filters, before there was only processed builds check

backend/kernelCI_app/views/hardwareDetailsView.py
WilsonNet: we are using build_id as the key for the record check, and it is only validating in the build handling logic
this add the test and validates inside the test handling logic

lfjnascimento: This is making no sense, and the prove is the fact that you are never populating `processe_test`.

dashboard/src/locales/messages/index.ts
```diff
@@ -76,11 +76,15 @@ export const messages = {
...
     'filter.filtering': 'Filtering',
+    'filter.globalFilter': 'Global filters',
+    'filter.globalFilterAllTabs': 'Global filters affects all the tabs',
     'filter.hardware': 'Hardware',
-    'filter.hardwareSubtitle': 'Please select one or more hardwares:',
+    'filter.hardwareSubtitle': 'Please select one an specific hardwares:',
     'filter.invalid': 'Please provide a time limit, in days, greater than 0',
     'filter.max': 'Max',
     'filter.min': 'Min',
+    'filter.onlySpecificTab': 'Only affects an specific tab',
```
MarceloRobert, line 86: should be "Only affects _a_ specifc tab"

lfjnascimento:
> `In the second commit you say Now build issues is only updated when is relevant... but I didn't see any change in the issue logic`
> 
> There is an `if` checking now to see if the build passed the filters, before there was only processed builds check

There is not true, the logic was checking the processed_builds and the filters, adding the `is_build_filtered_in` function was a nice touch but you can see in the diff that the logic still the same

dashboard/src/locales/messages/index.ts
```diff
@@ -76,11 +76,15 @@ export const messages = {
     'filter.configsSubtitle': 'Please select one or more configs:',
     'filter.durationSubtitle': 'Please select the duration range:',
     'filter.filtering': 'Filtering',
+    'filter.globalFilter': 'Global filters',
+    'filter.globalFilterAllTabs': 'Global filters affects all the tabs',
     'filter.hardware': 'Hardware',
-    'filter.hardwareSubtitle': 'Please select one or more hardwares:',
+    'filter.hardwareSubtitle': 'Please select one an specific hardwares:',
```
MarceloRobert, line 82: this phrase doesn't make a lot of sense and I think you should keep the "or more", maybe `Please select one or more specific hardwares`

dashboard/src/components/Filter/Drawer.tsx
```diff
@@ -81,11 +85,61 @@ export const DrawerSection = ({
...
+> = {
+  global: 'filter.globalFilterAllTabs',
+  tab: 'filter.onlySpecificTab',
+};
+
+export const FilterTypeIcon = ({ type }: FilterTypeIconProps): JSX.Element => {
+  return (
+    <Tooltip>
+      <TooltipTrigger>{Icons[type]}</TooltipTrigger>
+      <TooltipContent>
```
MarceloRobert, line 118: nit: IMO I think that a `font-normal` class here would be better

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -190,8 +203,24 @@ def pass_in_filters(self, data, filters):
...
+    def should_jump_test(
+        self,
+        table_test: Literal["boot", "test"],
+        record: Dict,
+        filters_map: Dict[str, List[Dict]],
+        processed_test: Set[str],
+    ) -> bool:
+        has_test_been_processed = record["id"] in processed_test
+        if has_test_been_processed:
+            return True
```
MarceloRobert, line 215: nit: I don't think there is a need for this one-use variable, you could make the if statement directly

WilsonNet:
> In the second commit you say Now build issues is only updated when is relevant... but I didn't see any change in the issue logic
> 
> There is an if checking now to see if the build passed the filters, before there was only processed builds check
> 
> Not true, the logic was checking the processed_builds and the filters, adding the is_build_filtered_in function was a nice touch but you can see in the diff that the logic still the same

That is a python issue, the if with filter was one indentation level higher

image

MarceloRobert (approved):
Only some nit comments from my part, pre-approving

backend/kernelCI_app/views/hardwareDetailsView.py
WilsonNet: oh got it, this is in an inner join from tests

dashboard/src/locales/messages/index.ts
WilsonNet: oh this was probably a missplaced copy and paste

lfjnascimento:
I also really liked the addition of typing, in my opinion we should do this all the time

lfjnascimento (approved):

PR #628, opened 2024-11-29
change: widen the modal and improve visibility

### LINKED ISSUE #595: Increase the width of the modal

murilx:
Increase the log sheet modal width and also improve the visibility of the log that's being shown by using a darker color to highlight the table row and by decreasing the opacity of the sheet overlay.

Closes #595

## Visual Reference

Before
image

After

Captura de tela de 2024-11-29 12-03-51
image

### DIFF
--- a/dashboard/src/components/Filter/CodeBlock.tsx
+++ b/dashboard/src/components/Filter/CodeBlock.tsx
@@ -72,7 +72,7 @@ const Code = ({
     <>
       <pre
         className={cn(
-          'w-full max-w-[calc(100vw_-_398px)] overflow-x-auto rounded-md bg-[#DDDDDD] p-4 font-mono text-[#767676]',
+          'w-full max-w-[100vw] overflow-x-auto rounded-md bg-[#DDDDDD] p-4 font-mono text-sm leading-4 text-[#767676]',
           className,
         )}
       >
--- a/dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
+++ b/dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
@@ -122,7 +122,7 @@ export const LogSheet = ({
     { enabled: !!logUrl },
   );
   return (
-    <SheetContent className="flex w-[25rem] flex-col sm:w-full sm:max-w-[44rem]">
+    <SheetContent className="flex flex-col sm:w-full sm:max-w-[clamp(650px,75vw,1300px)]">
       <SheetHeader className="mb-3">
         <SheetTitle className="text-[1.75rem]">
           <FormattedMessage id="logSheet.title" />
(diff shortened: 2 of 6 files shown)


dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
```diff
@@ -122,7 +122,7 @@ export const LogSheet = ({
     { enabled: !!logUrl },
   );
   return (
-    <SheetContent className="flex w-[25rem] flex-col sm:w-full sm:max-w-[44rem]">
+    <SheetContent className="flex flex-col sm:w-full sm:max-w-[clamp(44rem,75rem,100vw)]">
```
lfjnascimento, line 125: `rem` means the font-size of the root element, so it is not the proper unit to use in this case. It makes more sense to use `vw` for the preferred value (the second argument) and `px` for the minimum and maximum. The maximum should be the maximum width of the `SheetContent` that still leaves room for the left sidebar + the page content (at the background).

lfjnascimento:
The ticket also asks to check if it is possible to decrease the font size of the log to fit more lines which I think we can do. I would also try decreasing the line height.

Francisco2002:
> The ticket also asks to check if it is possible to decrease the font size of the log to fit more lines which I think we can do. I would also try decreasing the line height.

Currently

| Font Size | Line Height |
| ------------ | ---------------- |
| 16 px | 24px |

dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
```diff
@@ -122,7 +122,7 @@ export const LogSheet = ({
     { enabled: !!logUrl },
   );
   return (
-    <SheetContent className="flex w-[25rem] flex-col sm:w-full sm:max-w-[44rem]">
+    <SheetContent className="flex flex-col text-sm sm:w-full sm:max-w-[clamp(650px,75vw,1300px)]">
```
lfjnascimento, line 125: `text-sm` uses a `font-size: 0.875rem` so we ca use a `line-height: 1rem`

lfjnascimento (approved):
Looks good! pre-approving

dashboard/src/components/Filter/CodeBlock.tsx
```diff
@@ -220,14 +220,14 @@ const CodeBlock = ({
 
       <CodeBlockDialog>
         <MemoizedCode
-          className="max-h-[520px]"
+          className="max-h-[520px] text-sm leading-4"
```
Francisco2002, line 223: put the `text-sm leading-4` classes inside Code Component

Francisco2002 (approved):
Nice!! Pre approving

PR #636, opened 2024-12-02
Feat: add filters in hardware details

Francisco2002:
Add, in hardware details, filters by:
- compilers
- archs
- configs

**How to test**

Select a hardware and go to hardware details. Open the filters modal and try apply filters.

### DIFF
--- a/backend/kernelCI_app/views/hardwareDetailsView.py
+++ b/backend/kernelCI_app/views/hardwareDetailsView.py
@@ -20,6 +20,7 @@ from django.views.decorators.csrf import csrf_exempt
 DEFAULT_DAYS_INTERVAL = 3
 SELECTED_HEAD_TREE_VALUE = 'head'
 STATUS_FAILED_VALUE = "FAIL"
+NULL_STRINGS = ["null", "Unknown"]
 
 
 def properties2List(d, keys):
@@ -159,7 +160,7 @@ class HardwareDetails(View):
         processed_builds: Set[str],
     ) -> bool:
         is_build_not_processed = not build["id"] in processed_builds
-        is_build_filtered_out = len(filters_map["build"]) == 0 or self.pass_in_filters(
+        is_build_filtered_out = len(filters_map["build"]) == 0 or self.pass_test_filters(
             build, filters_map["build"]
         )
         return is_build_not_processed and is_build_filtered_out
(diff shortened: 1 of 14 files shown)


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -208,10 +208,24 @@ def sanitize_records(self, records, selected_trees, filters):
...
 
-            jump_test = not test_filter_is_empty and not test_pass_in_filters
+            build_pass_in_filters = True
+
+            if build_id not in processed_builds:
+                processed_builds.add(build_id)
+                build = get_build(r, current_tree["index"])
+
+                if len(filters_map['build']) == 0 or self.pass_in_filters(build, filters_map['build']):
+                    filtered_builds.add(build_id)
```
murilx, line 221: You are adding builds to this `filtered_builds` set but not using it anywhere else

MarceloRobert (commented):
The config filter doesn't seem to be working for boots and tests. Also, please return the "click to filter" in the arch summary/config cards in the frontend

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -224,7 +250,12 @@ def sanitize_records(self, records, selected_trees, filters):
 
         for r in records:
             current_tree = get_current_selected_tree(r, selected_trees)
-            if not current_tree or not is_record_selected(r, current_tree):
+
+            is_tree_selected = current_tree and is_record_selected(r, current_tree)
+            has_global_filter = len(filters_map['global']) == 0
+            pass_in_global_filters = has_global_filter or self.pass_build_filter(r, filters_map['global'])
```
murilx, line 256: Since `self.pass_build_filter` already returns True if `len(filters_map['global']) == 0` there's no need for `has_global_filter`

MarceloRobert, line 256: these names are a bit confusing, `has_global_filter` is True if the length of the filters is 0? Also, should `pass_in_global_filters` be using `or` in its comparison?

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -168,36 +168,46 @@ def get_filters(self, filters):
...
-                and value[0] not in data[field]
-            ):
+            build_field = f"build__{field}"
+
+            if (data[build_field] not in value):
                 return False
 
+        return True
+
+    def pass_in_filters(self, data, filters):
```
MarceloRobert, line 203: nit: rename this function to be something like "pass_test_filter" (either in singular or plural, but make it consistent with "pass_build_filter"

murilx (approved):

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -168,36 +168,49 @@ def get_filters(self, filters):
...
             filters_map[table].append({"field": field, "value": value})
 
         return filters_map
 
-    def pass_in_filters(self, data, filters):
+    def pass_build_filter(self, data, filters):
         for currentFilter in filters:
             field = currentFilter.get("field")
             value = currentFilter.get("value")
+            include_null = "null" in value or "Unknown" in value
```
MarceloRobert, line 195: you could extract these strings to a const at the top and do something like `include_null = value in null_strings` with `NULL_STRINGS = ["null", "Unknown"]`

MarceloRobert (approved):
LGTM

PR #647, opened 2024-12-05
Feat: add filters in hardware details

Francisco2002:
This PR refactor de logic of filters and add build, boots and tests duration filters.

Like duration in hardware details are nullable, I am sending a patch test to allow the test of duration filters

### DIFF
--- a/backend/kernelCI_app/utils.py
+++ b/backend/kernelCI_app/utils.py
@@ -111,9 +111,45 @@ class FilterParams:
 
     string_like_filters = ["boot.path", "test.path"]
 
-    def __init__(self, request):
+    def __init__(self, data, process_body=False):
         self.filters = []
-        self.create_filters_from_req(request)
+        if process_body:
+            self.create_filters_from_body(data)
+        else:
+            self.create_filters_from_req(data)
+
+    def create_filters_from_body(self, body):
+        filters = body.get("filter", {})
+        for k in filters.keys():
+            if not k.startswith(self.filter_param_prefix):
+                continue
+            # HACK: Flake8 will always bug with (): so we define a variable here
+            filter_param_prefix_length = len(self.filter_param_prefix)
+            filter_term = k[filter_param_prefix_length:]
+
+            # filter as list
+            filter_data = filters.get(k)
+
+            if type(filter_data) is list and len(filter_data) > 0:
+                field = filter_term
+                values = filter_data
+
+                for value in values:
+                    self.add_filter(field, value, "in")
+                continue
+
+            if filter_term in self.string_like_filters:
+                self.add_filter(filter_term, filter_data, "like")
+                continue
+
+            match = self.filter_reg.match(filter_term)
+            if match:
+                field = match.group(1)
+                comparison_op = match.group(2)
+                self.add_filter(field, filter_data, comparison_op)
+                continue
+
+            self.add_filter(filter_term, filter_data, "exact")
 
     def create_filters_from_req(self, request):
         for k in request.GET.keys():
--- a/backend/kernelCI_app/utils.py
+++ b/backend/kernelCI_app/utils.py
@@ -2,9 +2,11 @@ import json
 from typing import Union, TypedDict, List, Optional, Dict
 from django.utils import timezone
 from datetime import timedelta
+from django.http import HttpResponseBadRequest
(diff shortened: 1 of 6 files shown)


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -153,79 +156,221 @@ class HardwareDetails(View):
...
+
+    def __handle_boot_duration(self, current_filter):
+        value = current_filter["value"][0]
+        operation = current_filter["comparison_op"]
+        if operation == "lte":
+            self.filterBootDurationMax = toIntOrDefault(value, None)
+        else:
+            self.filterBootDurationMin = toIntOrDefault(value, None)
+
+    def __handle_test_status(self, current_filter):
```
WilsonNet, line 196: don't use `__` that was a missconception that we had, private methods have only one underscore before them

also, these handlers are very close to the one in hardware details view, se if you can unite them

the trick is to receive what they are modifying as a parameter
or to return the value
I'd rather have them returning the value

backend/kernelCI_app/utils.py
```diff
@@ -2,9 +2,11 @@
 from typing import Union, TypedDict, List, Optional, Dict
 from django.utils import timezone
 from datetime import timedelta
+from django.http import HttpResponseBadRequest
 import re
 
 DEFAULT_QUERY_TIME_INTERVAL = {"days": 7}
+NULL_STRINGS = ["null", "Unknown"]
```
murilx, line 9: You should use a set in this case

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -23,7 +25,7 @@
 NULL_STRINGS = ["null", "Unknown"]
```
murilx, line 25: You can import this from utils

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -153,79 +156,221 @@ class HardwareDetails(View):
...
+            not record['build__architecture']
+            and any(null_string in self.filterArchitecture for null_string in NULL_STRINGS)
+        )
+        record_compiler_null_accept = (
+            not record['build__compiler']
+            and any(null_string in self.filterTreeDetailsCompiler for null_string in NULL_STRINGS)
+        )
+        record_config_null_accept = (
+            not record['build__config_name']
+            and any(null_string in self.filterTreeDetailsConfigs for null_string in NULL_STRINGS)
```
murilx, line 254: Is there a reason here to not just use `self.filterArchitecture in NULL_STRINGS` (the same applies to the other instances of `any(null_string in _ for null_string in NULL_STRINGS)`?

Francisco2002: `self.filterArchitecture` is a set of architecture to filter.
`self.filterTreeDetailsCompiler` and `filterTreeDetailsConfigs` too

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -442,7 +378,8 @@ def get(self, request, commit_hash: str | None):
         origin_param = request.GET.get("origin")
         git_url_param = request.GET.get("git_url")
         git_branch_param = request.GET.get("git_branch")
-        self.__processFilters(request)
+        self.filterParams = FilterParams(request)
+        self.setup_filters()
```
WilsonNet, line 382: much better now

backend/kernelCI_app/utils.py
```diff
@@ -111,9 +113,137 @@ class FilterParams:
...
+        try:
+            for current_filter in self.filters:
+                field = current_filter["field"]
+                # Delegate to the appropriate handler based on the field
+                if field in self.filter_handlers:
+                    self.filter_handlers[field](current_filter)
+        except InvalidComparisonOP as e:
+            return HttpResponseBadRequest(getErrorResponseBody(str(e)))
+
+    def create_filters_from_body(self, body: Dict):
```
WilsonNet, line 214: one day we need to make this a little better but the improvements here are good for today

WilsonNet (approved):
niceo one

murilx (approved):
Nicely done!

PR #650, opened 2024-12-06
feat(tree-listing): reduce difference between listing and hardware

WilsonNet:
Reduce the difference between listing and hardware details

Part of #577

### DIFF
--- a/backend/kernelCI_app/views/hardwareDetailsView.py
+++ b/backend/kernelCI_app/views/hardwareDetailsView.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from django.db.models import Subquery
 import json
 from typing import Dict, List, Optional, Set, Literal
 from kernelCI_app.helpers.filters import (
@@ -26,19 +27,22 @@ from kernelCI_app.utils import (
 )
 from kernelCI_app.constants.general import DEFAULT_ORIGIN
 from django.views.decorators.csrf import csrf_exempt
+from kernelCI_app.helpers.trees import get_tree_heads
 
 DEFAULT_DAYS_INTERVAL = 3
 SELECTED_HEAD_TREE_VALUE = 'head'
 STATUS_FAILED_VALUE = "FAIL"
 
+BuildStatusType = Literal["valid", "invalid", "null"]
 
-def properties2List(d: Dict, keys: str):
-    for k in keys:
-        v = d[k]
-        if type(v) is dict:
-            d[k] = list(v.values())
-        elif type(v) is set:
-            d[k] = list(v)
+
+def mutate_properties_to_list(dict: Dict, keys: List[str]) -> None:
+    for key in keys:
+        value = dict[key]
+        if isinstance(value, Dict):
+            dict[key] = list(value.values())
+        elif isinstance(value, Set):
+            dict[key] = list(value)
 
 
 def get_arch_summary(record: Dict):
@@ -171,7 +175,7 @@ def generate_tree_status_summary_dict():
     }
 
 
-def get_build_status(is_build_valid):
+def get_build_status(is_build_valid) -> BuildStatusType:
     if is_build_valid is True:
         return "valid"
     elif is_build_valid is False:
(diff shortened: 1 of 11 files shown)


dashboard/src/api/hardwareDetails.ts
```diff
@@ -86,7 +88,8 @@ export const useHardwareDetails = (
 
   const body: fetchHardwareDetailsBody = {
     origin,
-    limitTimestampInSeconds,
+    startTimestampInSeconds,
```
WilsonNet, line 91: make it be in hours so the cache doesn't suffer that much

dashboard/src/utils/date.ts
```diff
@@ -8,3 +8,43 @@ export const dateObjectToTimestampInSeconds = (date: Date): number => {
...
+  const dateResult = date.toLocaleDateString('en-US', {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+  });
+
+  return dateResult;
+};
+
+type GetFormattedTime = string | undefined;
```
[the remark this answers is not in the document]
WilsonNet: this is a leftover, thx

MarceloRobert (approved):
Worked well on my tests, seems much better

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -26,13 +26,13 @@
...
 
 
-def properties2List(d: Dict, keys: str):
-    for k in keys:
-        v = d[k]
-        if type(v) is dict:
-            d[k] = list(v.values())
-        elif type(v) is set:
-            d[k] = list(v)
+def mutate_properties_to_list(dict: Dict, keys: List[str]):
```
murilx, line 29:
```suggestion
def mutate_properties_to_list(dict: Dict, keys: List[str]) -> None:
```

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -426,13 +426,13 @@ def sanitize_records(self, records, trees: List, is_all_selected: bool):
...
-        properties2List(tests, ["issues", "platformsFailing", "archSummary"])
-        properties2List(boots, ["issues", "platformsFailing", "archSummary"])
+        mutate_properties_to_list(builds, ["issues"])
+        mutate_properties_to_list(tests, ["issues", "platformsFailing", "archSummary"])
+        mutate_properties_to_list(boots, ["issues", "platformsFailing", "archSummary"])
 
         return (builds, tests, boots, tree_status_summary)
 
-    def get_trees(self, hardware_id: str, origin: str, limit_datetime: int):
+    def get_trees(self, hardware_id: str, origin: str, start_datetime: int, end_datetime: int):
```
murilx, line 435:
```suggestion
    def get_trees(self, hardware_id: str, origin: str, start_datetime: int, end_datetime: int) -> List[Dict]:
```

dashboard/src/utils/date.ts
```diff
@@ -8,3 +8,41 @@ export const dateObjectToTimestampInSeconds = (date: Date): number => {
...
+  }
+
+  const timeResult = date.toLocaleTimeString('en-US', {
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+  });
+
+  return timeResult;
+};
```
Francisco2002, line 48: we can unify this in a function.
```suggestion
export const getFormattedDatetime = (
  timestampInSeconds: number,
): string | undefined => {
  const date = new Date(timestampInSeconds * MILLISECONDS_IN_ONE_SECOND);
  if (!isValidDate(date)) {
    return undefined;
  }

  const dateResult = date.toLocaleDateString('en-US', {
    year: 'numeric',
    month: 'long',
    day: 'numeric',
    hour: '2-digit',
    minute: '2-digit',
    second: '2-digit',
  });

  return dateResult;
}
};
```

WilsonNet: I will leave separated because it is more flexible

murilx (approved):

PR #659, opened 2024-12-09
feat: update tanstack/react-router to v1.87.1

### LINKED ISSUE #658: Update tanstack/react-query package
Update tanstac/react-query package from version 1.46.1 to version 1.87.1. This update was motivated so that we can have access to some features included after the version 1.46.1, mainly search middlewares (first included in version 1.67.0 as can be seen here https://github.com/TanStack/router/commit/60b3f37e32f9ef58879d3169bc35f7c1a4d0b2f2) that will be very useful for shortening the URL.

murilx:
Closes #658

## How to test
Navigate the pages of the dashboard and add, remove filters and any other actions that change the URL to verify that the behaviour and URL is the same as it was before.

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -39,7 +39,7 @@
     "@tanstack/query-sync-storage-persister": "^5.56.2",
     "@tanstack/react-query": "^5.56.2",
     "@tanstack/react-query-persist-client": "^5.56.2",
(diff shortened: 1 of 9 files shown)


WilsonNet (approved):
Worked on my tests

MarceloRobert (approved):
Worked well on my tests

PR #666, opened 2024-12-10
feat: Remove default values from URL

### LINKED ISSUE #656: remove default values from URL

murilx:
- Added the search middleware `stripSearchParams` to remove search params with default values from the URL
- Changed `from` options to use route instead of index e.g from `/tree/$treeId/` to `/tree/$treeId`
- Added some default clauses to zod types to make the default values of search params be returned when they are stripped from the URL. Without this clause the returned value would be undefined

Closes #656

## How to test
Navigate through the app, verifying that default values aren't being shown in the URL while keeping the same behaviour. Also note that when changing something, the search params goes back to the URL.

### DIFF
--- a/dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
+++ b/dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
@@ -10,8 +10,6 @@ import { Skeleton } from '@/components/Skeleton';
 import { useTreeDetails } from '@/api/treeDetails';
 import BaseCard from '@/components/Cards/BaseCard';
 
-import type { TestsTableFilter } from '@/types/tree/TreeDetails';
-
 import MemoizedIssuesList from '@/components/Cards/IssuesList';
 import MemoizedHardwareTested from '@/components/Cards/HardwareTested';
 
@@ -26,7 +24,8 @@ import MemoizedConfigList from '@/components/Tabs/Tests/ConfigsList';
 import MemoizedErrorsSummary from '@/components/Tabs/Tests/ErrorsSummary';
 
 import MemoizedStatusCard from '@/components/Tabs/Tests/StatusCard';
-import type { TFilter } from '@/types/general';
+
+import type { TestsTableFilter, TFilter } from '@/types/general';
 
 import TreeCommitNavigationGraph from '@/pages/TreeDetails/Tabs/TreeCommitNavigationGraph';
 
--- a/dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
+++ b/dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
@@ -35,14 +35,14 @@ interface TestsTabProps {
 }
 
 const TestsTab = ({ reqFilter }: TestsTabProps): JSX.Element => {
-  const { treeId } = useParams({ from: '/tree/$treeId/' });
+  const { treeId } = useParams({ from: '/tree/$treeId' });
   const { isLoading, data, error } = useTreeDetails({
     treeId: treeId ?? '',
     filter: reqFilter,
   });
 
   const { tableFilter, diffFilter } = useSearch({
-    from: '/tree/$treeId/',
+    from: '/tree/$treeId',
   });
   const currentPathFilter = diffFilter.testPath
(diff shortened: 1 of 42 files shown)


dashboard/src/types/tree/TreeDetails.tsx
```diff
@@ -146,7 +163,8 @@ export const zTreeInformation = z
     commitName: z.string().optional().catch(''),
```
WilsonNet, line 163: another ticket for you, create a ticket to remove parameters that does not need to be in the URL

the URL is to remember user choices

dashboard/src/routes/hardware/route.tsx
```diff
@@ -1,14 +1,20 @@
...
 
+const defaultValues = {
+  intervalInDays: DEFAULT_HARDWARE_INTERVAL_IN_DAYS,
+  hardwareSearch: '',
+};
+
 const zHardwareSchema = z.object({
   intervalInDays: makeZIntervalInDays(DEFAULT_HARDWARE_INTERVAL_IN_DAYS),
-  hardwareSearch: z.string().optional().catch(undefined),
+  hardwareSearch: z.string().optional().catch(''),
```
WilsonNet, line 14: from what I'm understanding with stripSerach poarameters you don't need `.optional` anymore

dashboard/src/routes/tree/route.tsx
```diff
@@ -1,14 +1,21 @@
...
 import { DEFAULT_TIME_SEARCH } from '@/pages/treeConstants';
 
+const defaultValues = {
+  intervalInDays: DEFAULT_TIME_SEARCH,
+  treeSearch: '',
+};
+
 export const RootSearchSchema = z.object({
   intervalInDays: makeZIntervalInDays(DEFAULT_TIME_SEARCH),
+  treeSearch: z.string().optional().catch(''),
```
WilsonNet, line 15: ditto about optional

dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
```diff
@@ -35,14 +35,14 @@ interface TestsTabProps {
...
-  const { treeId } = useParams({ from: '/tree/$treeId/' });
+  const { treeId } = useParams({ from: '/tree/$treeId' });
   const { isLoading, data, error } = useTreeDetails({
     treeId: treeId ?? '',
     filter: reqFilter,
   });
 
   const { tableFilter, diffFilter } = useSearch({
-    from: '/tree/$treeId/',
+    from: '/tree/$treeId',
```
WilsonNet, line 45: Why did you change this?
`/` is suppose to be the leaf, and that is the more exact place in this current route

dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
```diff
@@ -35,14 +35,14 @@ interface TestsTabProps {
 }
 
 const TestsTab = ({ reqFilter }: TestsTabProps): JSX.Element => {
-  const { treeId } = useParams({ from: '/tree/$treeId/' });
+  const { treeId } = useParams({ from: '/tree/$treeId' });
```
WilsonNet, line 38: ditto about leaf

dashboard/src/types/tree/TreeDetails.tsx
murilx: Done https://github.com/kernelci/dashboard/issues/671

MarceloRobert (commented):
From my tests, it seems that the buildDetails page still receives an empty diffFilter, and the hardwareDetails page is behaving weirdly, I can't deselect trees and it seems laggy

dashboard/src/routes/hardware/route.tsx
murilx: Done

WilsonNet (approved):
It is working much better than it was, maybe down the line we can make the urls more precise

MarceloRobert (commented):
Going from a build details to a test details the default values are still showing - [example](http://localhost:5173/build/maestro%3A675ae0a0ca49c97d2997f2c4?intervalInDays=1&startTimestampInSeconds=1733943600&endTimestampInSeconds=1734030000)

MarceloRobert (approved):
LGTM, worked well on my tests

PR #669, opened 2024-12-11
Feat: hardware listing tab direction

### LINKED ISSUE #523: Hardware listing table directs to last viewed tab
The behavior of the tree listing table is to direct the user to the build tab on every column except for "boot status" that sends directly to the boots tab and "test status" that sends to tests tab. Currently, the hardware listing table is just forwarding the search params, not directing to a specific tab as it is done in the tree listing table. If you leave hardware details page and go back to it you will land on the same tab as you were previously.

## How to test:
Go to the hardware listing, enter in any hardware detail page, go to boots tab, go back to hardware listing; now any cell will direct you to boots tab. Keep navigating to hardware details, change the tab, go back to the listing and navigate back to any hardware details again, the tab you'll land will be the same that you left. Try doing the same with tree monitor, the tree listing table will force a specific tab depending on the column that you click.

MarceloRobert:
- Adds the correct tab direction to the hardwareListingTable by passing a custom property in the column meta
- Refactors the treeListingTable to use the same logic as the hardwareListingTable for tab direction

The tab direction was made by using the `meta` property for the column definition, which doesn't have explicit typing so I need to typecast it to a custom type. If I extended the `Column` or `ColumnDef` types, they got lost in the abstraction of the cell (I can define `columns` as a `CustomColumn` type, but `cell.column` will still be the original `Column` type, so I can't retrieve the custom property)

Another way to do this, which is told by [the documentation](https://tanstack.com/table/v8/docs/api/core/column-def#meta), is to extend the entire module of tanstack-table to define a global meta type. This works but it will type the meta parameter to all tables, even those who don't need it, so I preferred to use the typecast in the tables I actually need it. The module extension would be something like this:
```typescript
declare module '@tanstack/react-table' {
  interface ColumnMeta<TData extends RowData, TValue> {
    tabTarget?: string;
  }
}
```

## How to test:
Go to the hardwareListing or treeListing page and enter a details page by clicking in different columns, the boot status and test status should direct you to the build tab and tests tab respectively, while the other columns should direct to the build tab

Closes #523

### DIFF
--- a/dashboard/src/components/TreeListingPage/TreeTable.tsx
+++ b/dashboard/src/components/TreeListingPage/TreeTable.tsx
@@ -51,6 +51,8 @@ import { BuildStatus, GroupedTestStatus } from '@/components/Status/Status';
 import { TableHeader } from '@/components/Table/TableHeader';
 import { PaginationInfo } from '@/components/Table/PaginationInfo';
 
+import type { ListingTableColumnMeta } from '@/types/table';
+
 import { InputTime } from './InputTime';
 
 const MemoizedInputTime = memo(InputTime);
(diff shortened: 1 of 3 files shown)


murilx (approved):
Working well on my tests

Francisco2002 (approved):
nice!!

PR #670, opened 2024-12-11
feat(LogSheet): Update UI of log download links

### LINKED ISSUE #655: Log Viewer: full log downloads does't looks like downloadable. Improve UI to make it clear.

murilx:
- Added a new function `truncateUrl` used to truncate the URL in the log viewer
- Changed the title of the download log table to "Full logs"
- Changed the messaged of the QuerySwitcher to "Download your logs here"
- Changed the style of the log links, making their font size greater and their color blue with an underline hover effect

Closes #655 

## Visual Reference
Before
log_before

After
log_after


dashboard/src/lib/string.ts
```diff
@@ -8,3 +8,25 @@ export const truncateBigText = (
   text && text.length > maxTextLength
     ? text.slice(0, maxTextLength) + '...'
     : valueOrEmpty(text);
+
+export const truncateUrl = (
+  url: string | undefined,
+  domainLength = 50,
+  endPathLength = 20,
+): string => {
+  if ('parse' in URL && url) {
```
murilx, line 17: If 'parse' not in URL (meaning the browser does not support this) but url is not undefined, the function should call truncateBigText so that the URL is not shown unprocessed. Other solution would be to refactor this function to always use native/supported JS functions like split('/')

dashboard/src/lib/string.ts
```diff
@@ -8,3 +8,26 @@ export const truncateBigText = (
...
+  url: string | undefined,
+  domainLength = 50,
+  endPathLength = 20,
+): string => {
+  if (url) {
+    if (url.length <= domainLength + endPathLength) {
+      return url;
+    }
+    const protocolRe = /^\w+:\/\//;
+    const searchParamsRe = /\?.*$/;
```
WilsonNet, line 22:
```suggestion
    const protocolRegex = /^\w+:\/\//;
    const searchParamsRegex = /\?.*$/;
```

dashboard/src/lib/string.ts
```diff
@@ -8,3 +8,26 @@ export const truncateBigText = (
...
+    let replacedUrl = url.replace(protocolRe, '');
+    replacedUrl = replacedUrl.replace(searchParamsRe, '');
+    const splittedUrl = replacedUrl.split('/');
+    const hostname = splittedUrl[0];
```
WilsonNet, line 26:
```suggestion
    const hostname = splittedUrl[0] ?? '';
```

typescript strict mode doesn't check array index

dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
```diff
@@ -202,7 +205,7 @@ export const LogSheet = ({
           skeletonClassname="h-[3rem]"
           customError={
             <div className="p-4 text-center">
-              This log url is not supported in the log viewer yet
+              You can download your logs here
```
WilsonNet, line 208: I now this wasn't on a message but please put this on a message

murilx: Completely forgot it, you are right

dashboard/src/lib/string.ts
murilx: I'm checking for undefined below

WilsonNet (approved):
Looks good, just some minor pointers

dashboard/src/lib/string.ts
```diff
@@ -8,3 +8,26 @@ export const truncateBigText = (
...
+  endPathLength = 20,
+): string => {
+  if (url) {
+    if (url.length <= domainLength + endPathLength) {
+      return url;
+    }
+    const protocolRegex = /^\w+:\/\//;
+    const searchParamsRegex = /\?.*$/;
+    let replacedUrl = url.replace(protocolRegex, '');
+    replacedUrl = replacedUrl.replace(searchParamsRegex, '');
```
Francisco2002, line 24:
```suggestion
    let replacedUrl = url.replace(protocolRegex, '').replace(searchParamsRegex, '');
```

Francisco2002 (approved):

dashboard/src/pages/TreeDetails/Tabs/LogSheet.tsx
```diff
@@ -202,7 +205,7 @@ export const LogSheet = ({
           skeletonClassname="h-[3rem]"
           customError={
             <div className="p-4 text-center">
-              This log url is not supported in the log viewer yet
+              <FormattedMessage id="logSheet.logQueryCustomError" />
```
MarceloRobert, line 208: This message is for the log downloader, I don't think we should change it to a download info here; maybe change it above in the place of "Index of"

MarceloRobert (approved):
Looks great

Issue #640, opened 2024-12-03
buildCommitGraph: some global filters aren't being applied


MarceloRobert:
The Hardware filter is not being applied to the build commit history graph.
Also make sure that the global filters are working in the commitHistoryGraph for the hardwareDetails page,

- [x] Refactor Query
- [x] Implement Filters
  - [x] Build Status
  - [x] Boot Status
  - [x] Test Status
  - [x] Configs
  - [x] Archs
  - [x] Compiler
  - [x] Hardware
  - [x] Build Duration
  - [x] Boots Duration
  - [x] Tests Duration

**Implemented in:**
- [x] TreeDetails
- [x] HardwareDetails

MarceloRobert:
For the hardwareDetails, the global filters are tracked by #600

MarceloRobert:
I didn't tag this as "Bug" since this isn't something breaking but rather something that wasn't implemented

PR #676, opened 2024-12-13
refactor: commit history chart

Francisco2002:
- remove builds, tests, boots counts CTE from query
- change the query to get the earliest commits
- compute count in python
- add filters in commit graph

Part of #640


backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,285 @@
...
+            or (
+                len(self.filterTreeDetailsConfigs) > 0
+                and (config_name not in self.filterTreeDetailsConfigs)
+            )
+        ):
+            return False
+
+        return True
+
+    def _process_tests(self, row: Dict, commit_hash: str) -> None:
```
MarceloRobert, line 243: You don't need to pass `commit_hash` here since it is a field from `row`

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,285 @@
...
+                self._process_nonboots_count(
+                    row['test_status'] or "NULL",
+                    commit_hash,
+                    row['test_duration'],
+                    row['test_path']
+                )
 
-            if field != "environment_compatible":
-                self.filters_options[table]['filters'].append(clause)
+    def _process_builds(self, row: Dict, commit_hash: str) -> None:
```
MarceloRobert, line 263: ditto about `commit_hash` parameter

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,285 @@
...
+        if row["build_id"] is not None and row['build_id'] not in self.processed_builds:
+            self.processed_builds.add(row['build_id'])
+            self._process_builds_count(row["build_valid"], row['build_duration'], commit_hash)
 
-            if field in ["config_name", "architecture", "compiler", "environment_compatible"]:
-                self.filters_options['test']['filters'].append(clause)
-                self.filters_options['boot']['filters'].append(clause)
+    def _process_rows(self, rows: Dict) -> None:
+        rowsSanitize = self.sanitize_rows(rows)
```
MarceloRobert, line 269: use snake_case for variable names, `sanitized_rows` would also be a better name, in my opinion

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,285 @@
...
+        self.filterBootStatus = self.filterParams.filterBootStatus
+        self.filterTestStatus = self.filterParams.filterTestStatus
+        self.filterTreeDetailsConfigs = self.filterParams.filterConfigs
+        self.filterTreeDetailsCompiler = self.filterParams.filterCompiler
+        self.filterArchitecture = self.filterParams.filterArchitecture
+        self.filterHardware = self.filterParams.filterHardware
+        self.filterTestPath = self.filterParams.filterTestPath
+        self.filterBootPath = self.filterParams.filterBootPath
+        self.filterBuildValid = self.filterParams.filterBuildValid
+        self.filterIssues = self.filterParams.filterIssues
```
MarceloRobert, line 42: `self.filterIssues` is not being used, the commit graph is still not reacting to issue filters

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,285 @@
...
+                self.filterBuildDurationMin is not None and (
+                    toIntOrDefault(duration, 0) < self.filterBuildDurationMin
+                )
+            )
+        ):
+            return
+
+        label = 'null'
+        if build_valid is not None:
+            label = 'valid' if build_valid else 'invalid'
```
MarceloRobert, line 120: you could use `.lower()` as used above and used in the label for test status

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -1,155 +1,241 @@
...
+        self.filterBuildDurationMax = self.filterParams.filterBuildDurationMax
+        self.filterBootStatus = self.filterParams.filterBootStatus
+        self.filterTestStatus = self.filterParams.filterTestStatus
+        self.filterTreeDetailsConfigs = self.filterParams.filterConfigs
+        self.filterTreeDetailsCompiler = self.filterParams.filterCompiler
+        self.filterArchitecture = self.filterParams.filterArchitecture
+        self.filterHardware = self.filterParams.filterHardware
+        self.filterTestPath = self.filterParams.filterTestPath
+        self.filterBootPath = self.filterParams.filterBootPath
+        self.filterBuildValid = self.filterParams.filterBuildValid
```
MarceloRobert, line 42: The issueFilter is still not being applied to the commitHistoryGraph, I think it can be done in this PR as well, can't it?

backend/kernelCI_app/utils.py
```diff
@@ -7,7 +7,7 @@
 
 DEFAULT_QUERY_TIME_INTERVAL = {"days": 7}
 UNKNOWN_STRING = "Unknown"
-NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
+NULL_STRINGS = set(["null", "Unknown", "NULL"])
```
MarceloRobert: I don't see the reason for this change

backend/kernelCI_app/utils.py
MarceloRobert, line 1: I think the filtering functions would be better placed in /helpers/filters.py

Francisco2002 (commented):
The test table in hardware details isn't consistent with the layout

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -79,60 +90,186 @@ def _init_commit_entry(self):
...
-            architecture = "Unknown"
-            compiler = "Unknown"
-            config_name = "Unknown"
-
-            if row['hardware_compatibles'] is not None:
-                hardware_compatible = row['hardware_compatibles'][0]
-            if row['architecture'] is not None:
-                architecture = row['architecture']
-            if row['compiler'] is not None:
-                compiler = row['architecture']
```
MarceloRobert: I didn't notice this was wrong before, well seen

backend/kernelCI_app/utils.py
```diff
@@ -318,3 +320,168 @@ def get_grouped_filters(self):
...
+            )
+        )
+
+    def is_record_filtered_out(
+            self, *,
+            hardwares: Optional[List[str]] = None,
+            architecture: Optional[str],
+            compiler: Optional[str],
+            config_name: Optional[str],
+            hasHardware=True
```
MarceloRobert, line 358: the `hasHardware` parameter isn't being used when called, I don't think these parameters should assume a default value

backend/kernelCI_app/utils.py
Francisco2002: done

WilsonNet (approved):
it worked on my tests except for hardware that is being limited by time, but there is a ticket for that

PR #684, opened 2024-12-16
feat: Remove default values from URL

### LINKED ISSUE #656: remove default values from URL

murilx:
- Added the search middleware `stripSearchParams` to remove search params with default values from the URL
- Changed `from` options to use route instead of index e.g from `/tree/$treeId/` to `/tree/$treeId`
- Added some default clauses to zod types to make the default values of search params be returned when they are stripped from the URL. Without this clause the returned value would be undefined

Closes #656 

Solved the bug that caused the revert where trying to use the table filter inside a build detail would cause a error message. Also solved a bug where trying to change the table filter from 'all' to any other after expanding a row in the tests table would cause the filter to not change.

### DIFF
--- a/dashboard/src/pages/BuildDetails/BuildDetails.tsx
+++ b/dashboard/src/pages/BuildDetails/BuildDetails.tsx
@@ -8,7 +8,10 @@ import {
   useRouterState,
 } from '@tanstack/react-router';
 
-import type { TestsTableFilter } from '@/types/tree/TreeDetails';
+import {
+  zTableFilterInfoDefault,
+  type TestsTableFilter,
+} from '@/types/tree/TreeDetails';
 import BuildDetails from '@/components/BuildDetails/BuildDetails';
 
 import { RedirectFrom } from '@/types/general';
@@ -40,8 +43,7 @@ const BuildDetailsPage = (): JSX.Element => {
           return {
             ...previousParams,
             tableFilter: {
-              bootsTable: previousParams.tableFilter.bootsTable,
-              buildsTable: previousParams.tableFilter.buildsTable,
+              ...(previousParams.tableFilter ?? zTableFilterInfoDefault),
               testsTable: filter,
             },
           };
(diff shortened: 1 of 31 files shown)


dashboard/src/pages/TreeBuildDetails/TreeBuildDetails.tsx
```diff
@@ -90,7 +97,7 @@ const TreeBuildDetails = (): JSX.Element => {
       buildId={buildId}
       breadcrumb={breadcrumbElement}
       onClickFilter={onClickFilter}
-      tableFilter={searchParams.tableFilter}
+      tableFilter={searchParams.tableFilter ?? zTableFilterInfoDefault}
```
MarceloRobert, line 100: is this needed? here you are using searchParams, not previousParams
also check this on the other files that have the same check, the other "...BuildDetails" files

dashboard/src/pages/BuildDetails/BuildDetails.tsx
```diff
@@ -40,8 +43,12 @@ const BuildDetailsPage = (): JSX.Element => {
...
             ...previousParams,
             tableFilter: {
-              bootsTable: previousParams.tableFilter.bootsTable,
-              buildsTable: previousParams.tableFilter.buildsTable,
+              bootsTable:
+                previousParams.tableFilter?.bootsTable ??
+                zTableFilterInfoDefault.bootsTable,
+              buildsTable:
+                previousParams.tableFilter?.buildsTable ??
+                zTableFilterInfoDefault.buildsTable,
```
MarceloRobert, line 51: On some other files (hardwareDetails boots and tests tabs and treeDetails boots tab), the change in the table filter is passed such as 
```
tableFilter: {
    ...previousParams.tableFilter,
    bootsTable: newFilter,
}
```
that way is a bit cleaner, you could change it to that format if they show the same behavior (which in my tests they do)

murilx: Done!

dashboard/src/pages/TreeBuildDetails/TreeBuildDetails.tsx
murilx: Even though this shouldn't be the case (since buildDetailsSearchSchema defines tableFilter and a default value to it) sometimes when changing the table filter this value can still be `undefined` and this was the bug we where facing where trying to change the table filter would cause the message error "something went wrong" to appear. For that reason, I added this verification

MarceloRobert (approved):
Made more tests now and it seems to be working, LGTM

WilsonNet (approved):
worked great on my tests

Issue #375, opened 2024-10-10
Add unit tests to the project


MarceloRobert:
As said by @WilsonNet on https://github.com/kernelci/dashboard/pull/367#discussion_r1795278918 , it would be interesting to add unit tests (with [Vitest](https://vitest.dev/)) to make sure functions like regexes work well and also for future use in the project.

PR #685, opened 2024-12-16
feat: Change query params serialization

### LINKED ISSUE #657: turn json objects into smaller serialized structure (keys into queries params)

murilx:
Uses query-string parse/stringify as the new serialization and deserialization methods ensuring a cleaner URL as stated in the URL proposal document.

Closes #657 

## Discussion
- Currently, I'm using the character '|' to flatten the elements, but it's a character that can be URI encoded, so maybe we should think of another one to avoid making the URL too 'messy'. The same goes for the comma that I'm using to separate array elements in the URL (treeIndexes up to now). 
- Check if the way the data is structured is good. Because of the object 'flattening', the URL has ended up being larger (since it repeats the name 'treeInfo' several times, for example). I decided to start with the 'query-string' library, and that's why this 'flattening' was necessary. Another solution would be to use jsurl2 or create our own stringify solution that follows a different logic and wouldn't need to repeat these names multiple times.

## How to test
Navigate through all the application and interact with the elements that change the query parameters, e.g. change tabs, add or remove filters (try both by clicking in the cards and in the modal), change the table filter. In the hardware details page, change the selected trees.  

### Backward compatibility
To test link backward compatibility just navigate to a page in `staging.dashboard.kernelci.org:9000` and change all of this to `localhost:5173`, keeping all the path and query parameters unchanged. 

### Unit tests
To execute unit tests:
- Install vitest by running `pnpm install`
- Run the tests with `pnpm test`

You can also download the [vitest vscode extension](https://marketplace.visualstudio.com/items?itemName=vitest.explorer) to run the tests from inside the IDE.

This addition of unit tests is part of #375

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -50,6 +50,7 @@
     "date-fns": "^4.0.0",
     "eslint-import-resolver-typescript": "^3.6.3",
     "lucide-react": "^0.396.0",
+    "query-string": "^9.1.1",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
     "react-error-boundary": "^4.0.13",
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -6,6 +6,8 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
+    "test": "vitest",
+    "coverage": "vitest run --coverage",
     "storybook": "storybook dev -p 6006",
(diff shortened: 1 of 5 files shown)


MarceloRobert (approved):
Worked well on my tests, LGTM

Francisco2002 (approved):
nice!!

PR #686, opened 2024-12-16
fix: test table in build details wrong redirect

### LINKED ISSUE #681: Interacting with tests tables in the Build Details page causes it to go from TreeBuildDetails/HardwareBuildDetails to BuildDetails
This possibly happens because the component is navigating to /build/$buildId through useNavigate without passing the state when navigating, causing the BuildDetails component to lose the information from where the user is coming from. A possible solution would be to pass the state to the navigate function used by onClickFilter in these three components. For example in HardwareBuildDetails

```ts
 const onClickFilter = useCallback(
    (filter: TestsTableFilter): void => {
      navigate({
        search: previousParams => {
          return {
            ...previousParams,
            tableFilter: {
              bootsTable: previousParams.tableFilter?.bootsTable,
              buildsTable: previousParams.tableFilter?.buildsTable,
              testsTable: filter,
            },
          };
        },
        state: { id: hardwareId, from: RedirectFrom.Hardware }
      });
    },
    [navigate],
  );
```

murilx:
When interacting with the tests table in the Build Details page it would incorrectly redirect because navigate was not receiving the state. By passing the state to navigate this cease to happen.

Closes #681 

## TODO
- When clicking the tests table row it's still losing the state


dashboard/src/components/Table/TableComponents.tsx
```diff
@@ -1,4 +1,4 @@
-import type { LinkProps } from '@tanstack/react-router';
+import { type HistoryState, type LinkProps } from '@tanstack/react-router';
```
murilx, line 1:
```suggestion
import type { HistoryState, LinkProps } from '@tanstack/react-router';
```

WilsonNet (approved):
bug seems to be fixed on my tests

MarceloRobert (approved):
Worked on my tests, seems straightforward

Francisco2002 (approved):

PR #690, opened 2024-12-17
feat(url): truncate url with new function

### LINKED ISSUE #674: Verify truncated URLs all over the app (they should have tooltips for full URL)

Francisco2002:
- use truncateUrl function
- define default url length 12

Close #674


dashboard/src/utils/constants/general.ts
```diff
@@ -7,3 +7,4 @@ export const DOCUMENTATION_URL = 'https://docs.kernelci.org/';
 // eslint-disable-next-line no-magic-numbers
 export const ItemsPerPageValues = [5, 10, 20, 30, 40, 50];
 export const DEFAULT_TIME_SEARCH = 7;
+export const defaultUrlLength = 12;
```
murilx, line 10: `truncateBigText` has a default length of 50 characters, considering that `truncateUrl` has a default `endPathLength`of 20 we could up this to at least 30. Also, is this const really useful? Since we can pass the default value to the function argument itself (as it's already being done but with 50 instead of 12 or 30).

Francisco2002: When a pass no params, the git url don't truncate. So, when the value of url length was 12, I decide to keep this value and use in other url

murilx: This might be good for the git url present in the Tree Details table. In other pages like Build and Test Details I believe they have enough space to justify increasing the number of characters

murilx (approved):

dashboard/src/components/TreeListingPage/TreeTable.tsx
```diff
@@ -71,11 +71,7 @@ const columns: ColumnDef<TreeTableBody>[] = [
               {sanitizeTableValue(row.getValue('tree_name') ?? '', false)}
             </div>
           </TooltipTrigger>
-          <TooltipContent>
-            <a href={row.original.url} target="_blank" rel="noreferrer">
-              {sanitizeTableValue(row.original.url, false)}
-            </a>
```
MarceloRobert, line 77: I don't see a reason to remove this link, it seems that it is possible to open it in a new tab even if clicking on the cell will direct you to the treeDetails page

PR #703, opened 2024-12-19
feat: consider env_misc.platform as hardware

### LINKED ISSUE #692: Use platform from misc field (if available) instead of just showing ‘Unknown’ hardware.
For x86 platforms, we still don't have a solution for storing the platform name, so we will still have to rely on the misc fields.
Let's take advantage of the misc field and use it instead to fill the hardware information in places where we have to pass the hardware name, like in the `tree/<str:commit_hash>/full` API endpoint.

murilx:
Add a new logic to the backend where `environment_misc.platform` will work as a fallback case for when the `environment_compatabile` field is null in the database. When `environment_misc.plaftorm` is null, we'll use `build.misc.platform`, but the logic of the Hardware filter will be kept the same, meaning that the builds returned will be first the build associated with the tests that have that hardware. This means that even if `build.misc.platform` is null but the build is associated with a test that have `environment_compatible` or `environment_misc.platform` not null, it will also be added when filtering for that hardware

This drastically lowers the number of Unknown hardware but they can still occur if both `environment_compatible` and `environment_misc.platform` and `build.misc.platform` are null.

Closes #692 

## How to test
- Add hardware filters from the Hardware Used card, Hardware Tested card and Filter modal and verify that they are correctly being added by comparing with staging. Verify that the commit graph is also working properly, showing the correct graph and changing the commit correctly
- Add a platform filter (these are hardwares that do not appear in staging, e.g. `acer-R721T-grunt`) from the Hardware Used card, Hardware Tested card and Filter modal and make the same verifications as above, if the status are showing correctly (comparing the values in the Hardware Tested card and the value shown in the pie chart) and if the commit graph is working.
- Test in other origins to verify that the behaviour is consistent between them

### DIFF
--- a/backend/kernelCI_app/views/hardwareDetailsView.py
+++ b/backend/kernelCI_app/views/hardwareDetailsView.py
@@ -2,10 +2,7 @@ from collections import defaultdict
 from django.db.models import Subquery
 import json
 from typing import Dict, List, Optional, Set, Literal
-from kernelCI_app.helpers.filters import (
-    should_increment_test_issue,
-    is_build_invalid
-)
+from kernelCI_app.helpers.filters import should_increment_test_issue, is_build_invalid
 from kernelCI_app.helpers.logger import log_message
 from django.utils.decorators import method_decorator
 from django.views import View
@@ -14,22 +11,18 @@ from django.http import HttpResponseBadRequest, JsonResponse
 from kernelCI_app.cache import getQueryCache, setQueryCache
 from kernelCI_app.viewCommon import create_details_build_summary
 from kernelCI_app.models import Tests
-from kernelCI_app.utils import (
-    create_issue,
-    extract_error_message,
-    extract_platform,
-    getErrorResponseBody
-)
+from kernelCI_app.utils import create_issue, extract_error_message, getErrorResponseBody
 from kernelCI_app.constants.general import DEFAULT_ORIGIN
 from django.views.decorators.csrf import csrf_exempt
 from kernelCI_app.helpers.trees import get_tree_heads
-from kernelCI_app.helpers.filters import (
-    UNKNOWN_STRING,
-    FilterParams
+from kernelCI_app.helpers.filters import UNKNOWN_STRING, FilterParams
+from kernelCI_app.helpers.misc import (
+    handle_environment_misc,
+    env_misc_value_or_default
 )
 
 DEFAULT_DAYS_INTERVAL = 3
-SELECTED_HEAD_TREE_VALUE = 'head'
+SELECTED_HEAD_TREE_VALUE = "head"
 STATUS_FAILED_VALUE = "FAIL"
 
 BuildStatusType = Literal["valid", "invalid", "null"]
@@ -48,7 +41,7 @@ def get_arch_summary(record: Dict):
     return {
         "arch": record["build__architecture"],
         "compiler": record["build__compiler"],
-        "status": defaultdict(int)
+        "status": defaultdict(int),
     }
 
 
@@ -68,14 +61,16 @@ def get_build(record: Dict, tree_idx: int):
         "git_repository_branch": record["build__checkout__git_repository_branch"],
         "tree_name": record["build__checkout__tree_name"],
         "issue_id": record["build__incidents__issue__id"],
-        "tree_index": tree_idx
+        "tree_index": tree_idx,
     }
 
 
 def get_tree_key(record: Dict):
-    return record["build__checkout__tree_name"] + \
-        record["build__checkout__git_repository_branch"] + \
-        record["build__checkout__git_repository_url"]
+    return (
+        record["build__checkout__tree_name"]
+        + record["build__checkout__git_repository_branch"]
+        + record["build__checkout__git_repository_url"]
+    )
 
 
 def get_tree(record: Dict):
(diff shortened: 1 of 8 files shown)


backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -281,7 +281,7 @@ def handle_test(self, record, tests):
 
         if status == "ERROR" or status == "FAIL" or status == "MISS":
             tests["platformsFailing"].add(
-                extract_platform(record["environment_misc"])
+                handle_environment_misc(record["environment_misc"]).get("platform")
```
WilsonNet, line 284: I had no idea we were already getting platform faililng in hardware

backend/kernelCI_app/helpers/tests.py
```diff
@@ -0,0 +1,27 @@
+import json
+from typing import Union, TypedDict, Optional
+from kernelCI_app.helpers.filters import UNKNOWN_STRING
```
WilsonNet, line 3: `filters` is probably not the best place for `UNKNOWN_STRING` maybe `helper.constants` 

but I know it is not being introduced by this PR

backend/kernelCI_app/helpers/tests.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class EnvironmentMisc(TypedDict):
+    platform: str
+
+
+def handle_environment_misc(misc: Union[str, dict, None]) -> Optional[EnvironmentMisc]:
```
WilsonNet, line 10: the return here is not `Optional`, it is always returning

backend/kernelCI_app/helpers/tests.py
murilx: I agree with you. I can change this as well if you think it would be better

backend/kernelCI_app/helpers/tests.py
```diff
@@ -0,0 +1,27 @@
...
+        "platform": UNKNOWN_STRING,
+    }
+
+    if misc is None:
+        return parsed_misc
+    elif isinstance(misc, str):
+        try:
+            misc = json.loads(misc)
+        except json.JSONDecodeError:
+            return parsed_misc
```
WilsonNet, line 21: maybe it is a good idea to add the `log_message` here, for now it just prints but in the future will be doing something more

murilx: And keep returning the default value?

dashboard/src/pages/TreeDetails/Tabs/Tests/TestsTab.tsx
```diff
@@ -193,7 +196,10 @@ const TestsTab = ({ reqFilter }: TestsTabProps): JSX.Element => {
           <div>
             <MemoizedHardwareTested
               title={<FormattedMessage id="testsTab.hardwareTested" />}
-              environmentCompatible={data.testEnvironmentCompatible}
+              environmentCompatible={{
+                ...data.testEnvironmentCompatible,
+                ...data.testEnvironmentMisc,
+              }}
```
WilsonNet, line 202: `useMemo` especially when you have to spreads

backend/kernelCI_app/helpers/tests.py
WilsonNet: lets not worry for this for today, we need this quick

WilsonNet (approved):
it worked on my tests

backend/kernelCI_app/helpers/tests.py
WilsonNet: omg that was lack of attention, make it the return value back to optional and make it return None

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -218,8 +220,13 @@ def _process_rows(self, rows: Dict) -> None:
         sanitized_rows = self.sanitize_rows(rows)
 
         for row in sanitized_rows:
+            hardware_filter = row["hardware_compatibles"]
+            if hardware_filter is None or hardware_filter == UNKNOWN_STRING:
```
MarceloRobert, line 224: the sanitized rows are simply the original rows but as an object, the `row["hardware_compatibles"]` hasn't been sanitized to receive `UNKNOWN_STRING` yet. In other words, the comparison `hardware_filter == UNKNOWN_STRING` will never be true unless the original row in the database is literally "Unknown", so I think this comparison is not needed

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -218,8 +220,13 @@ def _process_rows(self, rows: Dict) -> None:
         sanitized_rows = self.sanitize_rows(rows)
 
         for row in sanitized_rows:
+            hardware_filter = row["hardware_compatibles"]
```
MarceloRobert, line 223: isn't this variable referring to the current hardware of a row? I don't think hardware\__filter_ is a good name for it

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -333,19 +349,26 @@ def _sanitize_rows(self, rows):
...
+                row_data["test_environment_misc"]
+            )
+            hardware_filter = test_environment_compatible
             git_repository_url = row_data["checkout_git_repository_url"]
 
-            if (test_environment_compatible):
+            if test_environment_compatible != UNKNOWN_STRING:
                 self.hardwareUsed.add(test_environment_compatible)
+            else:
+                hardware_filter = test_environment_misc.get("platform")
```
MarceloRobert, line 361: ditto about hardware_filter var name

MarceloRobert (approved):
Worked well on my tests, just a note to split the commits into the feature and the formatter refactor for better reviewing

PR #706, opened 2024-12-20
Fix: Navigation in Hardware Details by Commit History Chart

### LINKED ISSUE #688: We can't navigate in hardware details with commit chart
Go to this [hardware](http://localhost:5173/hardware/google%2Cjuniper?origin=maestro&intervalInDays=3&hardwareSearch=google&currentPageTab=global.boots&startTimestampInSeconds=1734121800&endTimestampInSeconds=1734381000&tableFilter=%7B%22buildsTable%22%3A%22all%22%2C%22bootsTable%22%3A%22all%22%2C%22testsTable%22%3A%22all%22%7D&diffFilter=%7B%7D&treeIndexes=%5B0%5D) with one tree selected in header and try to select other commit hash. The listing will be empty, even though the chart show counts.

Francisco2002:
Add startTimestamp and endTimestamp in commit history charts.
Close #688 

**How to Test**
- Select any Hardware and fix one tree to see the commits history. Try to select another commit in chart.
  - e.g. [hardware: google,juniper-sku16  tree: arbd/for-kernelci](http://localhost:5173/hardware/google%2Cjuniper-sku16?endTimestampInSeconds=1734714000&startTimestampInSeconds=1734454800&treeIndexes[]=0) 
- Select any Tree and see the same behavior as in staging

### DIFF
--- a/backend/kernelCI_app/views/treeCommitsHistory.py
+++ b/backend/kernelCI_app/views/treeCommitsHistory.py
@@ -2,6 +2,7 @@ from kernelCI_app.utils import getErrorResponseBody
 from rest_framework.views import APIView
 from rest_framework.response import Response
 from rest_framework import status
+from datetime import datetime, timezone
 from django.http import JsonResponse, HttpResponseBadRequest
 from django.db import connection
 from typing import Dict, List
@@ -16,6 +17,7 @@ from kernelCI_app.helpers.misc import (
     build_misc_value_or_default,
     env_misc_value_or_default,
 )
+from kernelCI_app.helpers.logger import log_message
 
 
 # TODO Move this endpoint to a function so it doesn't
@@ -24,6 +26,8 @@ class TreeCommitsHistory(APIView):
     def __init__(self):
         self.commit_hashes = {}
         self.filterParams = None
+        self.start_datetime = None
+        self.end_datetime = None
         self.field_values = dict()
         self.processed_builds = set()
         self.processed_tests = set()
@@ -201,6 +205,11 @@ class TreeCommitsHistory(APIView):
                 row["build_valid"], row["build_duration"], commit_hash
             )
 
+    def _is_record_in_time_period(self, start_time: datetime) -> bool:
+        if self.start_datetime is not None and self.end_datetime is not None:
+            return start_time >= self.start_datetime and start_time <= self.end_datetime
+        return True
+
     def _process_rows(self, rows: Dict) -> None:
         sanitized_rows = self.sanitize_rows(rows)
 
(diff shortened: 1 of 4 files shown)


backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -243,6 +253,17 @@ def _process_rows(self, rows: Dict) -> None:
...
+    def _process_time_range(self, request):
+        try:
+            self.end_datetime = datetime.fromtimestamp(
+                int(request.GET.get("endTimestampInSeconds")), timezone.utc
+            )
+            self.start_datetime = datetime.fromtimestamp(
+                int(request.GET.get("startTimestampInSeconds")), timezone.utc
+            )
+        except Exception as ex:
+            print(ex)
```
MarceloRobert, line 265: use `log_message` instead of print

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -262,6 +283,7 @@ def get(self, request, commit_hash):
                 status=status.HTTP_400_BAD_REQUEST,
             )
 
+        self._process_time_range(request)
```
MarceloRobert, line 286: I think you could process the time range only if it isn't None, otherwise it will throw an Exception every time you access this endpoint without the start/end time (which is what happens in any treeDetails page at the moment)

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -201,6 +204,11 @@ def _process_builds(self, row: Dict) -> None:
                 row["build_valid"], row["build_duration"], commit_hash
             )
 
+    def _is_record_in_time_period(self, start_time: datetime):
```
MarceloRobert, line 207:
```suggestion
    def _is_record_in_time_period(self, start_time: datetime) -> bool:
```

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -243,10 +254,23 @@ def _process_rows(self, rows: Dict) -> None:
             self._process_tests(row)
             self._process_builds(row)
 
+    def _process_time_range(self, request):
```
murilx, line 257:
```suggestion
    def _process_time_range(self, request) -> None:
```

MarceloRobert (approved):
LGTM, we should add an indicator that the data is loading when changing commits later, right now it seems to be loading slowly

murilx (approved):
Worked well on my tests

PR #711, opened 2024-12-23
Feat: IssueDetails test and build info

### LINKED ISSUE #705: IssueDetails: show tests and builds from an issue
Show a table with the builds and/or tests where it appears (one table for builds, another for tests)

MarceloRobert:
Adds /issues/<issue_id>/version/<version>/tests endpoint to get all tests related to an issue
Adds /issues/<issue_id>/version/<version>/builds endpoint to get all builds related to an issue
Adds request for these endpoints from the frontend
Adds a SectionError component for when a table in a details page can't load and refactors its use in BuildDetailsTestSection
Adds a TestsTable to the IssueDetails page when the current issue is related to a test
Adds a BuildsTable to the IssueDetails page when the current issue is related to a test

The relation with tests and builds are intuitively mutually exclusive, but - since the database allows for both `test_status` and `build_valid` to be not null in a single issue - it is technically possible for an issue to be associated with tests and builds at the same time. There are no issues related to both tests and builds currently.

It's also possible for an issue to not be related with either tests or builds, but there are no occurrences of that for the maestro origin, only for redhat.

## How to test
Go to an IssueDetails page with an issue related to tests and see the test table, such as this example:
http://localhost:5173/issue/maestro%3Af627e08f767a688209f3f9a2d41c6b4c260897b0/version/0

Go to an IssueDetails page with an issue related to builds and see the builds table, such as this example:
http://localhost:5173/issue/maestro:a8ba4fd7ab9f996d9e569f77d5cf91354e7a783b/version/0

An issue with no relation to either tests or builds for example: http://localhost:5173/issue/redhat:issue_2089/version/1727692707

Closes #705

### DIFF
--- a/backend/kernelCI_app/views/issueDetailsBuildsView.py
+++ b/backend/kernelCI_app/views/issueDetailsBuildsView.py
@@ -1,0 +1,61 @@
+from http import HTTPStatus
+from typing import Dict, Optional
+from django.http import JsonResponse
+from django.views import View
+from kernelCI_app.helpers.errorHandling import create_error_response
+from kernelCI_app.models import Incidents
+from kernelCI_app.typeModels.issueDetails import IssueDetailsPathParameters
+from pydantic import ValidationError
+
+
+class IssueDetailsBuilds(View):
+    def _fetch_incidents(self, *, issue_id: str, version: int) -> Optional[Dict]:
+        fields = [
+            "build__id",
+            "build__architecture",
+            "build__config_name",
+            "build__valid",
+            "build__start_time",
+            "build__duration",
+            "build__compiler",
+            "build__log_url",
+        ]
+
+        builds = Incidents.objects.filter(
+            issue_id=issue_id, issue_version=version
+        ).values(*fields)
+
+        return [
+            {
+                "id": build["build__id"],
+                "architecture": build["build__architecture"],
+                "config_name": build["build__config_name"],
+                "valid": build["build__valid"],
+                "start_time": build["build__start_time"],
+                "duration": build["build__duration"],
+                "compiler": build["build__compiler"],
+                "log_url": build["build__log_url"],
+            }
+            for build in builds
+        ]
+
+    def get(
+        self, _request, issue_id: Optional[str], version: Optional[str]
+    ) -> JsonResponse:
+        try:
+            parsed_params = IssueDetailsPathParameters(
+                issue_id=issue_id, version=version
+            )
+        except ValidationError as e:
+            return create_error_response(e.json())
+
+        builds_data = self._fetch_incidents(
+            issue_id=parsed_params.issue_id, version=parsed_params.version
+        )
+
+        if builds_data is None:
+            return create_error_response(
+                error_message="Builds not found", status_code=HTTPStatus.NOT_FOUND
+            )
+
+        return JsonResponse(builds_data, safe=False)
--- a/backend/kernelCI_app/views/issueDetailsTestsView.py
+++ b/backend/kernelCI_app/views/issueDetailsTestsView.py
@@ -1,0 +1,57 @@
+from http import HTTPStatus
+from typing import Dict, Optional
+from django.http import JsonResponse
+from django.views import View
+from kernelCI_app.helpers.errorHandling import create_error_response
+from kernelCI_app.models import Incidents
+from kernelCI_app.typeModels.issueDetails import IssueDetailsPathParameters
+from pydantic import ValidationError
+
+
+class IssueDetailsTests(View):
+    def _fetch_incidents(self, *, issue_id: str, version: int) -> Optional[Dict]:
+        fields = [
+            "test__id",
+            "test__duration",
+            "test__status",
+            "test__path",
+            "test__start_time",
+            "test__environment_compatible",
+        ]
+
+        tests = Incidents.objects.filter(
+            issue_id=issue_id, issue_version=version
+        ).values(*fields)
+
+        return [
+            {
+                "id": test["test__id"],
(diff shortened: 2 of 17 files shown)


dashboard/src/components/BuildDetails/BuildDetailsTestSection.tsx
```diff
@@ -12,6 +10,8 @@ import type { TableFilter, TestsTableFilter } from '@/types/tree/TreeDetails';
 
 import { TestsTable } from '@/components/TestsTable/TestsTable';
 
+import { MemoizedSectionError } from '../DetailsPages/SectionError';
```
murilx, line 13: Relative import

dashboard/src/components/IssueDetails/IssueDetails.tsx
```diff
@@ -14,20 +16,32 @@ import { getMiscSection } from '@/components/Section/MiscSection';
...
+  tableFilter,
+  onClickTestFilter,
+  getTestTableRowLink,
 }: IIssueDetails): JSX.Element => {
   const { data, error, isLoading } = useIssueDetails(
     issueId ?? '',
     versionNumber ?? '',
   );
 
+  const hasTest = data?.test_status !== null;
```
murilx, line 43: Wouldn't `data?.test_status` return undefined when it's empty?

MarceloRobert: you're right, I'll change the comparison to consider the test_status only when there is any data

backend/kernelCI_app/views/issueDetailsBuildsView.py
```diff
@@ -0,0 +1,60 @@
...
+        missing_params = []
+        if issue_id is None:
+            missing_params.append("issue_id")
+        if version is None:
+            missing_params.append("version")
+        if len(missing_params) != 0:
+            return HttpResponseBadRequest(
+                getErrorResponseBody("Missing parameters: ", missing_params)
+            )
+
```
WilsonNet, line 47: II'm pretty sure you can validate the parameters with pydantic

backend/kernelCI_app/views/issueDetailsBuildsView.py
```diff
@@ -0,0 +1,60 @@
...
+        try:
+            parsed_version = int(version)
+        except ValueError:
+            return HttpResponseBadRequest(
+                getErrorResponseBody("Invalid version parameter, must be an integer")
```
WilsonNet, line 52: we have `create_error_response` now

backend/kernelCI_app/views/issueDetailsBuildsView.py
```diff
@@ -0,0 +1,60 @@
...
+            )
+
+        builds_data = self._fetch_incidents(issue_id=issue_id, version=parsed_version)
+
+        if builds_data is None:
+            return HttpResponseNotFound(getErrorResponseBody("Builds not found"))
```
WilsonNet, line 58: ditto about create error response

backend/kernelCI_app/views/issueDetailsTestsView.py
```diff
@@ -0,0 +1,58 @@
...
+        ]
+
+    def get(self, _request, issue_id: Optional[str], version: Optional[str]):
+        missing_params = []
+        if issue_id is None:
+            missing_params.append("issue_id")
+        if version is None:
+            missing_params.append("version")
+        if len(missing_params) != 0:
+            return HttpResponseBadRequest(
```
WilsonNet, line 42: ditto for pydantic and create_error_response

dashboard/src/utils/utils.ts
```diff
@@ -94,6 +101,21 @@ export const sanitizeBuilds = (
   }));
 };
 
+export const sanitizeBuildTable = (
```
WilsonNet, line 104: This is creating an O(2N), and sometimes N equals 600000, maybe it is better to create an object with the sanitizers and place them on the row definition, but the only way of knowing is if we measure.

WilsonNet:
It is working nice, just some comments

dashboard/src/utils/utils.ts
MarceloRobert: The 600000 figure is for tests related to an issue, for builds the number is much lower. But I agree we should fix the typing of builds so that we won't need all this conversion.

MarceloRobert (commented):
Some tables aren't receiving the correct data. Builds table should also get the `log_url` from the backend and the tests table aren't showing the relevant data in the cells

Francisco2002 (approved):
nice job!!

dashboard/src/pages/IssueDetails/IssueDetails.tsx
```diff
@@ -1,13 +1,89 @@
...
+  const getBuildTableRowLink = useCallback(
+    (buildId: string): LinkProps => ({
+      to: '/build/$buildId',
+      params: {
+        buildId: buildId,
+      },
+      search: s => s,
+    }),
+    [],
+  );
```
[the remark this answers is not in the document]
MarceloRobert: This is how it is being done in the other buildDetails page files, if I were to change this it would be better to refactor over there as well, but that would pollute this PR even more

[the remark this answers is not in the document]
murilx: Maybe this would be good for the callbacks that do not have dependencies, like this one and the one in `pages/BuildDetails/BuildDetails.tsx` but I'm not so sure about the ones that have dependencies (even though it could also be possible). But I agree that this could be done in another PR

murilx (approved):
LGTM

PR #718, opened 2024-12-24
docs: add filter rules

WilsonNet:
Just some rules explaining how the filters should work


Francisco2002 (approved):

murilx (approved):

MarceloRobert (approved):
It's a bit tricky but makes sense

PR #724, opened 2025-01-06
Add latest tree data in route /<tree_name>/<branch>

### LINKED ISSUE #675: ‘dashboard.kernelci.org/<tree-name>/<branch>’ as shortcut to latest results
If the user goes to ‘dashboard.kernelci.org/<tree-name>/<branch>’ they should be redirected to

the latest results for that branch. It could include a URL redirect. This is useful for users to keep a link around to check from time to time or to script things through kci-dev.

For example: https://dashboard.kernelci.org/mainline/master should just open the latest mainline results.

murilx:
Add a new route `/<tree-name>/<branch>` that redirects to the TreeDetails for the latest checkout of the tree.

Closes #675 

## How to test
- To go `localhost:5173/<tree-name>/<branch>` for a valid `<tree-name>` and `<branch>` and see if it correctly redirects to the TreeDetails page. For example: http://localhost:5173/tree/android/android-mainline
- To go `localhost:5173/<tree-name>/<branch>` for an invalid `<tree-name>` and/or `<branch>` and see if it correctly shows an error message. For example: http://localhost:5173/tree/android/invalid-branch

### DIFF
--- a/backend/kernelCI_app/views/treeLatest.py
+++ b/backend/kernelCI_app/views/treeLatest.py
@@ -1,0 +1,83 @@
+from http import HTTPStatus
+from typing import Dict, Optional
+from urllib.parse import urlencode
+
+from django.http import JsonResponse
+from django.urls import reverse
+from django.views import View
+from pydantic import ValidationError
+
+from kernelCI_app.constants.general import DEFAULT_ORIGIN
+from kernelCI_app.helpers.errorHandling import create_error_response
+from kernelCI_app.models import Checkouts
+from kernelCI_app.typeModels.treeDetails import TreeLatestPathParameters
+
+
+class TreeLatest(View):
+    def _fetch_latest_tree(
+        self, tree_name: str, branch: str, origin: str
+    ) -> Optional[Dict]:
+        tree_fields = [
+            "git_commit_hash",
+            "git_commit_name",
+            "git_repository_url",
+        ]
+
+        query = (
+            Checkouts.objects.values(*tree_fields)
+            .filter(
+                origin=origin,
+                tree_name=tree_name,
+                git_repository_branch=branch,
+                git_commit_hash__isnull=False,
+            )
+            .order_by("-field_timestamp")
+            .first()
+        )
+
+        return query
+
+    def get(self, request, tree_name: str, branch: str) -> JsonResponse:
+        try:
+            parsed_params = TreeLatestPathParameters(tree_name=tree_name, branch=branch)
+        except ValidationError as e:
+            return create_error_response(e.json())
+
+        tree_not_found_error_message = "Tree not found."
+        origin = request.GET.get("origin")
+        if origin is None:
+            origin = DEFAULT_ORIGIN
+            tree_not_found_error_message += (
+                f" No origin was provided so it was defaulted to {DEFAULT_ORIGIN}"
+            )
+
+        tree_data = self._fetch_latest_tree(
+            tree_name=parsed_params.tree_name,
+            branch=parsed_params.branch,
+            origin=origin,
+        )
+
+        if tree_data is None:
+            return create_error_response(
+                error_message=tree_not_found_error_message,
+                status_code=HTTPStatus.NOT_FOUND,
+            )
+
+        base_url = reverse(
+            "treeDetailsView",
+            kwargs={"commit_hash": tree_data["git_commit_hash"]},
+        )
+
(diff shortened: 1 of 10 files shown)


dashboard/src/pages/TreeLatest.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Skeleton } from '@/components/Skeleton';
+
+export const TreeLatest = (): JSX.Element | void => {
+  const searchParams = useSearch({ from: '/tree/$treeName/$branch/' });
+  const navigate = useNavigate();
+  const { treeName, branch } = useParams({ from: '/tree/$treeName/$branch/' });
+  const { isLoading, data, error } = useTreeLatest(treeName, branch);
+
+  if (data) {
+    navigate({
```
WilsonNet, line 24: I think we should probably receive just a full link from the backend, that is more friendly for people hitting the API straight up

dashboard/src/types/tree/Tree.tsx
```diff
@@ -72,3 +72,12 @@ export type Tree = {
     miss: number;
   };
 };
+
+export type TreeLatestResponse = {
```
WilsonNet, line 76: ditto about receiving the full url instead of parts

MarceloRobert (commented):
If a /tree/branch doesn't exist, it just shows the 404 and the error message is only visible in the response of the network tab in dev tools. Also, if I try to go to /tree/mainline/master it doesn't show the same thing as going to it from the tree listing

dashboard/src/pages/TreeLatest.tsx
```diff
@@ -0,0 +1,55 @@
...
+        },
+      },
+    });
+  } else if (isLoading) {
+    return (
+      <Skeleton>
+        <FormattedMessage id="global.loading" />
+      </Skeleton>
+    );
+  } else {
```
Francisco2002, line 47: Since the `SectionError` have a loading message inside it, this component can be removed

murilx: Using the `SectionError` loading message would make it show an "inconsistent" loading message, since it has a different style than this one that's being used in the other pages (and this is not necessarily an error of the `SectionError` component, since it's designed to be used sections of details pages like the misc section in Build and Test details and not for whole pages)

backend/kernelCI_app/views/treeLatest.py
```diff
@@ -0,0 +1,70 @@
...
+        base_url = reverse(
+            "treeDetailsView",
+            kwargs={"commit_hash": tree_data["git_commit_hash"]},
+        )
+        query_params = {
+            "origin": origin,
+            "git_url": tree_data["git_repository_url"],
+            "git_branch": branch,
+        }
+        query_string = f"?{urlencode(query_params)}"
```
murilx, line 67: Newer versions of django added (or will add) a parameter to `reverse` to pass the query parameters directly to it. You can see more here: https://github.com/django/django/pull/18848 and here: https://code.djangoproject.com/ticket/25582#no1

WilsonNet: put this comment inside the code

dashboard/src/types/tree/Tree.tsx
```diff
@@ -72,3 +72,9 @@ export type Tree = {
...
```
WilsonNet, line 76: forgot to update the type here

murilx: Since `apiUrl` is not being used by the frontend, is there still a need to add it here?

backend/requests/latest-tree-get.sh
```diff
@@ -0,0 +1,21 @@
+http 'http://localhost:8000/api/tree/android/android-mainline'
```
WilsonNet, line 1: nit: `tree-lastest-get` is a better name so it get in mixed with other tree request in alphabetical order

WilsonNet (approved):
working great, pre approved with some comments

MarceloRobert (commented):
It a tree/branch only exists in an origin outside of maestro, the origin should be passed as a query parameter to the endpoint. You could add a message saying that the origin was defaulted to maestro in the not found response so that the person knows what happened. Example: http://localhost:5173/tree/c10s/ark-latest (origin=redhat)

MarceloRobert (approved):
Pre-approving, just the default message that is missing but that would also benefit from the future documentation we'll have

PR #734, opened 2025-01-07
Feat: issue details navigation

MarceloRobert:
Adds navigation between the issuesList card and issueDetails page through a new icon on the right
Adds navigation between build/testDetails issueSection and issueDetails page
Adds reactive breadcrumb and topBar to issueDetails using navigation state


dashboard/src/components/Issue/IssueSection.tsx
```diff
@@ -27,20 +27,23 @@ export const NoIssueFound = (): JSX.Element => {
...
     () =>
       data?.map(issue => (
         <Link
-          key={issue.id}
-          to={issue.report_url}
-          target="_blank"
-          className="flex [&:not(:last-child)]:mb-2 [&:not(:last-child)]:border-b [&:not(:last-child)]:pb-2"
+          key={issue.id + issue.version}
+          className="mb-16 flex [&:not(:last-child)]:mb-2 [&:not(:last-child)]:border-b [&:not(:last-child)]:pb-2"
+          to={`/issue/${issue.id}/version/${issue.version}`}
```
murilx, line 43: I believe it would be more aligned with tanstack router library and more aligned with rest of the code if you used the `params` prop of the Link component instead of adding the value directly to the `to` string. Something like:

```js
to="issue/$issueId/version/$versionNumber"
params={{ issueId: issue.id, versionNumber: version.id }}
```

there are some examples in the source code of this, like in the redirect function in `/tree/$treeId/test/$testId`

dashboard/src/components/Cards/IssuesList.tsx
```diff
@@ -29,6 +34,8 @@ interface IIssuesList {
   title: IBaseCard['title'];
   diffFilter: TFilter;
   issueFilterSection: TFilterObjectsKeys;
+  detailsId?: string;
+  pageFrom?: RedirectFrom;
```
murilx, line 38: In the issuesList component you are passing the state as two different props (`detailsId` and `pageFrom`) while in the IssueSection component you are passing the state as only one prop (`historyState`) making the implementation inconsistent, I'd argue that for a better semantic value both of them should receive the state as one prop. However, I understand that having two props makes it easier for the component to be more efficient since passing an object to a prop can cause unnecessary re-render and that would need to be addressed differently like saving the object with an useMemo, for example. Maybe that's something to discuss further

backend/kernelCI_app/helpers/filters.py
```diff
@@ -19,6 +19,7 @@ def is_issue_from_test(incident_test_id: Optional[str], is_known_issue: bool) ->
     return incident_test_id is not None or not is_known_issue
 
 
+# TODO: consider issue_version in the filter
```
Francisco2002, line 22: create a ticket for this

MarceloRobert: Created at #739

Francisco2002 (approved):
LGTM

murilx (approved):
Worked well on my tests

PR #735, opened 2025-01-07
feat: minify URL query parameters

### LINKED ISSUE #725: Add URL query params minification
Minify query parameters in the URL
Probably the best way right now is to hardcode a map for the known values, but it is better to investigate before develop

murilx:
- Added the object `minifiedParams` to map all the params to their minified forms.
- Added the function `minifyParmas` that return a object with the params minified based on the object `minifiedParams`
- Added the object `groupedMinifiedParams` that divide the minified params in five groups based on the depth of the nested objects in searchParams: general, df (for DiffFilter nested object), ti (for TreeInfo nested object), tf (for TableFilter nested object) and value (for minified values instead of keys). This was required to identify different attributes that were minified to the same value (e.g. 't' for treeInfo.treeName, TableFilter.testsTable and diffFilter.trees)
- Added the function `unminifyParams` that return a object with the original param names based on the object `groupedMinfiedParams`.
- Added `minifyParams` to `stringifySearch` so the processing steps are minify -> flatten -> stringify
- Added `unminifyParams` to `parseSearch` so the processing steps are parse -> unflatten -> unminify
- Added tests for the new functions and changed tests for the old functions

Minifying values instead of just keys might be a sensitive action that should be closely observed.

Closes https://github.com/kernelci/dashboard/issues/725

## How to test
- Run `pnpm test` to guarantee all the unit tests are passed.
- Navigate through the application, changing values that are stored in the query parameters (e.g. origin, interval in days and filters).
- Try to enter links directly instead of navigating to the pages to guarantee they are working.

### DIFF
--- a/dashboard/src/utils/search.test.ts
+++ b/dashboard/src/utils/search.test.ts
@@ -8,6 +8,8 @@ import {
   parseSearch,
   stringifySearch,
   isEncodedJSONArrayParam,
+  minifyParams,
+  unminifyParams,
 } from './search';
 
 const KEY_FLAT_CHAR = '|';
@@ -19,8 +21,14 @@ const simpleObject = {
   treeIndexes: [1, 2, 3],
 };
 
-const simpleObjectStringify =
-  '?origin=maestro&intervalInDays=7&treeIndexes[]=1,2,3';
+const simpleObjectMinify = {
+  o: 'maestro',
+  i: 7,
+  // eslint-disable-next-line no-magic-numbers
+  x: [1, 2, 3],
+};
+
+const simpleObjectStringify = '?o=maestro&i=7&x[]=1,2,3';
 
 const nestedObject = {
   origin: 'maestro',
@@ -32,7 +40,7 @@ const nestedObject = {
   },
   treeInfo: {
     treeName: 'android',
-    gitCommitHash: 'hash',
+    headCommitHash: 'hash',
   },
   diffFilter: {
     configs: { defconfig: true },
@@ -42,6 +50,33 @@ const nestedObject = {
   treeIndexes: [0, 1, 2],
 };
 
+const nestedObjectMinify = {
+  o: 'maestro',
+  i: 7,
+  tf: {
+    bt: 'a',
+    b: 'f',
+    t: 'a',
+  },
+  ti: {
+    t: 'android',
+    ch: 'hash',
+  },
+  df: {
+    c: { defconfig: true },
+    a: { arm: true },
+    tp: 'amlogic',
+  },
+  x: [0, 1, 2],
+};
+
+const nestedObjectStringify =
+  '?o=maestro&i=7' +
+  '&tf|bt=a&tf|b=f&tf|t=a' +
+  '&ti|t=android&ti|ch=hash' +
+  '&df|c|defconfig=true&df|a|arm=true&df|tp=amlogic' +
+  '&x[]=0,1,2';
+
 const flatObject = {
   origin: 'maestro',
   intervalInDays: 7,
@@ -49,19 +84,26 @@ const flatObject = {
   'tableFilter|buildsTable': 'failed',
   'tableFilter|testsTable': 'all',
   'treeInfo|treeName': 'android',
-  'treeInfo|gitCommitHash': 'hash',
+  'treeInfo|headCommitHash': 'hash',
   'diffFilter|configs|defconfig': true,
   'diffFilter|archs|arm': true,
   'diffFilter|testPath': 'amlogic',
   treeIndexes: [0, 1, 2],
 };
 
-const nestedObjectStringify =
-  '?origin=maestro&intervalInDays=7' +
-  '&tableFilter|bootsTable=all&tableFilter|buildsTable=failed&tableFilter|testsTable=all' +
-  '&treeInfo|treeName=android&treeInfo|gitCommitHash=hash' +
-  '&diffFilter|configs|defconfig=true&diffFilter|archs|arm=true&diffFilter|testPath=amlogic' +
-  '&treeIndexes[]=0,1,2';
+const flatObjectMinify = {
+  o: 'maestro',
+  i: 7,
+  'tf|bt': 'a',
+  'tf|b': 'f',
+  'tf|t': 'a',
+  'ti|t': 'android',
+  'ti|ch': 'hash',
+  'df|c|defconfig': true,
+  'df|a|arm': true,
+  'df|tp': 'amlogic',
+  x: [0, 1, 2],
+};
 
 describe('isEncodedArrayParam', () => {
   const emptyJSONArrayStr = 'treeIndexes=[]';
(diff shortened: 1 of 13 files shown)


dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+  bootStatus: 'bts',
+  testStatus: 'ts',
+  testPath: 'tp',
+  bootPath: 'bp',
+  buildDurationMax: 'bdx',
+  buildDurationMin: 'bdm',
+  bootDurationMax: 'btdx',
+  bootDurationMin: 'btdm',
+  testDurationMax: 'tdx',
+  testDurationMin: 'tdm',
```
Francisco2002, line 174: nit:
- max -> greater -> gr / gt / g
- min -> less -> le / ls / l

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+  hardware: 'h',
+  trees: 't',
+  buildPlatform: 'bp',
+  bootPlatform: 'btp',
+  testPlatform: 'ttp',
```
Francisco2002, line 179:
```suggestion
  buildPlatform: 'bpf',
  bootPlatform: 'btpf',
  testPlatform: 'tpf',
```

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
   }
   return result;
 };
+
+const minifiedParams = {
```
WilsonNet, line 134: we need a source of truth for the query parameters key

 but the devs need to know which keys are being minified

so, it needs to be something like

```typescript
type QueryParametersKeyNames = 'origin' | intervalInDays | ...
```
And restrict the key name types both in this `minifiedParameters` and all the schemas in the routes

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+
+  // Values
+  all: 'a',
+  success: 's',
+  failed: 'f',
+  inconclusive: 'i',
+  valid: 'v',
+  invalid: 'iv',
+  null: 'n',
+};
```
WilsonNet, line 192: I'm pretty sure you need as const to the type in the next line to work

dashboard/src/utils/search.test.ts
```diff
@@ -49,19 +57,53 @@ const flatObject = {
...
-  '&treeInfo|treeName=android&treeInfo|gitCommitHash=hash' +
-  '&diffFilter|configs|defconfig=true&diffFilter|archs|arm=true&diffFilter|testPath=amlogic' +
-  '&treeIndexes[]=0,1,2';
+  '?o=maestro&i=7' +
+  '&tf|bt=a&tf|b=f&tf|t=a' +
+  '&ti|t=android&ti|ch=hash' +
+  '&df|c|defconfig=true&df|a|arm=true&df|tp=amlogic' +
+  '&x[]=0,1,2';
+
+const nestedObjectMinify = {
```
WilsonNet, line 88: nit: but I think it makes more sense to put this closer to `nestedObject`

dashboard/src/utils/search.test.ts
```diff
@@ -49,19 +57,53 @@ const flatObject = {
...
   'treeInfo|treeName': 'android',
-  'treeInfo|gitCommitHash': 'hash',
+  'treeInfo|headCommitHash': 'hash',
   'diffFilter|configs|defconfig': true,
   'diffFilter|archs|arm': true,
   'diffFilter|testPath': 'amlogic',
   treeIndexes: [0, 1, 2],
 };
 
+const flatObjectMinify = {
```
WilsonNet, line 67: ditto about moving this closer to `flatObject`

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+type MinifiedParamsKeys = keyof typeof minifiedParams;
+
+export const minifyParams = (
+  searchParams: Record<string, unknown>,
+): Record<string, unknown> => {
+  const result: Record<string, unknown> = {};
+  for (const key in searchParams) {
+    if (Object.prototype.hasOwnProperty.call(searchParams, key)) {
+      const newKey =
+        key in minifiedParams ? minifiedParams[key as MinifiedParamsKeys] : key;
```
WilsonNet, line 202: I'm pretty sure with the `as const` in `minifiedParams` typescript can figure out the type by itself here and you won't need the as `MinifiedParamsKeys`

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+      if (isStringRecord(searchParams[key])) {
+        result[newKey] = minifyParams(searchParams[key]);
+      } else {
+        const value =
+          typeof searchParams[key] === 'string' &&
+          searchParams[key] in minifiedParams
+            ? minifiedParams[searchParams[key] as MinifiedParamsKeys]
```
WilsonNet, line 209: ditto about the as, I think the type narrowing will take care of this

dashboard/src/utils/search.ts
murilx: Maybe ceiling (c) and floor (f) would also be acceptable?

dashboard/src/utils/search.test.ts
murilx: `flatObject` is directly above `flatObjectMinify`

dashboard/src/utils/search.ts
```diff
@@ -122,3 +130,155 @@ export const unflattenObject = (
...
+    }
+  }
+  return result;
+};
+
+const GENERAL_LAST_ENTRY = 12;
+const TI_LAST_ENTRY = 17;
+const TF_LAST_ENTRY = 20;
+const DF_LAST_ENTRY = 41;
+const VALUE_LAST_ENTRY = 49;
```
WilsonNet, line 222: having hardcoded numbers is kind of cumbersome and error prone since everytime the dev adds a parameter needs to change everything here, can't you find a solution that does  not need it? (it can be a prefix, you can use typescript template literal string to still get the types)
or can be anything else that doesn't require someone to keep counting the keys

WilsonNet (commented):
it is working really nice, good job, a little more things to go

dashboard/src/utils/search.ts
murilx: Even with `as const`, removing `as MinifiedParamsKeys` raises an error. Maybe there's other way around that I'm not familiar with

dashboard/src/utils/search.ts
murilx: I did a solution using the length of each zod schema to determine where the slice should happen. Do you think that's acceptable?

dashboard/src/utils/search.ts
murilx: Same as the last comment

dashboard/src/utils/search.ts
WilsonNet: ok lets keep this way for now

WilsonNet (approved):
nice work

dashboard/src/utils/search.ts
```diff
@@ -122,3 +136,162 @@ export const unflattenObject = (
   }
   return result;
 };
+
+const minifiedParams = {
```
MarceloRobert: instead of having everything in a single variable and slicing it later, I think it is better to have the params in separate variables and combine them into another one with spread. That way you can use the individual sections more easily while keeping a variable with all of them.

MarceloRobert (approved):
worked well on my tests, great job

PR #738, opened 2025-01-08
feat/tree summary endpoint

WilsonNet:
# Description
This PR achieves the following:

- Refactor the Tree Details endpoint into some helper methods
- Add a Tree Summary endpoint
- Add Swagger 
- Add swagger docs for Tree Summary Endpoint

## How to test

1. See if regular tree details are still working
2. Check if localhost page return the same results as the staging page
3. Use the request file for the tree summary
4. Check the swagger by reaching the [new swagger-io url ](http://localhost:8000/api/schema/swagger-ui)

### DIFF
--- a/backend/kernelCI_app/helpers/treeDetails.py
+++ b/backend/kernelCI_app/helpers/treeDetails.py
@@ -1,0 +1,459 @@
+from collections.abc import Callable
+from typing import Any
+from kernelCI_app.helpers.filters import (
+    should_increment_test_issue,
+    UNKNOWN_STRING,
+)
+from kernelCI_app.utils import (
+    extract_error_message,
+    create_issue,
+)
+from kernelCI_app.helpers.misc import (
+    handle_build_misc,
+    handle_environment_misc,
+    build_misc_value_or_default,
+    env_misc_value_or_default,
+)
+from kernelCI_app.cache import getQueryCache, setQueryCache
+from django.db import connection
+
+
+def get_tree_details_data(request, commit_hash):
+    cache_key = "treeDetails"
+
+    origin_param = request.GET.get("origin")
+    git_url_param = request.GET.get("git_url")
+    git_branch_param = request.GET.get("git_branch")
+
+    params = {
+        "commit_hash": commit_hash,
+        "origin_param": origin_param,
+        "git_url_param": git_url_param,
+        "git_branch_param": git_branch_param,
+    }
+
+    rows = getQueryCache(cache_key, params)
+    if rows is None:
+        query = """
+        SELECT
+                tests.build_id AS tests_build_id,
+                tests.id AS tests_id,
+                tests.origin AS tests_origin,
+                tests.environment_comment AS tests_environment_comment,
+                tests.environment_misc AS tests_environment_misc,
+                tests.path AS tests_path,
+                tests.comment AS tests_comment,
+                tests.log_url AS tests_log_url,
+                tests.status AS tests_status,
+                tests.waived AS tests_waived,
+                tests.start_time AS tests_start_time,
+                tests.duration AS tests_duration,
+                tests.number_value AS tests_number_value,
+                tests.misc AS tests_misc,
+                tests.environment_compatible AS tests_environment_compatible,
+                builds_filter.*,
+                incidents.id AS incidents_id,
+                incidents.test_id AS incidents_test_id,
+                incidents.present AS incidents_present,
+                issues.id AS issues_id,
+                issues.version AS issues_version,
+                issues.comment AS issues_comment,
+                issues.report_url AS issues_report_url
+        FROM
+            (
+                SELECT
+                    builds.checkout_id AS builds_checkout_id,
+                    builds.id AS builds_id,
+                    builds.comment AS builds_comment,
+                    builds.start_time AS builds_start_time,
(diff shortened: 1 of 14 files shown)


backend/kernelCI_app/urls.py
```diff
@@ -33,6 +33,10 @@ def viewCache(view):
          views.TreeDetails.as_view(),
          name="treeDetailsView"
          ),
+    path("tree/<str:commit_hash>/summary",
+         views.TreeDetailsSummary.as_view(),
```
MarceloRobert, line 37: Just to confirm, you've meant to not use `viewCache()` here, right?

backend/kernelCI_app/typeModels/databases.py
```diff
@@ -1,4 +1,107 @@
...
+    log_excerpt: Optional[str] = None
+    status: Optional[DatabaseStatusValues] = None
+    waived: Optional[bool] = None
+    start_time: Optional[datetime] = None
+    duration: Optional[float] = None
+    output_files: Optional[Dict] = None
+    misc: Optional[Dict] = None
+
+
+type incident_id_pydantic = str
```
MarceloRobert, line 90: I don't see the need to create a new type that is just a str

backend/kernelCI_app/urls.py
```diff
@@ -76,4 +81,8 @@ def viewCache(view):
     path("issue/<str:issue_id>/version/<str:version>/builds",
          viewCache(views.IssueDetailsBuilds),
          name="issueDetailsBuilds"),
+    # DRF Spectacular
+    path('schema/', SpectacularAPIView.as_view(), name='schema'),
+    path('schema/swagger-ui/', SpectacularSwaggerView.as_view(url_name='schema'), name='swagger-ui'),
+    path('schema/redoc/', SpectacularRedocView.as_view(url_name='schema'), name='redoc'),
```
murilx, line 87: aren't there any advantages of using `viewCache` in this case?

backend/kernelCI_app/urls.py
WilsonNet: Yes, this will reduce inconsistency due to cache in case we don't have a way to invalidate

backend/kernelCI_app/urls.py
WilsonNet: I'm pretty sure they those pages are saved in phisical space

backend/kernelCI_app/helpers/treeDetails.py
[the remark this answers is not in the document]
WilsonNet: This already took too much time, let's keep this way for now

backend/kernelCI_app/views/treeDetailsSummaryView.py
```diff
@@ -0,0 +1,173 @@
...
+        test_id = row_data["test_id"]
+
+        if decide_if_is_boot_filtered_out(self, row_data):
+            return
+
+        process_builds_issue(self, row_data)
+
+        if test_id in self.processedTests:
+            return
+        self.processedTests.add(test_id)
```
murilx, line 75: Shouldn't `process_boots_summary` be called in the end of this function?

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -67,448 +67,94 @@ def __init__(self):
...
-                )
-        elif testStatus == "FAIL":
-            self.failedBootsWithUnknownIssues += 1
-
-        if testId in self.processedTests:
+        if test_id in self.processedTests:
             return
-        self.processedTests.add(testId)
+        self.processedTests.add(test_id)
+        self.bootHistory.append(history_item)
```
murilx, line 82: ditto about `process_boots_summary`

backend/kernelCI_app/typeModels/treeDetails.py
MarceloRobert, line 1: nit: for the types where they represent the counting of occurrences (like Status, TestEnvironmentCompatible, TestEnvironmentMisc) you could rename them to include "Count" at the end of the name so that they don't get confused with objects with specific properties

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -67,448 +68,95 @@ def __init__(self):
...
+        history_item = row_data["history_item"]
 
-        if is_boot_filter_out:
+        if decide_if_is_boot_filtered_out(self, row_data):
             return
 
-        can_insert_issue = should_increment_test_issue(
-            issue_id=issue_id, incident_test_id=incident_test_id
-        )
+        process_builds_issue(self, row_data)
```
MarceloRobert, line 78: shouldn't this be `process_boots_issue`?

backend/kernelCI_app/helpers/treeDetails.py
MarceloRobert, line 1: The issuesTable (or "processed issues" in a way) uses a tuple of `(issue_id, issue_version)` for its key, but you are removing the parenthesis here, making it two separate parameters. Whenever you see `issue_id, issue_version`, it should be `(issue_id, issue_version)` - with parenthesis

MarceloRobert (approved):
Worked on my tests, looks good to me

murilx (approved):
Nice job! Worked well on my tests

PR #748, opened 2025-01-10
Add priority logic to commit tag

### LINKED ISSUE #697: Review hash name priority for commits
Commits should have this display priority

commit tag (from database, first element of array) -> hover commit hash
if does not have tag, show commit hash -> hover show commit name 

**Disclaimer, a lot of stuff is using git commit name believing that is git commit tag, double check**

murilx:
Closes #697 

## How to test
- Check if the commit tags are following the logic described in #697. 

Example page (Tree Details): http://localhost:5173/tree/f4592015c1a994f1a92fa984bd008924e88f6c3a?treeInfo%7CcommitName=renesas-devel-2025-01-06-v6.13-rc6&treeInfo%7CgitBranch=master&treeInfo%7CgitUrl=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fgeert%2Frenesas-devel.git&treeInfo%7CheadCommitHash=f4592015c1a994f1a92fa984bd008924e88f6c3a&treeInfo%7CtreeName=renesas

Example page (Hardware Details): http://localhost:5173/hardware/radxa%2Crock-5b?et=1736541000&hs=radxa&st=1736109000

## Visual Reference
Tree Listing Page
Captura de tela de 2025-01-10 10-03-58

Tree Details
Captura de tela de 2025-01-10 10-06-34

Hardware Details
image

Build Details
image

Test Details
image

### DIFF
--- a/backend/kernelCI_app/typeModels/hardwareDetails.py
+++ b/backend/kernelCI_app/typeModels/hardwareDetails.py
@@ -42,4 +42,6 @@ class CommitHistoryValidCheckout(BaseModel):
     tree_name: str
     git_repository_branch: str
     git_repository_url: str
+    git_commit_tags: Optional[List[str]] = []
+    git_commit_name: Optional[str]
     start_time: datetime
--- a/backend/kernelCI_app/views/treeDetailsView.py
+++ b/backend/kernelCI_app/views/treeDetailsView.py
@@ -66,6 +66,7 @@ class TreeDetails(View):
         self.failed_builds_with_unknown_issues = 0
         self.processed_build_issues: IssueDict = {}
         self.tree_url = ""
+        self.git_commit_tags = []
 
     def setup_filters(self):
         self.filterTestDurationMin = self.filterParams.filterTestDurationMin
@@ -115,13 +116,14 @@ class TreeDetails(View):
             "checkout_id": currentRow[28],
             "checkout_git_repository_url": currentRow[29],
             "checkout_git_repository_branch": currentRow[30],
-            "incident_id": currentRow[31],
-            "incident_test_id": currentRow[32],
-            "incident_present": currentRow[33],
-            "issue_id": currentRow[34],
-            "issue_version": currentRow[35],
-            "issue_comment": currentRow[36],
-            "issue_report_url": currentRow[37],
+            "checkout_git_commit_tags": currentRow[31],
+            "incident_id": currentRow[32],
+            "incident_test_id": currentRow[33],
+            "incident_present": currentRow[34],
+            "issue_id": currentRow[35],
+            "issue_version": currentRow[36],
+            "issue_comment": currentRow[37],
+            "issue_report_url": currentRow[38],
         }
 
         environment_misc = handle_environment_misc(
@@ -398,6 +400,7 @@ class TreeDetails(View):
                 else:
                     self.__processNonBootsTest(row_data)
 
+        self.git_commit_tags = row_data["checkout_git_commit_tags"]
         self.testIssues = convert_issues_dict_to_list(self.testIssuesTable)
         self.bootIssues = convert_issues_dict_to_list(self.bootsIssuesTable)
         self.build_summary = create_details_build_summary(self.builds)
@@ -468,7 +471,8 @@ class TreeDetails(View):
                             SELECT
                                 checkouts.id AS checkout_id,
                                 checkouts.git_repository_url AS checkouts_git_repository_url,
-                                checkouts.git_repository_branch AS checkouts_git_repository_branch
+                                checkouts.git_repository_branch AS checkouts_git_repository_branch,
+                                checkouts.git_commit_tags
                             FROM
                                 checkouts
                             WHERE
@@ -529,6 +533,7 @@ class TreeDetails(View):
                 "buildsIssues": self.build_issues,
                 "failedBuildsWithUnknownIssues": self.failed_builds_with_unknown_issues,
                 "treeUrl": self.tree_url,
+                "git_commit_tags": self.git_commit_tags,
             },
             safe=False,
         )
--- a/backend/kernelCI_app/views/treeView.py
+++ b/backend/kernelCI_app/views/treeView.py
@@ -38,6 +38,17 @@ class TreeView(View):
                 checkouts.git_repository_branch,
                 checkouts.git_repository_url,
                 checkouts.git_commit_hash,
+                CASE
+                    WHEN COUNT(DISTINCT checkouts.git_commit_tags) > 0 THEN
+                    COALESCE(
+                        ARRAY_AGG(DISTINCT checkouts.git_commit_tags) FILTER (
+                            WHERE checkouts.git_commit_tags IS NOT NULL
+                            AND checkouts.git_commit_tags::TEXT <> '{}'
+                        ),
+                        ARRAY[]::TEXT[]
+                    )
+                    ELSE ARRAY[]::TEXT[]
+                END AS git_commit_tags,
                 MAX(checkouts.git_commit_name) AS git_commit_name,
                 MAX(checkouts.start_time) AS start_time,
                 COUNT(DISTINCT CASE WHEN builds.valid = true THEN builds.id END) AS valid_builds,
--- a/dashboard/src/components/Tooltip/CommitTagTooltip.tsx
+++ b/dashboard/src/components/Tooltip/CommitTagTooltip.tsx
@@ -1,0 +1,56 @@
+import CopyButton from '@/components/Button/CopyButton';
+
+import { Tooltip, TooltipTrigger, TooltipContent } from './Tooltip';
+
+interface gitValues {
+  commitTags?: string[];
+  commitHash?: string;
+  commitName?: string;
+}
+
+export const gitCommitValueSelector = ({
+  commitTags,
+  commitHash,
+  commitName,
+}: gitValues): { content: string; hover: string } => {
+  let content;
+  let hover;
+  if (commitTags && commitTags.length > 0) {
+    content = commitTags[0];
+    hover = commitHash ?? commitName ?? '';
+  } else {
+    content = commitHash ?? commitName ?? '';
+    hover = commitName ?? '';
+  }
+  return { content, hover };
+};
+
+interface ICommitTagTooltip {
+  commitName?: string;
+  commitHash?: string;
+  commitTags?: string[];
+  copyButton?: boolean;
+}
+
+export const CommitTagTooltip = ({
+  commitName,
+  commitHash,
+  commitTags,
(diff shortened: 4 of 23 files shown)


MarceloRobert (commented):
Looks good, some comments

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
```diff
@@ -0,0 +1,29 @@
+import { Tooltip, TooltipTrigger, TooltipContent } from './Tooltip';
+
+export const CommitTagTooltip = ({
+  commitName,
+  commitHash,
+  commitTags,
+}: {
+  commitName: string | undefined;
+  commitHash: string | undefined;
+  commitTags: string[] | undefined;
```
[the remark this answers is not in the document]
murilx: In this case I'd like to enforce that all the props where passed to the component, ensuring that if some of these values are not being used, it's because they are absent in the database instead of not being passed to the component. If I used `var?: type` the component could be called without one of the props, that's not the case with this notation as far as I'm concerned. However, we could discuss if that is necessary

backend/kernelCI_app/typeModels/hardwareDetails.py
```diff
@@ -37,9 +37,17 @@ class CommitHistoryPostBody(BaseModel):
     commitHeads: List[CommitHead]
 
 
+def process_tags(value: Any) -> Any:
+    if value is None:
+        return []
+    return value
```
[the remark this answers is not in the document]
murilx: Actually, after some thought I think I might be able to get the same result without this function by defining `git_commit_tags` as 

```python
    git_commit_tags: List[str] | None = []
```

I'll verify if that's the case. If not, I'll pass this function to a different file

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -468,7 +471,8 @@ def get(self, request, commit_hash: str | None):
                             SELECT
                                 checkouts.id AS checkout_id,
                                 checkouts.git_repository_url AS checkouts_git_repository_url,
-                                checkouts.git_repository_branch AS checkouts_git_repository_branch
+                                checkouts.git_repository_branch AS checkouts_git_repository_branch,
+                                checkouts.git_commit_tags
```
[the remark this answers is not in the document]
murilx: AFAIK it's not necessary. However, I could add this to make it consistent with the rest of the query

backend/kernelCI_app/typeModels/hardwareDetails.py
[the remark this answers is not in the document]
murilx: That seems to be the case: https://docs.pydantic.dev/2.10/concepts/fields/#mutable-default-values

backend/kernelCI_app/views/treeView.py
```diff
@@ -38,6 +38,17 @@ def get(self, request):
...
                 checkouts.git_repository_url,
                 checkouts.git_commit_hash,
+                CASE
+                    WHEN COUNT(DISTINCT checkouts.git_commit_tags) > 0 THEN
+                    COALESCE(
+                        ARRAY_AGG(DISTINCT checkouts.git_commit_tags) FILTER (
+                            WHERE checkouts.git_commit_tags IS NOT NULL
+                            AND checkouts.git_commit_tags::TEXT <> '{}'
+                        ),
+                        ARRAY[]::TEXT[]
```
Francisco2002, line 48: nice finding!!!!!!!!
for reference: [9.20. Aggregate Functions](https://www.postgresql.org/docs/9.5/functions-aggregate.html)

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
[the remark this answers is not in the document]
Francisco2002: In this case, why you don't use `type | null`?

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+  copyButton?: boolean;
+}): JSX.Element => {
+  let hover;
+  let content;
+
+  if (commitTags !== undefined && commitTags.length > 0) {
```
Francisco2002, line 19:
```suggestion
  if (commitTags?.length > 0) {
```

dashboard/src/components/TreeListingPage/TreeListingPage.tsx
```diff
@@ -84,6 +86,13 @@ const TreeListingPage = ({ inputFilter }: ITreeListingPage): JSX.Element => {
         return {
           commitHash: tree.git_commit_hash ?? '',
           commitName: tree.git_commit_name ?? '',
+          commitTag: (
+            <CommitTagTooltip
+              commitHash={tree.git_commit_hash ?? undefined}
+              commitName={tree.git_commit_name ?? undefined}
+              commitTags={tree.git_commit_tags ?? undefined}
```
Francisco2002, line 93: let's discuss about using `null` type for this props

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -132,6 +150,28 @@ const CommitSelector = ({
...
+          {gitValues[`${treeIndex}_${commitHash}`]?.gitCommitHash && (
+            <>
+              {gitValues[`${treeIndex}_${commitHash}`]?.gitCommitHash} <br />
+            </>
+          )}
+          {gitValues[`${treeIndex}_${commitHash}`]?.gitCommitName && (
+            <>
+              {gitValues[`${treeIndex}_${commitHash}`]?.gitCommitName} <br />
+            </>
+          )}
```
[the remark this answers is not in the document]
murilx: I tried creating a string that joinend all `gitValues` using `\n` but it still rendered everything as one line instead of correctly skipping to the next line for every `\n`. It was something like this

```js
const tooltipContent = [
  gitCommitTags,
  gitCommitHash,
  gitCommitName,
].filter(Boolean).join('\n');
```

So I combined both strategies, using a tooltipContent array and mapping it to a JSX.Element `<>{value}<br /></>`

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
murilx: This modification does not seem to work properly

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export const gitCommitValueSelector = ({
+  commitTags,
+  commitHash,
+  commitName,
+}: gitValues): { content: string; hover: string } => {
+  let content;
+  let hover;
+  if (commitTags !== undefined && commitTags.length > 0) {
```
WilsonNet, line 18: you got javascripted here, `if (commitTags !== undefined && commitTags.length > 0) {` 

won't work if `commitTags` is `null`,  you can do `if(commitTags && commitTags.length >0)`

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
```diff
@@ -0,0 +1,56 @@
...
+    content = commitHash ?? commitName ?? '';
+    hover = commitName ?? '';
+  }
+  return { content, hover };
+};
+
+interface ICommitTagTooltip {
+  commitName: string | undefined;
+  commitHash: string | undefined;
+  commitTags: string[] | undefined;
```
WilsonNet, line 31:
```suggestion
  commitName?: string;
  commitHash?: string;
  commitTags?: string[] ;
```

WilsonNet: or maybe you can go with
```
commitName?: string; | null
  commitHash?: string; | null
  commitTags?: string; | null
```
so typescript will catch that error

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
murilx: Would typescript allow that I pass a `null` commitTags as a parameter even though the type of the parameter is `commitTags?: string[]`?

dashboard/src/components/TreeListingPage/TreeTable.tsx
```diff
@@ -285,7 +292,25 @@ export function TreeTable({ treeTableRows }: ITreeTable): JSX.Element {
...
+                <CopyButton
+                  value={
+                    gitCommitValueSelector({
+                      commitHash: cell.row.original.commitHash,
+                      commitName: cell.row.original.commitName,
+                      commitTags: cell.row.original.commitTag,
+                    }).content
+                  }
+                />
+              </TableCell>
```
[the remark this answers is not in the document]
murilx: Basically, I couldn't find a way of having a link to tree details in the tooltip but do not have a link in the copy button (if that where the case the Link would be a priority over the copy button and go to the tree details page instead of adding the commit hash to the clipboard). I tried modifying `CommitTagTooltip` to receive `LinkProps` so that I could wrap the tooltip in a `Link` component inside the `CommitTagTooltip` component itself. The problem was that, for this to work, I'd need that the cell of the 'commitTag' column to be a function that receives the `LinkProps` and returns the component, but I couldn't find a way to do something like that. For some reason, when using `cell.getValue()` I received an empty array instead of the function I defined, so I decided to go for this solution that was easier to implement. I could try working more on the other idea, if you find this implementation unsatisfying, or I can add an explanation of this type to the code itself as well

WilsonNet (commented):
Nice, just some more comments

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
WilsonNet: I remember to have commented this but it does not showing, but you got javascripted here:
`  if (commitTags !== undefined && commitTags.length > 0) {`
`commitTags !== undefined` won't be caught if the value is null

`commitTags?.length > 0` should work, I'm not sure why this is not working 

but you can go with `commitTags && commitTags.length > 0`

dashboard/src/components/Tooltip/CommitTagTooltip.tsx
murilx: Done

dashboard/src/components/TreeListingPage/TreeTable.tsx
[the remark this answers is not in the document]
murilx: Added the comment to the code as well

Francisco2002 (approved):
NICE!!!

PR #769, opened 2025-01-16
Feat: add alternative route for /tree

### LINKED ISSUE #766: /t as an alternative route for /tree
From this comment on discord: https://discord.com/channels/1245820301053530313/1301896040349433957/1329065228452888627

Francisco2002:
The alternative routes `/t` will be redirect to a corresponding `/tree` route

**Example**

- http://localhost/t -> http://localhost/tree
- http://localhost/t/${commit_hash} -> http://localhost/tree/${commit_hash}

Close #766


dashboard/src/routes/tree/$treeName/$branch/route.tsx
```diff
@@ -0,0 +1,3 @@
+import { createFileRoute } from '@tanstack/react-router';
+
+export const Route = createFileRoute('/tree/$treeName/$branch')();
```
WilsonNet, line 3: why this route is being created?

dashboard/src/routes/(alternatives)/t/$treeName/index.tsx
```diff
@@ -0,0 +1,12 @@
+import { createFileRoute, redirect } from '@tanstack/react-router';
+
+export const Route = createFileRoute('/(alternatives)/t/$treeName/')({
+  loaderDeps: ({ search }) => ({ search }),
+  loader: ({ deps, params }) => {
+    throw redirect({
+      to: '/tree/$treeName',
```
WilsonNet, line 7:
```suggestion
      to: '/tree/$treeName/',
```

that way tou don't need to create that  extra `route`

when it finishes with `/` is a leaf

dashboard/src/routes/tree/$treeName/$branch/route.tsx
Francisco2002: When I try to access `/t/${treeName}/${branch}`, is redirect to `/t/${treeName}`. Add this file resolve this problem

dashboard/src/routes/(alternatives)/t/$treeName/index.tsx
Francisco2002: When I put this code, I receive a type error and it suggests `/tree/$treeName`

murilx: The reason behind this is that `/tree/$treeName` does not have an index, only a route. To solve this you can create the index for `/tree/$treeName` or keep `to: '/tree/$treeName'. Maybe the former would be the correct approach here, I'm not so sure it's needed though

WilsonNet (dismissed):
working nice here

WilsonNet (approved):

murilx (approved):
Worked well on my tests

Issue #86, opened 2024-08-05
Generate schema using open api standard


bdilly:
https://www.django-rest-framework.org/topics/documenting-your-api/#drf-spectacular
https://drf-spectacular.readthedocs.io/en/latest/blueprints.html#pydantic
- [x] Generate schema with drf-spectacular or some other tool
- [x] Add swagger
- [x] Add Request schema
- [x] Add Response schema

**UPDATE: Right now we have added in the app but it need to cover all endpoints**

PR #779, opened 2025-01-17
Feat: hardware details summary

MarceloRobert:
Refactors some typing in the backend
Adds hardwareDetailsSummary endpoint


backend/kernelCI_app/utils.py
```diff
@@ -29,7 +18,7 @@ def create_issue(
 ) -> Issue:
     return {
         "id": issue_id,
-        "version": issue_version,
+        "version": str(issue_version),
```
murilx, line 21: This conversion, in theory, shouldn't be necessary since the argument is already a string. In this situation any conversions would make more sense when the argument is being passed to the function, like

```python
new_issue = create_issue(issue_id, str(issue_version), issue_comment, issue_report_url)
```

To make it consistent update the type of `issue_version` to account for it possibly not being a string and keep the conversion inside the function, or convert the value that's being passed to the function like I exemplified earlier

backend/kernelCI_app/views/issuesView.py
```diff
@@ -5,12 +5,11 @@
 
 from kernelCI_app.helpers.errorHandling import create_error_response
 from kernelCI_app.utils import (
-    Issue,
     convert_issues_dict_to_list,
     create_issue,
 )
 from http import HTTPStatus
-from kernelCI_app.typeModels.issueDetails import IssueDict
+from kernelCI_app.typeModels.issues import Issue, IssueDict
```
murilx, line 12: We should add `IssueDetailsPathParameters` validation to this view as well. Commenting this just so we can remember later, I don't this should be added in this PR

backend/kernelCI_app/helpers/hardwareDetails.py
```diff
@@ -589,12 +596,13 @@ def get_filter_options(
     return list(configs), list(archs), list(compilers)
 
 
-def is_record_tree_selected(*, record, tree: Optional[Dict], is_all_selected: bool):
+def is_record_tree_selected(*, record, tree: Tree, is_all_selected: bool):
```
murilx, line 797: nit: return is bool

backend/kernelCI_app/views/issuesView.py
MarceloRobert: This isn't the IssueDetails page, but it will be covered under #86

backend/kernelCI_app/utils.py
MarceloRobert: In hardware, since we are getting the records typed from the database definitions, they arrive as int and because of the type validation it was complaining that it was always converting from int to str. But I understand that it should be used from the calling, not from here since Issue is also used in treeDetails where it doesn't come as int

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+        self.builds_summary.unknown_issues = self.processed_issues["build"][
+            "failedWithUnknownIssues"
+        ]
+        self.boots_summary.unknown_issues = self.processed_issues["boot"][
+            "failedWithUnknownIssues"
+        ]
+        self.tests_summary.unknown_issues = self.processed_issues["test"][
+            "failedWithUnknownIssues"
+        ]
```
WilsonNet, line 195: this part seems a little redundant, you are mapping 1x1, it could use the instance property from the start

PR #794, opened 2025-01-20
fix(get-tree-details-data): fix return position

### LINKED ISSUE #785: Make staging great again
- [x] Add validation to TreeDetailsSummaryView (not needed)
- [x] Possibly change validation to TreeDetails*View (if `rows` is None, then `len(rows) == 0` would return an error)
- [x] Verify if the cache is being correctly updated when multiple endpoints are accessing it at the same time
- [x] Verify if the query is correct (unlikely to be the problem here, but it's better safe then sorry)
 
The problem is caused only when cache is active  so to reproduce the error change this in `settings.py`:

```python
if DEBUG:
    CORS_ALLOWED_ORIGIN_REGEXES = [
        r"^http://localhost",  # dashboard dev server
    ]
    SESSION_COOKIE_SECURE = False
    CSRF_COOKIE_SECURE = False
    SECURE_SSL_REDIRECT = False
    SECURE_HSTS_SECONDS = 3600
    # CACHE_TIMEOUT = 0
    CACHE_TIMEOUT = 3600
```

murilx:
Closes #785

## How to test
- Test with no cache (default behavior in development environment)
- Test with cache by changing the `settings.py` file to: 
```python
if DEBUG:
    CORS_ALLOWED_ORIGIN_REGEXES = [
        r"^http://localhost",  # dashboard dev server
    ]
    SESSION_COOKIE_SECURE = False
    CSRF_COOKIE_SECURE = False
    SECURE_SSL_REDIRECT = False
    SECURE_HSTS_SECONDS = 3600
    # CACHE_TIMEOUT = 0
    CACHE_TIMEOUT = 3600
```

### DIFF
--- a/backend/kernelCI_app/helpers/treeDetails.py
+++ b/backend/kernelCI_app/helpers/treeDetails.py
@@ -115,7 +115,7 @@ def get_tree_details_data(request, commit_hash):
             rows = cursor.fetchall()
             setQueryCache(cache_key, params, rows)
 
-        return rows
+    return rows
 
 
 def get_current_row_data(current_row: dict) -> dict:


Francisco2002 (approved):
Nice job!!!!

PR #830, opened 2025-01-22
fix/lazy load summary to full

### LINKED ISSUE #791: Tree - Summary -> Full
Right now the tree details is making four request for the same context (Summary, Builds, Boots, Tests)

We can just two, Summary to Full

WilsonNet:
# Description
Make lazy load be Summary to Full instead of Summary to 3 endpoints

## How to test
Enter in any tree details with the Network tab open and see how it goes.
Compare with the same result on staging

Closes #791


MarceloRobert (approved):
Some renaming nits, but it worked well on my tests. The only downside is that with no cache it will wait for the query in both requests, but that's not an issue for production/staging, only dev env. LGTM

PR #836, opened 2025-01-23
Feat: logspec section

### LINKED ISSUE #804: Optimistic logspec in Issue Details pages
Create a generic logspec  for the most common pattern of logspec

``` 
const LogSpecField  ({ fieldName: fieldValue} ) => {
   case (log_excerpt):
           <CodeViewer />
   default:

}
```

murilx:
- Added `children` prop to Section. The children will be showed after all links in a single column instead of two as the links do
- Added `contentHandler` function to handle the different types of data that can be showed in a misc section of a details page and also handles special fields like 'log_excerpt' with pattern matching
- Added the `getLogspecSection` function that creates the json needed by `SectionGroup` to create the section correctly. It creates a titled subsection for every object in the logpsec field in misc up to depth 2 and Nested objects beyond depth 2 can be viewed through the Json viewer

Closes #804 

## Visual Reference
image


WilsonNet:
we need a minimum height so the full screen button doesn't end up floating in the middle of nowhere
http://localhost:5173/issue/maestro%3Ada694c56147298d223ee432ad8d6a8ee311b773a/version/1?p=bt&ti%7Cc=ASB-2025-01-05_12-5.4-2-g8cc44cf295628&ti%7Cch=8cc44cf2956283500f128e77f13ce09e4890298e&ti%7Cgb=android12-5.4&ti%7Cgu=https%3A%2F%2Fandroid.googlesource.com%2Fkernel%2Fcommon&ti%7Ct=android

image

WilsonNet:
maybe when we have an empty log_exceprt we should show the fallback text

dashboard/src/utils/utils.ts
```diff
@@ -188,3 +188,14 @@ export const mapFiltersKeysToBackendCompatible = (
...
 };
+
+export const isStringRecord = (
+  obj: unknown,
+): obj is Record<string, unknown> => {
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    !Array.isArray(obj) &&
+    Object.keys(obj).every(key => typeof key === 'string')
```
WilsonNet, line 199: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys

Object keys already return string keys, I don't think objects can have any other type of key, maps can though

dashboard/src/components/Sheet/LogOrJsonSheetContent.tsx
```diff
@@ -22,7 +22,7 @@ interface ILogSheet {
 }
 
 export interface IJsonContent {
-  src?: JSON;
+  src?: object;
```
WilsonNet, line 25: `Record<string, unknown>`
functions are object in typescript

dashboard/src/utils/utils.ts
murilx: This function might still make sense though?

```
➜  ~ node
Welcome to Node.js v22.3.0.
Type ".help" for more information.
> map1 = new Map()
Map(0) {}
> map1.set(1, 'a')
Map(1) { 1 => 'a' }
> map1.get(1)
'a'
> typeof map1
'object'
> typeof Array.from(map1.keys())[0]
'number'
```
But maybe implement it differently because if `obj` is a map and not a record than `Object.keys(obj).every(key => typeof key === 'string')` would still return true

murilx: Maybe
```js
    typeof obj === 'object' &&
    obj !== null &&
    !(obj instanceof Map) &&
    !(obj instanceof Set) &&
    !Array.isArray(obj)
```
(would need to double check if there are any other objects that are not records in js)

WilsonNet: I don't think we will ever see a map being used in the frontend

dashboard/src/utils/misc.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const JsonViewIcon = (): JSX.Element => (
+  <Tooltip>
+    <TooltipTrigger>
+      <AiOutlineFileSearch className="text-lg text-blue" />
+    </TooltipTrigger>
+    <TooltipContent>
+      <FormattedMessage id="global.viewJson" />
+    </TooltipContent>
+  </Tooltip>
+);
```
MarceloRobert, line 32: Move this to the icons folder

WilsonNet (approved):
nice

dashboard/src/utils/misc.tsx
```diff
@@ -0,0 +1,91 @@
...
+    .with({ log_excerpt: '' }, _ => false)
+    .with({ log_excerpt: P.string }, field => {
+      returnObj.children = <CodeBlock code={field.log_excerpt} />;
+      return true;
+    })
+    .otherwise(_ => {
+      returnObj.linkText = !shouldTruncate(stringField) ? (
+        stringField
+      ) : (
+        <TruncatedValueTooltip value={stringField} isUrl={isUrlResult} />
```
MarceloRobert, line 85:
```suggestion
        <TruncatedValueTooltip value={stringField} isUrl={isUrlResult || isObjectField} />
```
this parameter will tell the field to use a cursor-pointer class, you can change the name of the parameter if you want

MarceloRobert (commented):
IMO it would be useful if there was an info icon to the right of "Logspec Data" saying that it is the same logspec that appears under the misc section, just so it is clearer to the user that sees the logspec in two parts of the same page

dashboard/src/utils/misc.tsx
murilx: This is good but in the case that a json that is not truncated the pointer won't show. Maybe it's better if I add this to LinkWithIcon component as well

dashboard/src/locales/messages/index.ts
```diff
@@ -206,6 +208,8 @@ export const messages = {
       'This log url is not supported in the log viewer yet, but you can still download the log in the link above',
     'logSheet.noLogFound': 'No logs available',
     'logSheet.title': 'Logs Viewer',
+    'logspec.info':
+      "This is the same logspec data that's in the misc data section",
```
MarceloRobert, line 212:
```suggestion
      "This is the same logspec data that is in the misc data section",
```

MarceloRobert (approved):
LGTM

PR #842, opened 2025-01-27
feat(hardware-details): add compatibles filter

### LINKED ISSUE #712: feat(HardwareDetails): add filters to compatibles card

murilx:
- Add filter for compatibles card in hardware details
- Remove the hardware id of the page from compatibles card

Closes #712

# How to test
From my experience, most of the hardwares have tests that share the same environment_compatible array. For that reason, is kind of hard to find a hardware detail where the filter has an effect. This is one of them: http://localhost:5173/hardware/amlogic%2Cg12b?et=1737981000&p=bt&st=1737549000. In this example, we have two different sets of tests, one where the environment_compatible includes `libretech,aml-a311d-cc` and the other where environment_compatible includes `khadas,vim3`. Testing in this page would go:

- Filter by `libretech,aml-a311d-cc`. Save the status values for later and verify that all the boots/tests have `libretech/aml-a311d-cc` in the hardware field. For builds you could open each tree named in the build details page and check if `libretech,aml-a311d-cc` is one of the hardwares used.
- Filter by `khadas,vim3`. In this case, since there aren't hardwares that include both `khadas,vim3` and `libretech,aml-a311d-cc` the status values should be equal to `unfiltered_status_values - libretech_status_values`. You can also perform the same tests in last path, opening tests details to verify `khadas,vim3` appears in hardware (and that `libretech.aml-a311d-cc` does not appear since the values correspond to them being exclusive values). Alternatively you could write a SQL query that gets all rows in a time period that contains `amlogic,g12b`, `libretech,aml-a311d-cc` and `khadas,vim3` and the result should be empty
- Filter by `amlogic,a311d` and check that it's the same value as the unfiltered page since `amlogc.a311d` is in all tests.

### DIFF
--- a/backend/kernelCI_app/helpers/hardwareDetails.py
+++ b/backend/kernelCI_app/helpers/hardwareDetails.py
@@ -720,6 +720,7 @@ def decide_if_is_full_record_filtered_out(
         return True
 
     is_record_filtered_out_result = instance.filters.is_record_filtered_out(
+        hardwares=record["environment_compatible"],
         architecture=record["build__architecture"],
         compiler=record["build__compiler"],
         config_name=record["build__config_name"],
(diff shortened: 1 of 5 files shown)


MarceloRobert (approved):
It was more straightforward than I thought, good job! Worked well on my tests

WilsonNet (approved):
working nice

PR #849, opened 2025-01-28
Fix: hardware commit selector

### LINKED ISSUE #783: Bug(Hardware details): when switching commit history in select menu, it goes back right after it changes
If you got to any hardware details, and change the value of a commit.

It change the results, but short after go back to the original state

murilx:
Added a new state `treeIndexesLength` that keep track of the size of number of trees there are in the header table. This was necessary because the backend expects a object that maps each tree index to its respective selected commit, but it couldn't correctly get all the tree indexes from the query parameter since it defaults to an empty array when all trees are selected. This state is then used when the query param is an empty array to create this object
    
Closes #783 

## How to test
This hardware is a good one for testing: [google,kingoftown](http://localhost:5173/hardware/google%2Ckingoftown?et=1738087200&st=1737655200), specially the `mainline/master` branch that has a lot of commits and different values between them. 
- Try to change the commits and see if the status in the header table change accordingly, also check for the data in the tabs. 
- Try to select or deselect trees from the header table and check if the data from the tabs are updated correctly. Also try to do it with changed commits in other trees to see if they don't update alone when selecting or deselecting trees.

### DIFF
--- a/dashboard/src/api/hardwareDetails.ts
+++ b/dashboard/src/api/hardwareDetails.ts
@@ -18,18 +18,29 @@ import type {
 } from '@/types/general';
 import { getTargetFilter } from '@/types/general';
 
+import { isEmptyObject } from '@/utils/utils';
+
 import { RequestData } from './commonRequest';
 
 const TREE_SELECT_HEAD_VALUE = 'head';
 
 const mapIndexesToSelectedTrees = (
   selectedIndexes: number[],
+  treeIndexesLength?: number,
   treeCommits: TTreeCommits = {},
 ): Record<string, string> => {
   const selectedTrees: Record<string, string> = {};
 
-  selectedIndexes.forEach(idx => {
-    const key = idx.toString();
+  if (selectedIndexes.length === 0 && isEmptyObject(treeCommits))
+    return selectedTrees;
+
+  const selectedArray =
+    treeIndexesLength && selectedIndexes.length === 0
+      ? Array.from({ length: treeIndexesLength }, (_, i) => i)
+      : selectedIndexes;
+
+  selectedArray.forEach(i => {
+    const key = i.toString();
     const value = treeCommits[key] || TREE_SELECT_HEAD_VALUE;
     selectedTrees[key] = value;
   });
@@ -114,6 +125,7 @@ export type UseHardwareDetailsWithoutVariant = {
   filter: TFilter;
   selectedIndexes: number[];
   treeCommits: TTreeCommits;
+  treeIndexesLength?: number;
   enabled?: boolean;
 };
 
@@ -129,6 +141,7 @@ export const useHardwareDetails = <T extends HardwareDetailsVariants>({
   filter,
   selectedIndexes,
   treeCommits,
+  treeIndexesLength,
   enabled = true,
   variant,
 }: UseHardwareDetailsParameters<T>): UseQueryResult<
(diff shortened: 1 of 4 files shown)


dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -142,7 +142,11 @@ const CommitSelector = ({
 
   return (
     <div className="flex items-center gap-4">
-      <Select onValueChange={navigateToThePast} disabled={isMainPageLoading}>
+      <Select
+        onValueChange={navigateToThePast}
+        value={treeCommits[treeIndex] ?? headCommitHash}
```
Francisco2002, line 152: nit: use `commitHash` defined in line 140

dashboard/src/api/hardwareDetails.ts
```diff
@@ -24,12 +24,19 @@ const TREE_SELECT_HEAD_VALUE = 'head';
...
 ): Record<string, string> => {
   const selectedTrees: Record<string, string> = {};
 
-  selectedIndexes.forEach(idx => {
-    const key = idx.toString();
+  const selectedArray = treeIndexesLength
+    ? selectedIndexes.length === 0
+      ? Array.from({ length: treeIndexesLength }, (_, i) => i)
+      : selectedIndexes
+    : selectedIndexes;
```
Francisco2002, line 36:
```suggestion
  const selectedArray = treeIndexesLength && selectedIndexes.length === 0
      ? Array.from({ length: treeIndexesLength }, (_, i) => i)
      : selectedIndexes;
```

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
```diff
@@ -137,6 +137,7 @@ function HardwareDetails(): JSX.Element {
     });
   }, [navigate]);
 
+  const [treeIndexesLength, setTreeIndexesLength] = useState(0);
```
Francisco2002, line 140: This props don't need to update the component when it changes, because he is a dependency of `useQuery`. So, you could use `useRef` here and reduce the re-renders.

MarceloRobert:
IMO you shouldn't send the selected tree list to the backend when all of the trees are selected and all of them are in the head commit hash, meaning that you should only send it if any tree was unselected _or_ any of them had the commit hash changed

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
murilx: Might still need to use `useState` in this case since `treeIndexesLength` is being read by `useHardwareDetailsLazyLoaded` and AFAIK this happens during the component rendering and refs shouldn't be read during rendering.

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -366,6 +386,8 @@ export function HardwareHeader({
     );
   }, [selectedIndexes, treeItems.length]);
 
+  const columns = getColumns(setTreeIndexesLength);
```
MarceloRobert, line 389: `useMemo` here

MarceloRobert (commented):
Seems to be working on my tests now

dashboard/src/utils/utils.ts
```diff
@@ -199,3 +199,10 @@ export const isStringRecord = (
     !Array.isArray(obj)
   );
 };
+
+export const isEmptyObject = (obj: Record<string, unknown>): boolean => {
+  for (const prop in obj) {
+    if (Object.prototype.hasOwnProperty.call(obj, prop)) return false;
+  }
+  return true;
+};
```
Francisco2002, line 208: you can do `Object.keys(obj).length == 0` to test if a object is empty

Francisco2002 (approved):
approved with a comment

dashboard/src/utils/utils.ts
murilx: Although in the worst case scenario both of these solutions are `O(n)`, in the average scenario this function would work much faster. In the situation it's being used I'd say it would finish in the first iteration most of the time. Now, the truth is that `treeCommits` is a very little object, in the worst case it would have something like a hundred elements so it's probably not something to care about now

PR #850, opened 2025-01-28
feat/cache unsync detectable

### LINKED ISSUE #833: front cache invalidation

WilsonNet:
# Description 

Make frontend detects when there is a inconsistency between listing and details and bust the frontend cache

### `mitmproxy` installation
1.  Install mitmproxy (you can install through `pip`)  (If it breaks use a virtual environment)
2. You can proxy the entire system but I recommend proxy only firefox, set the proxy only for http://localhost:8080 (HTTP without the S). If you proxy https it will pollute everything with your regular web browsing
3. Go to about:config in Firefox and  set network.proxy.allow_hijacking_localhost to true
4. run mitmproxy web with `mitmweb`
5. Install the Certificates for mitmproxy by going to this link http://mitm.it/

## Usage tutorial

1. with mitmweb open and traffic logging filter filter the results to show requests with `sumary` only
2. Go to any details page and add the backend request to the intercept
3. Go back to tree listing (if you refresh it will lose the internal router state from tree listing)
4. `sessionStorage.clear()`
5. Enter in the same details page that the backend request for summary is intercepted
6. It will intercept the request, clik play a single time, now you will have the response intercepted, edit the response (use an online json editor to make things easier). It will only bust cache if something that tree listing is aware which are the `tests/builds/boots` status number
7. Click play again
8. See the edited response being reflected in the app and see how another request has been triggered
9. Click play two times to get the right data from the database
 
### Video tutorial on how to test

https://github.com/user-attachments/assets/dea79d45-6da6-42c2-a788-7cec67e53087

Closes #833


dashboard/src/utils/records.ts
```diff
@@ -0,0 +1,24 @@
+const isRecord = (obj: unknown): obj is Record<string, unknown> => {
+  return typeof obj === 'object' && !Array.isArray(obj);
+};
```
murilx, line 3: You might want to add
```suggestion
const isRecord = (obj: unknown): obj is Record<string, unknown> => {
  return typeof obj === 'object' && obj !== null && !Array.isArray(obj);
};
```
Unless you want null to be considered a record. Also, we already have the function `isStringRecord` in utils.ts. Please remove it from there and substitute the occurrences of `isStringRecord` to `isRecord`

dashboard/src/utils/records.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  for (const key in obj1) {
+    if (isRecord(obj1[key]) && isRecord(obj2[key])) {
+      if (!deepCompare(obj1[key], obj2[key])) {
+        return false;
+      }
+    } else if (obj1[key] !== obj2[key]) {
+      return false;
+    }
```
murilx, line 20: This does not consider that the objects might have an array as a property value

dashboard/src/utils/records.ts
```diff
@@ -0,0 +1,24 @@
+const isRecord = (obj: unknown): obj is Record<string, unknown> => {
+  return typeof obj === 'object' && !Array.isArray(obj);
+};
+
+export const deepCompare = (
```
murilx, line 5: nit: Could add some unit tests for this function if you feel like it

dashboard/src/main.tsx
```diff
@@ -40,6 +44,16 @@ declare module '@tanstack/react-router' {
...
     from?: RedirectFrom;
+    treeStatusCount?: {
+      buildsStatus?: BuildStatus;
+      bootStatus?: RequiredStatusCount;
+      testStatus?: RequiredStatusCount;
+    };
+    hardwareStatusCount?: {
+      builds: BuildStatus;
+      boot: RequiredStatusCount;
+      test: RequiredStatusCount;
```
MarceloRobert, line 55: You don't need them to be different names here, keep both as `buildsStatus` or just `builds` and the same to boot and test; Also check the pluralization between builds (plural) and boot/test (singular)

WilsonNet: yeah, this is leftover form the CI error, good catch

murilx (approved):
LGTM. Sometimes it does an additional request but it does not seem to be a big problem now or something that we can easily solve at the moment

Francisco2002 (approved):
Nice job!!!!

PR #859, opened 2025-01-29
Feat: add platforms to hardware column

### LINKED ISSUE #854: Show misc.platform in test table if hardware is undefined
There is a column in the tests table which shows the hardware tested, but if the hardware is undefined it should show the platform information

murilx:
- Added hardware column to boots table
- Added `environment_misc.platform` as a fallback to hardware when `environment_compatible` is full

Closes #854 

**Important:** In the current form boots table is showing the hardware column in the hardware details page but tests table is not. Should hardware column be added to the tests table, should it be removed from the boots table or should I keep it like it is already?

**Update:** It was decided that a hardware column in the boots table in hardware details wouldn't be of so much value. In tree details, however, it was added and in the tests table it was kept the same, since there they provide more value to the user.


dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -99,13 +99,18 @@ export function TestsTable({
           };
         }
         groups[group].total_tests++;
+        const hardware = e.environment_compatible
+          ? e.environment_compatible
+          : e.misc?.platform
+            ? [e.misc?.platform]
+            : undefined;
```
Francisco2002, line 106: Create a function to this

dashboard/src/utils/table.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const buildHardwareArray = (
+  environment_compatible?: TestHistory['environment_compatible'],
+  misc?: TestHistory['misc'],
+): string[] | undefined => {
+  return environment_compatible
+    ? environment_compatible
+    : misc?.platform
+      ? [misc?.platform]
+      : undefined;
```
WilsonNet, line 11: this is pretty much the same as `misc?.platform`

dashboard/src/components/BootsTable/BootsTable.tsx
```diff
@@ -42,66 +38,17 @@ import { usePaginationState } from '@/hooks/usePaginationState';
...
-  },
-  {
-    id: DETAILS_COLUMN_ID,
-    header: (): JSX.Element => <MoreDetailsTableHeader />,
-    cell: (): JSX.Element => <MoreDetailsIcon />,
-  },
-];
+import { buildHardwareArray } from '@/utils/table';
+
+import { defaultColumns } from './DefaultBootsColumns';
```
WilsonNet, line 45: there is pretty much no gain moving to another file it will be only used in a single place

dashboard/src/utils/table.ts
murilx: `misc?.platform` is type `string | undefined` and I need the result to be `string[] | undefined`. Doing only `[misc?.platform]`, wouldn't work as well because that would be type `(string | undefined)[]`

dashboard/src/components/BootsTable/BootsTable.tsx
murilx: I was following the same structure that was done to `TestsTable` but it makes sense, I'll change it back

WilsonNet: You can leave this way but I prefer on the same file since it doesn't pollute the imports

Francisco2002 (approved):
LGTM

WilsonNet (approved):
looks good

PR #866, opened 2025-02-03
Update issues list tooltip

padovan:


MarceloRobert (approved):
Thanks for the feedback, LGTM

PR #876, opened 2025-02-03
feat/platform support for hardware details

### LINKED ISSUE #865: HardwareDetails: get details from platforms
Add extra column, platform. Name is still the compatible.
When there is an entry without Name (compatible), show Platform info also in Name.

Adapt hardwareDetails to also work with platforms instead of just compatibles.

WilsonNet:
## Description

Makes Hardware comaptible with the platform field in the JSON.

Closes #865 

## How to test

Use known platforms that are not compatibles:
http://localhost:5173/hardware/kubernetes?et=1738672200&st=1738240200

http://localhost:5173/hardware/sc7180-trogdor-lazor-limozeen?et=1738672200&st=1738240200

http://localhost:5173/hardware/acer-chromebox-cxi4-puff?et=1738672200&st=1738240200

### DIFF
--- a/backend/kernelCI_app/typeModels/hardwareDetails.py
+++ b/backend/kernelCI_app/typeModels/hardwareDetails.py
@@ -10,6 +10,7 @@ from kernelCI_app.typeModels.commonDetails import (
 )
 
 from kernelCI_app.typeModels.databases import (
+    Issue__Id,
     StatusValues,
     Checkout__TreeName,
     Checkout__GitRepositoryBranch,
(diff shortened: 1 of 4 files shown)


backend/kernelCI_app/typeModels/hardwareDetails.py
```diff
@@ -94,7 +96,7 @@ class HardwareTestHistoryItem(TestHistoryItem):
 class HardwareBuildHistoryItem(BuildHistoryItem):
     tree_name: Optional[str]
     issue_id: Optional[str]
-    issue_version: Optional[int]
+    issue_version: Optional[Issue__Version]
```
MarceloRobert, line 99: nit: I don't see a need to change this here, unless you change other fields as well, but it will work nonetheless

MarceloRobert (approved):
Seems to be working well

backend/kernelCI_app/typeModels/hardwareDetails.py
murilx: I have the opinion that in this case specifically (moving to the standardized types) any gradual change is already good. But I agree that if this change was not motivated by any other thing other than simply changing the type, I'd like to see this happen at least in every field of this model

murilx (approved):
django `models.Q` surprised me because I thought this would be harder to implement. Worked well on my test!

PR #877, opened 2025-02-03
feat: change hardware priority in tests table

### LINKED ISSUE #864: Add Platform>Compatibles to all pages (Issue Details, Tree Details, Hardware Details etc)
If there is Platform and Compatibles, hover shows Compatibles. Platform has higher priority (show it when available) than Compatibles.

murilx:
- Show platform first and enviroment_compatible in the hover if it exists. Show only environment_compatible if platform is undefined
- [x] Add platforms to hardware column in the Issue Details tests table
- [x] Add platforms to hardware column in the Build Details tests table

## How to test
- Go to a tree details page and verify that the hardware priority is working correctly (to ensure that the tooltip is working you can filter for a hardware and check if the tooltip shows)
- Do the same in a build details page. In this case you can open one of the tests and check if the platform is the one showing in the table and if the tooltip is correctly showing the environment_compatible (field "hardware" in test details) ou showing nothing if hardware is "Unknown"
- Do the same in an issue details page. To quickly find an issue details page with tests you can go to a boots tab with issues and open one of the issues there.

Closes #864

### DIFF
--- a/backend/kernelCI_app/typeModels/buildDetails.py
+++ b/backend/kernelCI_app/typeModels/buildDetails.py
@@ -21,6 +21,7 @@ from kernelCI_app.typeModels.databases import (
     Test__Status,
     Test__StartTime,
     Test__EnvironmentCompatible,
+    Test__EnvironmentMisc,
 )
 
 
@@ -46,6 +47,7 @@ class BuildTestItem(BaseModel):
     path: Test__Path
     start_time: Test__StartTime
     environment_compatible: Test__EnvironmentCompatible
+    environment_misc: Test__EnvironmentMisc
 
 
 class BuildTestsResponse(RootModel):
--- a/backend/kernelCI_app/typeModels/issueDetails.py
+++ b/backend/kernelCI_app/typeModels/issueDetails.py
@@ -28,6 +28,7 @@ from kernelCI_app.typeModels.databases import (
     Test__Path,
     Test__StartTime,
     Test__EnvironmentCompatible,
+    Test__EnvironmentMisc,
     Timestamp,
     Checkout__TreeName,
     Checkout__GitRepositoryBranch,
@@ -62,6 +63,7 @@ class IssueTestItem(BaseModel):
     environment_compatible: Test__EnvironmentCompatible = Field(
         alias="test__environment_compatible"
     )
+    environment_misc: Test__EnvironmentMisc = Field(alias="test__environment_misc")
     tree_name: Checkout__TreeName = Field(alias="test__build__checkout__tree_name")
     git_repository_branch: Checkout__GitRepositoryBranch = Field(
         alias="test__build__checkout__git_repository_branch"
--- a/dashboard/src/components/Table/TooltipHardware.tsx
+++ b/dashboard/src/components/Table/TooltipHardware.tsx
@@ -4,27 +4,33 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
 
 import { sanitizeTableValue } from '@/components/Table/tableUtils';
 
+const tooltipStartIdx = 1;
+
 const TooltipHardware = ({
   hardwares,
 }: {
   hardwares: string[] | undefined;
 }): JSX.Element => {
+  const shouldHaveTooltip = hardwares && hardwares.length > 1;
+
   const hardwaresTooltip = useMemo(() => {
-    return hardwares
-      ? hardwares.map(hardware => (
+    return shouldHaveTooltip
+      ? hardwares.slice(tooltipStartIdx).map(hardware => (
           <div key={hardware} className="text-center">
             <span>{hardware}</span>
             <br />
           </div>
         ))
       : '-';
-  }, [hardwares]);
+  }, [hardwares, shouldHaveTooltip]);
+
+  const hardwareTrigger = sanitizeTableValue(hardwares?.[0], false);
+
+  if (!shouldHaveTooltip) return <span>{hardwareTrigger}</span>;
 
   return (
     <Tooltip>
-      <TooltipTrigger>
-        {sanitizeTableValue(hardwares?.[0], false)}
-      </TooltipTrigger>
+      <TooltipTrigger>{hardwareTrigger}</TooltipTrigger>
       <TooltipContent>{hardwaresTooltip}</TooltipContent>
     </Tooltip>
   );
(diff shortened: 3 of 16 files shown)


backend/kernelCI_app/typeModels/buildDetails.py
```diff
@@ -46,6 +47,7 @@ class BuildTestItem(BaseModel):
     path: Test__Path
     start_time: Test__StartTime
     environment_compatible: Test__EnvironmentCompatible
+    misc: Test__EnvironmentMisc = Field(alias="environment_misc")
```
WilsonNet, line 50: I'm not so confident on this alias since there is a `misc` column in the database, we may be requested in the future to use that

backend/kernelCI_app/typeModels/issueDetails.py
```diff
@@ -62,6 +63,7 @@ class IssueTestItem(BaseModel):
     environment_compatible: Test__EnvironmentCompatible = Field(
         alias="test__environment_compatible"
     )
+    misc: Test__EnvironmentMisc = Field(alias="test__environment_misc")
```
WilsonNet, line 66: ditto about the alias

dashboard/src/components/Table/TooltipHardware.tsx
```diff
@@ -9,25 +9,31 @@ const TooltipHardware = ({
 }: {
   hardwares: string[] | undefined;
 }): JSX.Element => {
+  const shouldHaveTooltip = hardwares && hardwares.length > 1;
+
   const hardwaresTooltip = useMemo(() => {
-    return hardwares
-      ? hardwares.map(hardware => (
+    return shouldHaveTooltip
+      ? hardwares.slice(1).map(hardware => (
```
WilsonNet, line 16: nit: magic number

dashboard/src/utils/table.ts
```diff
@@ -6,11 +6,9 @@ export const buildHardwareArray = (
...
   misc?: TestHistory['misc'],
 ): string[] | undefined => {
-  return environment_compatible
-    ? environment_compatible
-    : misc?.platform
-      ? [misc?.platform]
-      : undefined;
+  const miscArray: (string | undefined)[] = [misc?.platform];
+  const envArray = environment_compatible ?? [environment_compatible];
+  return miscArray?.concat(envArray).filter(i => i !== undefined && i !== null);
```
WilsonNet, line 11: I personally would do something like
```suggestion
  const miscArray: (string | undefined)[] = misc?.platform ?? [];
  const envArray = environment_compatible ?? [];
  return miscArray.concat(envArray);
```
now you don't need the filter

backend/kernelCI_app/typeModels/buildDetails.py
murilx: This field name was a kind of "old" implementation I believe I did (not really a wise idea to call it 'misc' in retrospect) so I had to change this in other places of the code as well. I did test to check if everything was still working (ran all of the requests scripts and also navigated on the application) but I don't exclude the idea that some new bug might be created from this change, so please test the application in general if possible as well.

I also initially changed this for test-related items only. There's a misc in build items that we need to verify if comes from "build__misc" database field and if that's the case I don't see a problem with keeping it as only misc

backend/kernelCI_app/typeModels/issueDetails.py
murilx: answered in the comment above

dashboard/src/utils/table.ts
```diff
@@ -4,13 +4,13 @@ import type { TestHistory } from '@/types/general';
...
-  return environment_compatible
-    ? environment_compatible
-    : misc?.platform
-      ? [misc?.platform]
-      : undefined;
+  const miscArray: string[] = environment_misc?.platform
+    ? [environment_misc.platform]
+    : [];
+  const envArray: string[] = environment_compatible ?? [];
+  return miscArray?.concat(envArray);
```
WilsonNet, line 13: you don't need the optional chaining here

dashboard/src/components/Table/TooltipHardware.tsx
```diff
@@ -9,25 +9,32 @@ const TooltipHardware = ({
 }: {
   hardwares: string[] | undefined;
 }): JSX.Element => {
+  const shouldHaveTooltip = hardwares && hardwares.length > 1;
+  const tooltipStartIdx = 1;
```
Francisco2002, line 13: Define this as a constant outside the component

dashboard/src/components/Table/TooltipHardware.tsx
```diff
@@ -9,25 +9,32 @@ const TooltipHardware = ({
 }: {
   hardwares: string[] | undefined;
 }): JSX.Element => {
+  const shouldHaveTooltip = hardwares && hardwares.length > 1;
```
Francisco2002, line 12:
```suggestion
  const shouldHaveTooltip = hardwares?.length > 1;
```

dashboard/src/components/Table/TooltipHardware.tsx
```diff
@@ -9,25 +9,32 @@ const TooltipHardware = ({
...
+  if (shouldHaveTooltip) {
+    return (
+      <Tooltip>
+        <TooltipTrigger>{hardwareTrigger}</TooltipTrigger>
+        <TooltipContent>{hardwaresTooltip}</TooltipContent>
+      </Tooltip>
+    );
+  } else {
+    return <span>{hardwareTrigger}</span>;
+  }
```
Francisco2002, line 37:
```suggestion
  if (!shouldHaveTooltip)
    return <span>{hardwareTrigger}</span>;
    
  return (
    <Tooltip>
      <TooltipTrigger>{hardwareTrigger}</TooltipTrigger>
      <TooltipContent>{hardwaresTooltip}</TooltipContent>
    </Tooltip>
  );
```

dashboard/src/components/Table/TooltipHardware.tsx
murilx: doing so raises the error `'hardwares.length' is possibly 'undefined'.ts(18048)` by typescript because I can't compare undefined with a number I imagine

WilsonNet (approved):
Works nice, I don't think adding to hardware details is part of this ticket, so I'm approving now

Francisco2002 (approved):
Nice job!!

PR #881, opened 2025-02-04
Feat: add size listing parameter

### LINKED ISSUE #860: Add GET parameter for size of the tree/hardware list
Feedback from a user:

```
One thing that would be convenient for me is if there were some way to
set the "Items per page" for my android homepage since the default of
10 is a little too small so I always have to change it to 20. Could we
add a setting that can be appended to the URL to set this?

Maybe an "items=" setting so that instead of my normal top-level page:

https://dashboard.kernelci.org/tree?ts=android

I'd append an "items=20":

https://dashboard.kernelci.org/tree?ts=android&items=20
```

Francisco2002:
Add `listingSize` parameter to tree and hardware listing. The alias to this param is `ls`.

### **How to test**

1. Go to tree listing or hardware listing
2. On the URL, append the `ls` parameter
    -  http://localhost:5173/tree?ls=13
    -  http://localhost:5173/tree?ts=android&ls=13
3. See that the page number is the next decade. If you give `ls=13`, the page size will be 20. The max page size is 50.

Close #860

### DIFF
--- a/dashboard/src/hooks/usePaginationState.tsx
+++ b/dashboard/src/hooks/usePaginationState.tsx
@@ -4,18 +4,39 @@ import type { Updater, PaginationState } from '@tanstack/react-table';
 
 import type { TableKeys } from '@/utils/constants/tables';
 
+import { ItemsPerPageValues } from '@/utils/constants/general';
+
 import { useFeatureFlag } from './useFeatureFlag';
 
 const DEFAULT_PAGINATION_STATE = { pageIndex: 0, pageSize: 10 } as const;
 
+const DEFAULT_LISTING_SIZE = 10;
+
+const findClosestGreaterNumber = (sortedList: number[], x: number): number => {
+  for (let i = 0; i < sortedList.length; i++) {
+    if (sortedList[i] >= x) return sortedList[i];
+  }
+
+  return sortedList.slice(-1)[0] ?? DEFAULT_LISTING_SIZE;
+};
+
 export const usePaginationState = (
   selectedTable: TableKeys,
+  defaultValue?: number,
 ): {
   pagination: PaginationState;
   paginationUpdater: (updater: Updater<PaginationState>) => void;
 } => {
   const { showDev } = useFeatureFlag();
   const [pagination, setPagination] = useState<PaginationState>(() => {
+    if (defaultValue) {
+      const pageSize = findClosestGreaterNumber(
+        ItemsPerPageValues,
+        defaultValue,
+      );
+      return { pageIndex: DEFAULT_PAGINATION_STATE.pageIndex, pageSize };
+    }
+
     const storageData = window.localStorage.getItem(selectedTable);
     if (storageData && showDev) {
       try {
--- a/dashboard/src/routes/(alternatives)/t/route.tsx
+++ b/dashboard/src/routes/(alternatives)/t/route.tsx
@@ -7,7 +7,7 @@ import {
 import { z } from 'zod';
 
 import { makeZIntervalInDays, type SearchSchema } from '@/types/general';
-import { DEFAULT_TIME_SEARCH } from '@/pages/treeConstants';
+import { DEFAULT_TIME_SEARCH } from '@/utils/constants/general';
 
 const defaultValues = {
   intervalInDays: DEFAULT_TIME_SEARCH,
--- a/dashboard/src/routes/(alternatives)/t/route.tsx
+++ b/dashboard/src/routes/(alternatives)/t/route.tsx
@@ -6,17 +6,26 @@ import {
 
 import { z } from 'zod';
 
-import { makeZIntervalInDays, type SearchSchema } from '@/types/general';
-import { DEFAULT_TIME_SEARCH } from '@/utils/constants/general';
+import {
+  makeZIntervalInDays,
+  zListingSize,
+  type SearchSchema,
+} from '@/types/general';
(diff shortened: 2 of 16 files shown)


dashboard/src/hooks/usePaginationState.tsx
```diff
@@ -8,14 +8,29 @@ import { useFeatureFlag } from './useFeatureFlag';
...
+  defaultValue?: number,
 ): {
   pagination: PaginationState;
   paginationUpdater: (updater: Updater<PaginationState>) => void;
 } => {
   const { showDev } = useFeatureFlag();
   const [pagination, setPagination] = useState<PaginationState>(() => {
+    if (defaultValue) {
+      const pageSize = definePageSize(defaultValue);
+      return { pageIndex: 0, pageSize };
```
murilx, line 31:
```suggestion
      return { pageIndex: DEFAULT_PAGINATION_STATE.pageIndex, pageSize };
```

murilx (commented):
Seems to be working fine, just some minor comments

dashboard/src/hooks/usePaginationState.tsx
```diff
@@ -8,14 +8,29 @@ import { useFeatureFlag } from './useFeatureFlag';
...
 const DEFAULT_PAGINATION_STATE = { pageIndex: 0, pageSize: 10 } as const;
 
+const DECADE = 10;
+const MAX_PAGE = 50;
+
+function definePageSize(value: number): number {
+  if (value <= DECADE / 2) return DECADE / 2;
+  if (value % DECADE === 0) return Math.min(value, MAX_PAGE);
+  return Math.min(Math.ceil(value / DECADE) * DECADE, MAX_PAGE);
+}
```
MarceloRobert, line 18: This function seems very detached from the array of possible tab sizes, maybe a function that finds the closest number from an arbitrary array would be better than specifying the possible values again as "5 and multiples of 10 up to `MAX_PAGE`", as it is virtually being done here. With a "find closest value" function, you could use the already defined array of possible values for the page size.

dashboard/src/routes/(alternatives)/t/route.tsx
```diff
@@ -7,16 +7,21 @@ import {
...
 const defaultValues = {
   intervalInDays: DEFAULT_TIME_SEARCH,
   treeSearch: '',
+  listingSize: DEFAULT_LISTING_ITEMS,
 };
 
 export const RootSearchSchema = z.object({
   intervalInDays: makeZIntervalInDays(DEFAULT_TIME_SEARCH),
   treeSearch: z.string().catch(''),
+  listingSize: z.optional(z.number().min(1).catch(DEFAULT_LISTING_ITEMS)),
```
MarceloRobert, line 24: extract this definition to a zod validator so you can share it between the routes, since they should be the same

MarceloRobert (commented):
There seems to be a problem when you try to assign a negative number to the `ls` parameter, the search parameter goes back to the default but the table doesn't

dashboard/src/types/general.ts
```diff
@@ -128,6 +128,9 @@ export type ArchCompilerStatus = {
   status: StatusCounts;
 };
 
+export const zOptional = z.optional;
+export const zListingSize = z.number().min(1);
```
murilx, line 132: I'd say you can add `.catch` and `.default` here and remove it from the routes

dashboard/src/hooks/usePaginationState.tsx
```diff
@@ -4,18 +4,36 @@ import type { Updater, PaginationState } from '@tanstack/react-table';
...
 
+const DECADE = 10;
+
+const findClosestNumber = (list: number[], x: number): number => {
+  for (let i = 0; i < list.length; i++) {
+    if (list[i] >= x) return list[i];
+  }
+
+  return list.slice(-1)[0] ?? DECADE;
+};
```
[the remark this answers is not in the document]
murilx: nit: would change this to this name or something like this `findClosestNumber` -> `findClosestGreaterNumber`, since it does not exactly find the closest number

murilx (commented):
Seems to be working fine, there's just some minor improvements to do

dashboard/src/hooks/usePaginationState.tsx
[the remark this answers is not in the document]
Francisco2002: Done!
If, in the future, this function is necessary in other context, I will sort the list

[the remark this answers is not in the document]
murilx: Fair point

dashboard/src/hooks/usePaginationState.tsx
```diff
@@ -4,18 +4,39 @@ import type { Updater, PaginationState } from '@tanstack/react-table';
 
 import type { TableKeys } from '@/utils/constants/tables';
 
+import { ItemsPerPageValues } from '@/utils/constants/general';
+
 import { useFeatureFlag } from './useFeatureFlag';
 
 const DEFAULT_PAGINATION_STATE = { pageIndex: 0, pageSize: 10 } as const;
 
+const DECADE = 10;
```
MarceloRobert, line 13: nit: You could use `DEFAULT_LISTING_SIZE` here

MarceloRobert (approved):
Worked well on my tests

murilx (approved):
Nice job!

PR #882, opened 2025-02-05
fix(tests-table): filter by individual table path

### LINKED ISSUE #875: Individual Tests Table path filter not working
For treeDetails and hardwareDetails the frontend filtering is not used since we can filter in the backend, but for pages such as buildDetails and issueDetails - where we don't have backend filtering - the path filter should consider the individual tests, not just the `path_group`

murilx:
Filter by the path of every individual test when `updatePathFilter` is undefined

Closes #875

## How to test
- Ensure that the behavior of the tests table in Tree Details and Hardware Details are the same. Since they use `updatePathFilter` there shouldn't be any differences from the version on staging. However, they are still influenced by this implementation since all of the status counting is done in the filteredData `useMemo`.
- Check if the tests tables in other pages (issue details and build details) can correctly filter by path now. In maestro, issue details generally have only one test group (boot) so it's recommended to test on build details or test in an issue details from redhat, for example, http://localhost:5173/issue/redhat%3Aissue_3137

### DIFF
--- a/dashboard/src/components/TestsTable/TestsTable.tsx
+++ b/dashboard/src/components/TestsTable/TestsTable.tsx
@@ -56,6 +56,44 @@ export interface ITestsTable {
   currentPathFilter?: string;
 }
 
+type TPathTestsStatus = Pick<
+  TPathTests,
+  | 'done_tests'
+  | 'error_tests'
+  | 'fail_tests'
+  | 'miss_tests'
+  | 'pass_tests'
+  | 'skip_tests'
+  | 'null_tests'
+  | 'total_tests'
+>;
+
+const countStatus = (group: TPathTestsStatus, status?: string): void => {
+  group.total_tests++;
+  switch (status?.toUpperCase()) {
+    case StatusTable.DONE:
+      group.done_tests++;
+      break;
+    case StatusTable.ERROR:
+      group.error_tests++;
+      break;
+    case StatusTable.FAIL:
+      group.fail_tests++;
+      break;
+    case StatusTable.MISS:
+      group.miss_tests++;
+      break;
+    case StatusTable.PASS:
+      group.pass_tests++;
+      break;
+    case StatusTable.SKIP:
+      group.skip_tests++;
+      break;
+    default:
+      group.null_tests++;
+  }
+};
+
 // TODO: would be useful if the navigation happened within the table, so the parent component would only be required to pass the navigation url instead of the whole function for the update and the currentPath diffFilter (boots/tests Table)
 export function TestsTable({
   tableKey,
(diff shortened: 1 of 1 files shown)


WilsonNet:
this is a better test subject than the one in the link, since there are tests paths that are different from the group path
http://localhost:5173/test/maestro%3A67a256ad661a7bc8748bcd53?p=t&ti%7Cc=android11-5.4.289_r00-13-gc1ccccc765f0e&ti%7Cch=c1ccccc765f0e088c26aaab5b5f283d09a248185&ti%7Cgb=android11-5.4&ti%7Cgu=https%3A%2F%2Fandroid.googlesource.com%2Fkernel%2Fcommon&ti%7Ct=android

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -107,39 +145,69 @@ export function TestsTable({
...
+        miss_tests: 0,
+        pass_tests: 0,
+        null_tests: 0,
+        skip_tests: 0,
+        error_tests: 0,
+        total_tests: 0,
+      };
+      const individualTest = test.individual_tests.filter(t => {
+        let dataIncludesPath = true;
+        if (isValidPath) dataIncludesPath = t.path.includes(path);
```
WilsonNet, line 183: nit:
```suggestion
        if (isValidPath) {
         dataIncludesPath = t.path.includes(path);
         }
```
we should probably add this to eslint

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -107,39 +145,69 @@ export function TestsTable({
...
+        if (dataIncludesPath) {
+          statusCounter(localGroup, t.status);
+          statusCounter(globalGroup, t.status);
```
WilsonNet, line 186: I think this is missplaced, the filter is affecting these counts just like the global group

image

murilx: I believe this is the expected behavior based on what we currently see in tree and hardware details. `globalGroup` would be the sum of all `localGroup`. If all tests are filtered out, then `globalGroup` will correctly be 0-0-0-0

WilsonNet (approved):
it is working on my tests

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -56,6 +56,44 @@ export interface ITestsTable {
...
+  | 'error_tests'
+  | 'fail_tests'
+  | 'miss_tests'
+  | 'pass_tests'
+  | 'skip_tests'
+  | 'null_tests'
+  | 'total_tests'
+>;
+
+const statusCounter = (group: TPathTestsStatus, status?: string): void => {
```
MarceloRobert, line 71: since this is a function that is not a component, name it a verb to describe what it does; for example "`countStatus`" or "`addStatusToCounter`"

MarceloRobert (approved):
Working well

PR #883, opened 2025-02-05
feat(hardware-listing): add non compatibles platforms

### LINKED ISSUE #867: Add platforms as option for hardware listing
Similar to #865 

Add an extra column, platform. The name will still the compatible.
When there is an entry without name (without compatibles), show platform info also in name.
Make possible to filter by platform name in the hardware search.

WilsonNet:
# Description

- List hardware that is not an environment_compatible but has platform
- Add platform column to the hardware listing 
- Add swagger documentation

  Closes #867 #887

## How to test
Go to the hardware listing page and try to enter in a hardware details page that has the same repeat the process.

### DIFF
--- a/backend/kernelCI_app/typeModels/hardwareListing.py
+++ b/backend/kernelCI_app/typeModels/hardwareListing.py
@@ -1,0 +1,38 @@
+from datetime import datetime
+from pydantic import BaseModel
+from typing import List, TypedDict, Union, Set
+
+from kernelCI_app.constants.general import DEFAULT_ORIGIN
+
+
+class BuildStatusCountDict(TypedDict):
+    valid: int
+    invalid: int
+    null: int
+
+
+class HardwareItem(TypedDict):
+    hardware_name: str
+    platform: Union[str, Set[str]]
+    test_status_summary: dict[str, int]
+    boot_status_summary: dict[str, int]
+    build_status_summary: BuildStatusCountDict
+
+
+class HardwareResponse(BaseModel):
+    hardware: List[HardwareItem]
+
+
+# Since OpenAPI does not support timestamp as datetime we add an extra model just for
+# documentation purposes. This model is not used in the code.
+# TODO Remove timestamp from the api and this model
+class HardwareQueryParamsDocumentationOnly(BaseModel):
+    origin: str = DEFAULT_ORIGIN
+    startTimestampInSeconds: str
+    endTimeStampInSeconds: str
+
+
+class HardwareQueryParams(BaseModel):
+    origin: str = DEFAULT_ORIGIN
+    start_date: datetime
+    end_date: datetime
--- a/backend/kernelCI_app/views/hardwareView.py
+++ b/backend/kernelCI_app/views/hardwareView.py
@@ -1,37 +1,58 @@
 from collections import defaultdict
 from datetime import datetime
+from typing import Dict, List, Set
+from django.db import models
 from django.db.models import Subquery
-from django.http import JsonResponse
-from django.views import View
 from http import HTTPStatus
-from kernelCI_app.constants.general import DEFAULT_ORIGIN
+
+from drf_spectacular.utils import extend_schema
+from pydantic import ValidationError
+from rest_framework.request import Request
+from rest_framework.response import Response
+from rest_framework.views import APIView
 from kernelCI_app.helpers.build import build_status_map
-from kernelCI_app.helpers.date import (
-    parse_start_and_end_timestamps_in_seconds_to_datetime,
-)
 from kernelCI_app.helpers.errorHandling import (
-    ExceptionWithJsonResponse,
-    create_error_response,
+    create_api_error_response,
 )
+from kernelCI_app.helpers.filters import UNKNOWN_STRING
+from kernelCI_app.helpers.logger import log_message
+from kernelCI_app.helpers.misc import env_misc_value_or_default, handle_environment_misc
 from kernelCI_app.helpers.trees import get_tree_heads
 from kernelCI_app.models import Tests
+from kernelCI_app.typeModels.hardwareListing import (
+    HardwareItem,
+    HardwareQueryParams,
+    HardwareQueryParamsDocumentationOnly,
+    HardwareResponse,
+)
+from kernelCI_app.utils import is_boot
 
 
-class HardwareView(View):
-    def _getResults(self, start_date: datetime, end_date: datetime, origin: str):
-        processedBuilds = set()
-        processedTests = set()
+class HardwareView(APIView):
+    def _make_hardware_item(self, *, compatible: str, platform: str) -> HardwareItem:
+        return {
+            "hardware_name": compatible,
+            "platform": platform,
+            "test_status_summary": defaultdict(int),
+            "boot_status_summary": defaultdict(int),
+            "build_status_summary": {"valid": 0, "invalid": 0, "null": 0},
+        }
 
+    def _get_tests_from_database(
+        self, start_date: datetime, end_date: datetime, origin: str
+    ):
         checkouts_subquery = get_tree_heads(origin, start_date, end_date)
 
-        tests = Tests.objects.filter(
-            environment_compatible__isnull=False,
+        return Tests.objects.filter(
+            models.Q(environment_compatible__isnull=False)
+            | models.Q(environment_misc__platform__isnull=False),
             start_time__gte=start_date,
             start_time__lte=end_date,
             build__checkout__origin=origin,
             build__checkout__git_commit_hash__in=Subquery(checkouts_subquery),
         ).values(
             "environment_compatible",
+            "environment_misc",
             "status",
             "build__valid",
             "path",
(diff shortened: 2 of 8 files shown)


backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,121 @@
...
+            "build_status_summary": {"valid": 0, "invalid": 0, "null": 0},
+        }
+
+    def _get_tests_from_database(
+        self, start_date: datetime, end_date: datetime, origin: str
+    ):
         checkouts_subquery = get_tree_heads(origin, start_date, end_date)
 
-        tests = Tests.objects.filter(
-            environment_compatible__isnull=False,
```
MarceloRobert, line 28: in this case I think that the right condition is "if the environment compatible is not null _or_ if the environment misc is not null". By just removing this condition you allow getting results where both of them are null

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,121 @@
...
+
+        if self.hardware.get(current_hardware) is None:
+            self.hardware[current_hardware] = self._make_hardware_item(
+                compatible=current_hardware, platform=current_platform
+            )
+
+        status_count = "test_status_summary"
+
+        if test["path"].startswith("boot.") or test["path"] == "boot":
+            status_count = "boot_status_summary"
```
MarceloRobert, line 84: use the `is_boot` method from the utils folder

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,121 @@
...
+
+        build_key = test["build__id"] + current_hardware
+        if build_key in self.processed_builds:
+            return
+
+        self.processed_builds.add(build_key)
+
+        build_status = build_status_map.get(test["build__valid"])
+
+        if build_status is not None:
```
MarceloRobert, line 99: can `build_status` be None if you get its value from `build_status_map`?

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,121 @@
...
+                handle_environment_misc(test["environment_misc"])
+            )
+            platform = environment_misc["platform"]
+            environment_compatible = test["environment_compatible"]
+
+            if environment_compatible is None and platform is UNKNOWN_STRING:
+                continue
+
+            if environment_compatible is None:
+                print("platform: ", platform)
```
MarceloRobert, line 120:
```suggestion

```

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,121 @@
...
+    invalid: int
+    null: int
+
+
+class HardwareItem(TypedDict):
+    hardware_name: str
+    platform: str
+    test_status_summary: dict[str, int]
+    boot_status_summary: dict[str, int]
+    build_status_summary: BuildStatusCountDict
```
MarceloRobert, line 34: Is there a reason to not make them pydantic models and put them in the typedModels folder?

MarceloRobert (commented):
Seems to be working well but I made some comments in the code

backend/kernelCI_app/views/hardwareView.py
murilx: If you go the pydantic route, there's `BuildStatusCount` in `commonDetails` that's the exact same thing as `BuildStatusCoundDict`. But there may be a reason to use `TypedDict` in this case

backend/kernelCI_app/views/hardwareView.py
murilx: `build_status` could be `None` if for some reason `test["build__valid"]` was not mapped in `build_status`

```sh
Python 3.12.7 (main, Nov 18 2024, 15:03:02) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> build_status = {True: "valid", False: "invalid", None: "null"}
>>> build_status.get(True)
'valid'
>>> build_status.get(False)
'invalid'
>>> build_status.get(None)
'null'
>>> build_status.get('other thing')
>>> build_status.get('other thing') is None
True
```

But in theory this should never happen since `test[build_valid]` in the database is a nullable boolean

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,120 @@
...
+
+        status_count = "test_status_summary"
+
+        if test["path"].startswith("boot.") or test["path"] == "boot":
+            status_count = "boot_status_summary"
+
+        if test["status"] is None:
+            self.hardware[current_hardware][status_count]["NULL"] += 1
+        else:
+            self.hardware[current_hardware][status_count][test["status"]] += 1
```
Francisco2002, line 89:
```suggestion
        test_status = "NULL" if test["status"] is None else test["status"]:
        
        self.hardware[current_hardware][status_count][test_status] += 1
```

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -13,68 +14,120 @@
...
+
+        build_key = test["build__id"] + current_hardware
+        if build_key in self.processed_builds:
+            return
+
+        self.processed_builds.add(build_key)
+
+        build_status = build_status_map.get(test["build__valid"])
```
Francisco2002, line 97:
```suggestion
        build_status = build_status_map.get(test["build__valid"], "null")
```

backend/kernelCI_app/views/hardwareView.py
murilx: if `build_status_map.get(...` returns None it means there's a problem in the code. Or the database is sending unexpected build status and we should start processing them, or we are messing in some other place (our query is returning a strange value, or other thing). So I believe it shouldn't count as "null" in this case. Maybe keep the safeguard to future-proof the code or leave it with nothing so it purposefully breaks when some of these conditions are met

murilx (commented):
We have cases where the same compatible can have mulitple platforms (this does not matter the other way around since `environment_compatible` has priority over `platform` when creating a hardware details page). One example is [amlogic,a311d (localhost)](http://localhost:5173/hardware/amlogic%2Ca311d?et=1738846800&hs=amlogic&st=1738414800). Shouldn't we account for this case? Maybe add a tooltip in the hardware listing page to show all of them or even just adding all of them to the cell with something `platforms.map(platform => <span>{platform}<br /></span>` 

One other option would be to leave it like it is now and discuss this in the mid spring meeting

murilx (commented):
If we want to keep both tree and hardware listing consistent, the search filter in hardware listing should also filter in the platforms column (like we can filter by tree commits in the tree listing for example)

backend/kernelCI_app/views/hardwareView.py
WilsonNet: that is personal preference, I think the bigger one is easier to read

backend/kernelCI_app/views/hardwareView.py
MarceloRobert: I'd say to add at least an `else: log_message("something")` here to know if this happens in the future then

backend/kernelCI_app/views/hardwareView.py
WilsonNet: done

backend/kernelCI_app/views/hardwareView.py
WilsonNet: done

MarceloRobert (approved):
LGTM, nice work

backend/kernelCI_app/typeModels/hardwareListing.py
```diff
@@ -0,0 +1,38 @@
...
+
+class HardwareResponse(BaseModel):
+    hardware: List[HardwareItem]
+
+
+# Since OpenAPI does not support timestamp as datetime we add an extra model just for
+# documentation purposes. This model is not used in the code.
+# TODO Remove timestamp from the api and this model
+class HardwareQueryParamsDocumentationOnly(BaseModel):
+    origin: str = Field(default=DEFAULT_ORIGIN)
```
murilx, line 30: nit: I believe there's no need for `Field(default=DEFAULT_ORIGIN)` here. Only this should be sufficient
```suggestion
    origin: str = DEFAULT_ORIGIN
```

backend/kernelCI_app/typeModels/hardwareListing.py
```diff
@@ -0,0 +1,38 @@
...
+    startTimestampInSeconds: str
+    endTimeStampInSeconds: str
+
+
+class HardwareQueryParams(BaseModel):
+    origin: str = Field(default=DEFAULT_ORIGIN)
```
murilx, line 36: ditto about `Field`

backend/kernelCI_app/views/hardwareView.py
```diff
@@ -1,102 +1,190 @@
...
+
+    @extend_schema(
+        parameters=[HardwareQueryParamsDocumentationOnly], responses=HardwareResponse
+    )
+    def get(self, request: Request):
+        try:
+            query_params = HardwareQueryParams(
+                start_date=request.GET.get("startTimestampInSeconds"),
+                end_date=request.GET.get("endTimeStampInSeconds"),
+                origin=request.GET.get("origin"),
```
murilx, line 171: That's nice to see!

murilx (approved):
Well done!

PR #889, opened 2025-02-06
Feat: add treeIssues endpoint and detailed issues info

### LINKED ISSUE #869: IssueList backend changes
Subissue from #868 
- [ ] Add specific endpoint to find issues for build/boots/tests of a specific tree
- [ ] Change query to return the start_date of the first version of the list of issues
- [ ] Change query to return the trees where this issue appears

MarceloRobert:
## Changes
- Adds a new endpoint to get the detailed information of a list of issues (id and version)
- Adds query to get the timestamp of the first incident of an issue
- Adds query to get the trees where that issue_id and issue_version have had incidents
- The list of trees is also used to add specific tags to the issues

## How to test
Acquire a list of issues that you want to get more details from, then go to `http://localhost:5173/issues/extras/` and in the swagger UI you can place a body for the request, such as:
```
{
    "issues": [
        ["maestro:e602fca280d85d8e603f7c0aff68363bb0cd7993", 1],
        ["maestro:da694c56147298d223ee432ad8d6a8ee311b773a", 1]
    ]
}
```

You can verify if the trees are coming correctly for an issue by running the following query (change the issue id and version at the end):
```
SELECT DISTINCT
	ON (
		C.TREE_NAME,
		C.GIT_REPOSITORY_URL,
		C.GIT_REPOSITORY_BRANCH
	) C.ID,
	C.TREE_NAME,
	C.GIT_REPOSITORY_URL,
	C.GIT_REPOSITORY_BRANCH,
	IC.ISSUE_ID,
	IC.ISSUE_VERSION
FROM
	INCIDENTS IC
	LEFT JOIN TESTS T ON IC.TEST_ID = T.ID
	JOIN BUILDS B ON (
		IC.BUILD_ID = B.ID
		OR T.BUILD_ID = B.ID
	)
	JOIN CHECKOUTS C ON B.CHECKOUT_ID = C.ID
WHERE
	IC.ISSUE_ID = 'redhat:issue_3332'
	AND IC.ISSUE_VERSION = 1734005602
```

The return of the new endpoint is the primary key for the issues (id and version) and the extra information only.

Closes #869

### DIFF
--- a/backend/kernelCI_app/helpers/issueExtras.py
+++ b/backend/kernelCI_app/helpers/issueExtras.py
@@ -1,0 +1,159 @@
+from collections import defaultdict
+from typing import Dict, List, Set, Tuple
+
+from kernelCI_app.helpers.logger import log_message
+from django.db.models import Min
+from kernelCI_app.typeModels.issues import (
+    IssueWithExtraInfo,
+    ProcessedExtraDetailedIssues,
+    TreeSetItem,
+)
+from kernelCI_app.models import Incidents, Checkouts
+
+
+class TagUrls:
+    MAINLINE_URL = "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
+    STABLE_URL = "https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git"
+    LINUX_NEXT_URL = (
+        "https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git"
+    )
+
+
+def process_issues_extra_details(
+    *,
+    issue_key_list: List[Tuple[str, int]],
+    processed_issues_table: ProcessedExtraDetailedIssues,
+) -> None:
+    # TODO: combine both queries into one
+    assign_issue_first_seen(
+        issue_key_list=issue_key_list,
+        processed_issues_table=processed_issues_table,
+    )
+    assign_issue_trees(
+        issue_key_list=issue_key_list,
+        processed_issues_table=processed_issues_table,
+    )
+
+
+def assign_issue_first_seen(
+    *,
+    issue_key_list: List[Tuple[str, int]],
+    processed_issues_table: ProcessedExtraDetailedIssues,
+) -> None:
+    issue_id_list: List[str] = []
+    versions_per_issue: Dict[str, Set[int]] = defaultdict(set)
+    for issue_key in issue_key_list:
+        issue_id, issue_version = issue_key
+        issue_id_list.append(issue_id)
+        versions_per_issue[issue_id].add(issue_version)
+
+    # TODO: use '=' instead of 'IN' when the list has a single element
+    incident_records = (
+        Incidents.objects.filter(issue_id__in=issue_id_list)
+        .values("issue_id")
+        .annotate(
+            first_seen=Min("field_timestamp"),
+            issue_version=Min("issue_version"),
+        )
+    )
+
+    for record in incident_records:
+        record_issue_id = record["issue_id"]
+        first_seen = record["first_seen"]
+
+        processed_issue_from_id = processed_issues_table.get(record_issue_id)
+        if processed_issue_from_id is None:
+            processed_issues_table[record_issue_id] = {}
+            processed_issue_from_id = processed_issues_table[record_issue_id]
+
+        for version in versions_per_issue[record_issue_id]:
+            processed_issue_from_version = processed_issue_from_id.get(version)
+            if processed_issue_from_version is None:
+                processed_issue_from_id[version] = IssueWithExtraInfo(
+                    id=record_issue_id, version=version, first_seen=first_seen
+                )
+            else:
+                processed_issue_from_version.first_seen = first_seen
+
+
+def assign_issue_trees(
+    *,
+    issue_key_list: List[Tuple[str, int]],
+    processed_issues_table: ProcessedExtraDetailedIssues,
+) -> None:
+
+    # TODO: use '=' instead of 'IN' when the list has a single element
+    tuple_param_list = []
+    params = {}
+
+    for index, key in enumerate(issue_key_list):
+        id_key = f"id{index}"
+        version_key = f"version{index}"
+
+        tuple_string = f"(%({id_key})s, %({version_key})s)"
+
+        tuple_param_list.append(tuple_string)
+        params[id_key] = key[0]
+        params[version_key] = key[1]
+    tuple_str = ", ".join(tuple_param_list)
+
+    trees_records = Checkouts.objects.raw(
+        f"""
+        SELECT DISTINCT
+          ON (
+            C.TREE_NAME,
+            C.GIT_REPOSITORY_URL,
+            C.GIT_REPOSITORY_BRANCH,
+            IC.ISSUE_ID,
+            IC.ISSUE_VERSION
+          )
+          C.ID,
+          C.TREE_NAME,
+          C.GIT_REPOSITORY_URL,
+          C.GIT_REPOSITORY_BRANCH,
+          IC.ISSUE_ID,
+          IC.ISSUE_VERSION
+        FROM
+          INCIDENTS IC
+          LEFT JOIN TESTS T ON IC.TEST_ID = T.ID
+          LEFT JOIN BUILDS B ON (
+            IC.BUILD_ID = B.ID
+            OR T.BUILD_ID = B.ID
+          )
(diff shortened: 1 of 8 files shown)


murilx (approved):
Worked well on my tests! Nice job

backend/kernelCI_app/views/treeIssuesView.py
```diff
@@ -0,0 +1,242 @@
...
+            "test": {},
+        }
+
+        self.build_issues: List[NestedIssueList] = []
+        self.boot_issues: List[NestedIssueList] = []
+        self.test_issues: List[NestedIssueList] = []
+
+        self.failed_builds_with_unknown_issues = 0
+        self.failedTestsWithUnknownIssues = 0
+        self.failedBootsWithUnknownIssues = 0
```
WilsonNet, line 69: snake case

backend/kernelCI_app/views/treeIssuesView.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class TreeIssues(APIView):
+    def __init__(self):
+        self.filters = None
+        self.processed_builds = set()
+        self.processed_tests = set()
+
+        self.testIssuesTable: IssueDict = {}
+        self.bootsIssuesTable: IssueDict = {}
```
WilsonNet, line 54: snake case

backend/kernelCI_app/views/treeIssuesView.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @extend_schema(
+        parameters=[TreeQueryParameters],
+        responses=TreeDetailsIssuesResponse,
+        methods=["GET"],
+    )
+    def get(self, request, commit_hash: str | None) -> Response:
+        self.filters = FilterParams(request)
+
+        treeData = get_tree_details_data(request, commit_hash)
```
WilsonNet, line 208: snake case

WilsonNet, line 208: this query is unnecessary, since the frontend should have that data, but if the case was maintainability you can use the query cache for this

backend/kernelCI_app/typeModels/issues.py
```diff
@@ -14,4 +15,37 @@ class Issue(BaseModel):
...
+    trees: Optional[List[TreeSetItem]] = []
+    tags: Optional[Set[PossibleIssueTags]] = set()
+
+
+class NestedIssueDict(BaseModel):
+    id: str
+    first_seen: Optional[datetime]
+    occurrences: Dict[int, DetailedIssue]
+
+    def to_list_model(self):
```
WilsonNet, line 40: I believe it is better for the frontend if `Ocurrences` return a `Dict` 

That way to attach new data they can just do `issuesLazyLoaded[issueId][issueVersion]` widthout having to resort on iterations or sorting 

but you can also go with a complete replace strategir where you replace the entire old data with the new data, for that to work you will need both to be equally sorted

backend/kernelCI_app/views/treeIssuesView.py
WilsonNet: thinking again, this won't work with `hardwareDetails`, if it is better to receive a request body with all the necessary data for the issues now, otherwise we will have to rework the frontend

backend/kernelCI_app/helpers/treeIssues.py
```diff
@@ -0,0 +1,146 @@
...
+        issue_id = record.issue_id
+        issue_version = record.issue_version
+        git_repository_url = record.git_repository_url
+        git_repository_branch = record.git_repository_branch
+
+        tab: PossibleTable = "build"
+        if (issue_id, issue_version) in boot_issue_keys:
+            tab = "boot"
+        elif (issue_id, issue_version) in test_issue_keys:
+            tab = "test"
```
Francisco2002, line 119: nit: create a function to this logic. Something like `get_tab_name(issue_id, issue_version) -> PossibleTable`

MarceloRobert: With the recent changes this won't be needed

backend/kernelCI_app/typeModels/issues.py
MarceloRobert: I changed the response to be a `Dict[str, Dict[int, IssueWithExtraInfo]]`, so now your approach would work

backend/kernelCI_app/views/treeIssuesView.py
MarceloRobert: these variables were written in so that I wouldn't have to change their use in the treeDetails helper file

backend/kernelCI_app/views/issueExtrasView.py
```diff
@@ -0,0 +1,55 @@
...
+from pydantic import ValidationError
+
+
+# disable django csrf protection https://docs.djangoproject.com/en/5.0/ref/csrf/
+# that protection is recommended for ‘unsafe’ methods (POST, PUT, and DELETE)
+# but we are using POST here just to follow the convention to use the request body
+# also the csrf protection require the usage of cookies which is not currently
+# supported in this project
+@method_decorator(csrf_exempt, name="dispatch")
+class IssueExtras(APIView):
```
WilsonNet, line 26: nit: I think `IssueExtraDetails` is a better name

backend/kernelCI_app/typeModels/issues.py
```diff
@@ -1,17 +1,46 @@
...
+from typing import Dict, List, Literal, Optional, Set, Tuple
 from pydantic import BaseModel
 
 
 class IncidentInfo(BaseModel):
     incidentsCount: int
 
 
-class Issue(BaseModel):
+class BaseIssueKeys(BaseModel):
```
murilx, line 10: nit: `Base` seems unnecessary in this name

backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -0,0 +1,158 @@
...
+        # TODO: create a more modular way to add tags
+        if git_repository_url == MAINLINE_URL and git_repository_branch == "master":
+            current_detailed_issue.tags.add("mainline")
+        elif git_repository_url == STABLE_URL and git_repository_branch is not None:
+            current_detailed_issue.tags.add("stable")
+        elif git_repository_url == LINUX_NEXT_URL and (
+            git_repository_branch == "master"
+            or git_repository_branch == "pending-fixes"
+        ):
+            current_detailed_issue.tags.add("linux-next")
```
WilsonNet, line 158: give it a try to this one 
```python
match (git_repository_url, git_repository_branch):
    case (MAINLINE_URL, "master"):
        current_detailed_issue.tags.add("mainline")
    case (STABLE_URL, branch) if branch is not None:
        current_detailed_issue.tags.add("stable")
    case (LINUX_NEXT_URL, "master" | "pending-fixes"):
        current_detailed_issue.tags.add("linux-next")
```

backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -0,0 +1,160 @@
...
+            continue
+
+        current_detailed_issue.trees.append(
+            TreeSetItem(
+                tree_name=record.tree_name,
+                git_repository_branch=git_repository_branch,
+            )
+        )
+
+        # TODO: create a more modular way to add tags
```
murilx, line 153: Would the pattern matching be a more modular way of adding tags? If that's the case you can remove the TODO

WilsonNet (approved):
looks nice

murilx (approved):
Re-approving after the changes. Seems to be working nice, just a minor comment about the documentation

backend/kernelCI_app/typeModels/issues.py
```diff
@@ -1,17 +1,46 @@
...
+
+
+class IssueWithExtraInfo(IssueKeys):
+    first_seen: Optional[datetime] = None
+    trees: Optional[List[TreeSetItem]] = []
+    tags: Optional[Set[PossibleIssueTags]] = set()
+
+
+class IssueExtraDetailsRequest(BaseModel):
+    issues: List[Tuple[str, int]]
```
murilx, line 39: Unfortunately this type doesn't seem to be supported by the schema creation since it's showing as `issues: null` in the documentation but I don't know if there's a easy way to circumvent this. One way would be to create a model with `id: str` and `version: int` to replace `Tuple[str, int]` but that would mean doing more refactoring to the code. Because of this I'll keep this as an optional recommendation but definitely preferable.

PR #890, opened 2025-02-06
feat: add documentation to some endpoints

### LINKED ISSUE #808: tree listing

Francisco2002:
- [ ] tree lsiting
- [ ] tree listing fast
- [ ] tree latest
- [ ] tree details full

Part of #86
Close #808


backend/kernelCI_app/typeModels/treeListing.py
```diff
@@ -0,0 +1,44 @@
...
+    git_repository_branch: Checkout__GitRepositoryBranch
+    git_repository_url: Checkout__GitRepositoryUrl
+    start_time: Timestamp
+
+
+class TreeListingResponse(RootModel):
+    root: List[Checkout]
+
+
+class TreeListingParameters(BaseModel):
```
murilx, line 42: nit: Would call this `TreeListingQueryParameters` even though this name is a little bit long

backend/kernelCI_app/typeModels/treeListing.py
```diff
@@ -0,0 +1,44 @@
...
+    origin: str = "maestro"
```
murilx, line 43: Since we are not using this class to validate the query params we receive, I'd recommend against adding a default value. It can make it easier in the documentation for someone to try and fire a request there but at the same time the parameter appears as optional in the documentation, which can be misleading and plain wrong since our endpoint does not handle an optional `origin` param. If we were using this class to validate the query params this wouldn't be a problem but since that's not the case (at least for now) than it's problematic

backend/kernelCI_app/views/treeView.py
```diff
@@ -138,15 +186,23 @@ def get(self, request):
...
+        with connection.cursor() as cursor:
+            cursor.execute(query, params)
+            checkouts = cursor.fetchall()
 
         if not checkouts:
-            return create_error_response(
-                error_message="Trees not found", status_code=HTTPStatus.OK
+            return create_api_error_response(error_message="Trees not found", status_code=HTTPStatus.OK)
+
+        checkouts = [self._sanitize_tree(checkout) for checkout in checkouts]
```
murilx, line 199: As discussed in the meeting me and Francisco had, we could remove this `_sanatize_tree` using Pydantic alias but that would mean that instead of `checkouts` being a list of tuples, we would need to change this to a list of objects, like `checkouts = [{"checkout": checkouts[0]}, {"checkout": checkouts[1]}, ...]` because of the way `AliasPath` of pydantic works (it gets the element in index `n` of an object field that has an iterable as value). We judged both of the solutions to be more or less equal so decided not to change at first and would like the opinion of the other devs in this matter

backend/kernelCI_app/typeModels/treeDetails.py
```diff
@@ -30,6 +42,15 @@ class TreeDetailsBuildsResponse(BaseModel):
...
     git_branch: str
+
+
+class TreeDetailsFullResponse(
+    TreeDetailsBuildsResponse,
+    CommonDetailsBootsResponse,
+    CommonDetailsTestsResponse,
+    SummaryResponse
+):
+    pass
```
murilx, line 56: I believe this makes sense in this code (since the full endpoint must be the combination of all the other tree endpoints) but I wanna see what the others think as well

backend/kernelCI_app/typeModels/treeDetails.py
```diff
@@ -4,15 +4,27 @@
...
+from kernelCI_app.constants.general import DEFAULT_ORIGIN
 
 
 class TreeLatestPathParameters(BaseModel):
     tree_name: str
     branch: str
 
 
+class TreeLatestQueryParameters(BaseModel):
+    origin: str = DEFAULT_ORIGIN
```
murilx, line 21: In this case a default value to origin makes sense. Nice

backend/kernelCI_app/typeModels/treeListing.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+class TestStatusCount(BaseModel):
+    pass_count: int = Field(default=0, alias='pass')
+    error_count: int = Field(default=0, alias='error')
+    fail_count: int = Field(default=0, alias='fail')
+    skip_count: int = Field(default=0, alias='skip')
+    miss_count: int = Field(default=0, alias='miss')
+    done_count: int = Field(default=0, alias='done')
+    null_count: int = Field(default=0, alias='null')
```
murilx, line 24: I'm not sure but I believe `COUNT(exp)` will never return None so this default value shouldn't be necessary

murilx: I also did something like this in my PR, I think it make sense to combine the two classes when rebasing the PR. (This is just something to think in the future and maybe I'll have to do it if your PR is merged thinks but I think it's worth the heads up)

Francisco2002: We can unify this in a type only and reuse in both cases. I will do the same as you did in your PR, removing this default value

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -159,75 +165,86 @@ def _sanitize_rows(self, rows):
...
-                        "architectures": list(self.test_arch_summary.values()),
-                        "configs": self.test_configs,
-                        "issues": self.testIssues,
-                        "unknown_issues": self.failedTestsWithUnknownIssues,
-                        "environment_compatible": self.testEnvironmentCompatible,
-                        "environment_misc": self.testEnvironmentMisc,
-                        "fail_reasons": self.testFailReasons,
-                        "failed_platforms": list(self.testPlatformsWithErrors),
-                    },
+        result = {
```
MarceloRobert, line 185: you could create the result directly as a `TreeDetailsFullResponse()` inside of the try catch for validation

murilx (approved):
Seems to be working fine. I think we might still have some discussion to do about `_sanitize_tree` if we'd like that the function returns a dict with the already correct field names or if we'd like to use pydantic to sort this out, but for now I think this is working well

MarceloRobert (approved):
It is good as it is, even though as we discussed we could change the `_sanitize_tree`. Pre approving

PR #918, opened 2025-02-12
Feat: add link to tree details in test details page

### LINKED ISSUE #914: Add link to the treeDetails from which a test is related
We currently have a link to the tree details from a test details through the breadcrumb only, it would be useful if there was a link to the treeDetails in the git checkout too

Francisco2002:
### **How to test**
Go to a test details page and click in git commit hash. You should be redirect to a tree details page related with this test.

Close #914


MarceloRobert (approved):
Seems to be working well

WilsonNet (approved):
nice job

PR #925, opened 2025-02-13
Refactor: Change issues from unknown to uncategorized

### LINKED ISSUE #910: Refactor "Unknown" issues to "Uncategorized"
Rename how it is being displayed in the frontend from "Unknown" issues in the IssueList and filters to "Uncategorized"
Change the backend treatment to consider a flag for unknown issues, since those aren't real issues

MarceloRobert:
- Renames Unknown issues to Uncategorized issues
- Changes the behavior of the filters to return a flag for if it has uncategorized issues, instead of considering them as a normal issue
- Refactors TreeDetailsView and TreeDetailsSummaryView to use pydantic models directly

## How to test
Go to a tree or hardware details page where there are builds/boots/tests with uncategorized issues
Filter them through the issues list and through the filter modal
Check if everything is being filtered as intended

Closes #910

### DIFF
--- a/backend/kernelCI_app/helpers/filters.py
+++ b/backend/kernelCI_app/helpers/filters.py
@@ -1,6 +1,8 @@
 from typing import Optional, Dict, List, Tuple, TypedDict, Literal, Any, Union
 from django.http import HttpResponseBadRequest
 import re
+from kernelCI_app.constants.general import UNCATEGORIZED_STRING
+from kernelCI_app.helpers.commonDetails import PossibleTabs
 from kernelCI_app.typeModels.databases import PASS_STATUS, failure_status_list
 from kernelCI_app.utils import getErrorResponseBody
 from kernelCI_app.constants.general import UNKNOWN_STRING
@@ -20,8 +22,18 @@ def is_known_issue(issue_id: Optional[str], issue_version: Optional[int]) -> boo
     return (
         issue_id is not None
         and issue_version is not None
-        and issue_id is not UNKNOWN_STRING
+        and issue_id is not UNCATEGORIZED_STRING
     )
+
+
+def is_unknown_build_issue(
+    issue_id: Optional[str],
+    issue_version: Optional[int],
+    incident_test_id: Optional[str],
+) -> bool:
+    is_unknown_issue = not is_known_issue(issue_id, issue_version)
+    is_issue_from_build = incident_test_id is None
+    return is_unknown_issue and is_issue_from_build
 
 
 def is_exclusively_build_issue(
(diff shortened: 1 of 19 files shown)


backend/kernelCI_app/helpers/filters.py
```diff
@@ -283,7 +284,11 @@ def __init__(self, data: Dict, process_body=False) -> None:
         self.filterTestPath = ""
         self.filterBootPath = ""
         self.filterBuildValid = set()
-        self.filterIssues = {"build": set(), "boot": set(), "test": set()}
+        self.filterIssues: Dict[PossibleTabs, Set[str, Optional[int]]] = {
```
murilx, line 287: Is `Set[str, Optional[int]]` a valid type? I couldn't find anything to support this annotation. You might have forgot a `Union` here, I'm not sure

backend/kernelCI_app/views/treeDetailsSummaryView.py
```diff
@@ -168,64 +187,71 @@ def get(self, request, commit_hash: str | None):
...
+                        ],
+                    ),
+                ),
+            )
         except ValidationError as e:
-            return Response(data=e.errors())
+            return Response(data=e.errors(), status=HTTPStatus.INTERNAL_SERVER_ERROR)
 
-        return Response(response)
+        return Response(valid_response.model_dump(), status=HTTPStatus.OK)
```
murilx, line 257: `Response` already defaults status to OK
https://www.django-rest-framework.org/api-guide/responses/#response

backend/kernelCI_app/helpers/filters.py
```diff
@@ -283,7 +299,11 @@ def __init__(self, data: Dict, process_body=False) -> None:
         self.filterTestPath = ""
         self.filterBootPath = ""
         self.filterBuildValid = set()
-        self.filterIssues = {"build": set(), "boot": set(), "test": set()}
+        self.filterIssues: dict[PossibleTabs, set[tuple[str, Optional[int]]]] = {
```
Francisco2002, line 302:
```suggestion
        self.filterIssues: dict[PossibleTabs, set[Tuple[str, Optional[int]]]] = {
```

Francisco2002 (approved):
Nice job!! Just a comment

backend/kernelCI_app/helpers/filters.py
MarceloRobert: This is being used to start following #930

murilx: Nice call. Since this ticket is not a priority we shouldn't change what we've already done but I think we should stop to use the deprecated typing for new additions

backend/kernelCI_app/views/treeDetailsSummaryView.py
```diff
@@ -1,5 +1,7 @@
+from typing import Dict
```
murilx, line 1: deprecated
https://docs.python.org/3/library/typing.html#typing.Dict

murilx (approved):
Worked well on my tests

PR #932, opened 2025-02-14
Chore: update react 19

### LINKED ISSUE #931: Update to React 19

Francisco2002:
Close #931 

### **How to test**
- run the frontend to test if shows any strange behavior.
- try apply lint rules


dashboard/tsconfig.json
```diff
@@ -10,6 +10,11 @@
   ],
   "compilerOptions": {
     "baseUrl": ".",
+    "jsx": "react-jsx",
+    "strict": true, // Optional: Enables strict mode
+    "skipLibCheck": true,
```
WilsonNet, line 15: is this needed?

Francisco2002: This lines was added to handle a dependency that is no longer in the project. I will remove it!

dashboard/src/components/Tabs/FilterLink.tsx
```diff
@@ -11,7 +13,7 @@ interface FilterLinkProps {
   diffFilter: TFilter;
   className?: string;
   from?: LinkProps['from'];
-  to?: LinkProps['to'];
+  to?: LinkProps['to'] | string;
```
murilx, line 16: Might be good to check why the linter is complaining about this typing later. For now I don't think it's important for this PR

murilx (commented):
Filtering through the modal seems to be working fine, but filtering using the cards (like clicking in a hardware or an issue or a config) is not working.

Another minor but kind of irritating problem is that every time I change filters (in this case it's happening when I click in a card to filter for that row or when I change the table filter), it's going back to the top of the page, which is not happening on the dashboard right now.

dashboard/src/components/Tabs/FilterLink.tsx
WilsonNet, line 16: add a TODO here for that, also you can do  `to?: LinkProps['to'] | (string & {});` which will tricky the compiler to not reduce this prop to string and will preserve autocomplete

WilsonNet (approved):
LGTM

dashboard/src/components/Tabs/tabsUtils.ts
```diff
@@ -33,29 +31,22 @@ export const cleanFalseFilters = (diffFilter: TFilter): TFilter => {
...
-  if (!currentDiffFilter) {
-    return {};
-  }
-
-  // This seems redundant but we do this to keep the pointer to newFilter[filterSection]
-  newFilter[filterSection] = newFilter[filterSection] ?? {};
+  const newFilter = structuredClone(currentDiffFilter) || {};
 
-  const configs = newFilter[filterSection];
+  const configs = newFilter[filterSection] ?? {};
```
MarceloRobert, line 41: nit: I don't think that `configs` is the best name here since a filter can be from the "config" property or any other property

MarceloRobert (commented):
It seems that every time we click on something it is going to the top of the page, when it wasn't doing that before

murilx (approved):
Worked well on my test. Great job!

PR #933, opened 2025-02-14
feat: add regex support to tree and hardware filters

### LINKED ISSUE #913: Add Regex to listing search
Adding regex to the listing search (especially hardware listing) will allow for a more powerful search, including multiple and different results

WilsonNet:
- Updated TreeListingPage to use matchesRegexOrIncludes for filtering.
- Updated HardwareListingPage to use matchesRegexOrIncludes and
  includesInAnStringOrStringArray for filtering.
  functions to string utility library.

Closes #913

## How to test
Just use a regex on the search bars or click on the link below
http://localhost:5173/tree?ls=50&ts=android%7Cbroonie%7Cmainline

### DIFF
--- a/dashboard/src/lib/string.ts
+++ b/dashboard/src/lib/string.ts
@@ -47,3 +47,31 @@ export const truncateUrl = (
   const lastPath = pathname ? pathname.slice(-endPathLength) : '';
   return `${domain}...${lastPath}`;
 };
+
+export const matchesRegexOrIncludes = (
+  text: string | undefined | null,
+  pattern: string,
+): boolean => {
+  if (!text) {
+    return false;
+  }
+  try {
+    const regex = new RegExp(pattern, 'i');
+    return regex.test(text);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (e) {
+    return text.includes(pattern);
+  }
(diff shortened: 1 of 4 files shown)


dashboard/src/lib/string.ts
```diff
@@ -47,3 +47,31 @@ export const truncateUrl = (
...
+  pattern: string,
+): boolean => {
+  if (!text) {
+    return false;
+  }
+  try {
+    const regex = new RegExp(pattern, 'i');
+    return regex.test(text);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  } catch (e) {
```
MarceloRobert, line 62: Nit: you could use a `console.error(e)` and remove the eslint comment if that's not gonna pollute the console

MarceloRobert (approved):
Working well, this will be a nice addition

Francisco2002 (approved):
Amazing!!
Nice job!

dashboard/src/lib/string.ts
WilsonNet: providing invalid regex is a pretty common use case, specially in the middle of typing, or maybe there is a tree which the regex is an invalid regex

PR #949, opened 2025-02-17
Refactor: rename build valid/invalid to pass/fail

### LINKED ISSUE #922: Rename build valid/invalid to pass/fail
On buildDetails and see if it shows in any other page as well

MarceloRobert:
This change is only for the status display in the builds table and build details, the variables are still called valid since that is how it is called in the backend.
The name in the diffFilter was left unchanged, but the table filter was unified with the way it is being done in the other tables

## How to test
Go to a buildDetails page and check the change in the "status" field.
Go to a page with a build table (such as treeDetails, hardwareDetails or issueDetails) and check the table to see if it is behaving correctly, also check the status filters.
Try to filter by build status and check if it is behaving correctly, also check the url to see if it is called by the correct name.

Closes #922

### DIFF
--- a/dashboard/src/components/BuildsTable/BuildsTable.tsx
+++ b/dashboard/src/components/BuildsTable/BuildsTable.tsx
@@ -19,11 +19,11 @@ import BaseTable, { TableHead } from '@/components/Table/BaseTable';
 import { PaginationInfo } from '@/components/Table/PaginationInfo';
 import TableStatusFilter from '@/components/Table/TableStatusFilter';
 import { TableBody, TableCell, TableRow } from '@/components/ui/table';
-import type {
-  AccordionItemBuilds,
-  BuildsTableFilter,
+import {
+  possibleTableFilters,
+  type AccordionItemBuilds,
+  type PossibleTableFilters,
 } from '@/types/tree/TreeDetails';
-import { possibleBuildsTableFilter } from '@/types/tree/TreeDetails';
 
 import WrapperTableWithLogSheet from '@/pages/TreeDetails/Tabs/WrapperTableWithLogSheet';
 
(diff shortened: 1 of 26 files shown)


dashboard/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -96,26 +98,22 @@ export function BuildsTable({
...
-        if (dataFilter) {
-          acc[currentFilter] = dataFilter?.reduce(
-            (total, row) => (row.status === currentFilter ? total + 1 : total),
-            0,
-          );
-        }
-        return acc;
-      },
-      {} as Record<(typeof possibleBuildsTableFilter)[number], number>,
-    );
```
MarceloRobert, line 117: Since there were two `reduce`s, the dataFilter was being looped through as many times as the amount of filters. This has been changed to follow the way it is being done in the BootsTable

dashboard/src/components/BuildsTable/BuildsTable.tsx
```diff
@@ -125,34 +123,34 @@ export function BuildsTable({
       {
         label: intl.formatMessage(
           { id: 'global.allCount' },
-          { count: filterCount[possibleBuildsTableFilter[2]] },
+          { count: filterCount[possibleTableFilters[0]] },
         ),
-        value: possibleBuildsTableFilter[2],
-        isSelected: filter === possibleBuildsTableFilter[2],
+        value: possibleTableFilters[0],
+        isSelected: filter === possibleTableFilters[0],
```
MarceloRobert, line 129: The order of `possibleBuildsTableFilter` was different than `possibleTestsTableFilter`, that's why there is a change between indexes 2 and 0 ('all' and 'fail')

dashboard/src/utils/utils.ts
```diff
@@ -128,10 +128,14 @@ const isBuildError = (build_valid: boolean | null): number => {
...
-  build_valid: boolean | null,
+export const getBuildStatus = (
+  build_valid?: boolean | null,
 ): AccordionItemBuilds['status'] => {
-  return build_valid === null ? 'null' : build_valid ? 'valid' : 'invalid';
+  return build_valid === null || build_valid === undefined
+    ? 'null'
+    : build_valid
+      ? 'pass'
+      : 'fail';
```
WilsonNet, line 138:
```suggestion
  if build_vallid === true; 
    return 'pass'
  if build_valid === false;
     reuturn 'fail';
  else:
    return null
     
```

WilsonNet: much easier to read

murilx (approved):
It's great we could unify both `TestsTableFilter` and `BuildsTableFilter`! LGTM

WilsonNet (approved):
LGTM

PR #956, opened 2025-02-18
feat(issue-details): add remaining extra data

### LINKED ISSUE #915: Add extra issue details to IssueDetails page
Add all the extra information to the issueDetails endpoint, make it fetch all data in a single request

murilx:
- Updated issue details view
- Created the BranchBadge component
- Added subtitle to the Section component
- Added tags in the issue details page as a subtitle to the general section

Other extra data is already present in the 'First Incident Data' or indirectly in the Builds/Tests tables

Closes #915

## Visual Reference
http://localhost:5173/issue/maestro%3Ada694c56147298d223ee432ad8d6a8ee311b773a?iv=1&ls=10&p=bt&ti%7Cc=v6.14-rc3-27-gb92c76cc17f9&ti%7Cch=b92c76cc17f9780732ed9a145f4e24fc4439bd11&ti%7Cgb=for-next&ti%7Cgu=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fbroonie%2Fspi.git&ti%7Ct=broonie-spi
image

### DIFF
--- a/backend/kernelCI_app/views/issueDetailsView.py
+++ b/backend/kernelCI_app/views/issueDetailsView.py
@@ -5,7 +5,7 @@ from kernelCI_app.helpers.issueDetails import fetch_latest_issue_version
 from kernelCI_app.models import Issues
 from kernelCI_app.typeModels.issueDetails import (
     IssueDetailsPathParameters,
-    IssueDetailsRequest,
+    IssueDetailsQueryParameters,
     IssueDetailsResponse,
 )
 from kernelCI_app.helpers.issueExtras import process_issues_extra_details
@@ -44,26 +44,31 @@ class IssueDetails(APIView):
         return query
 
     @extend_schema(
-        request=IssueDetailsRequest, responses=IssueDetailsResponse, methods=["GET"]
+        parameters=[IssueDetailsQueryParameters],
+        responses=IssueDetailsResponse,
+        methods=["GET"],
     )
     def get(self, _request, issue_id: Optional[str]) -> Response:
         try:
+            version = _request.GET.get("version")
             parsed_params = IssueDetailsPathParameters(issue_id=issue_id)
+            parsed_query = IssueDetailsQueryParameters(version=version)
         except ValidationError as e:
             return Response(data=e.json(), status=HTTPStatus.BAD_REQUEST)
 
-        version = _request.GET.get("version")
-        if version is None:
+        if parsed_query.version is None:
             version_row = fetch_latest_issue_version(issue_id=parsed_params.issue_id)
             if version_row is None:
                 return create_api_error_response(
                     error_message="Issue not found", status_code=HTTPStatus.OK
                 )
-            version = version_row["version"]
+            parsed_query.version = version_row["version"]
 
-        issue_data = self._fetch_issue(issue_id=parsed_params.issue_id, version=version)
+        issue_data = self._fetch_issue(
+            issue_id=parsed_params.issue_id, version=parsed_query.version
+        )
         process_issues_extra_details(
-            issue_key_list=[(issue_id, version)],
+            issue_key_list=[(issue_id, parsed_query.version)],
             processed_issues_table=self.processed_issue_extras,
         )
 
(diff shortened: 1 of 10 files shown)


dashboard/src/components/IssueDetails/IssueDetails.tsx
```diff
@@ -120,6 +122,18 @@ export const IssueDetails = ({
     });
   }, [data?.extra, formatMessage, issueId]);
 
+  const tagPills = useMemo(() => {
+    if (versionNumber !== undefined) {
```
MarceloRobert, line 126: There is no reason for this condition, even if `versionNumber` is undefined, the data for the tags will be retrieved. You can check this page with no versionNumber (http://localhost:5173/issue/maestro%3Afee96dbe5cc38db1f2f98c76e813cc162d91f232) and you'll see through the network tab that the tags are still being returned by the backend.

dashboard/src/components/IssueDetails/IssueDetails.tsx
```diff
@@ -120,6 +122,18 @@ export const IssueDetails = ({
...
+      return (
+        <div className="flex gap-3">
+          {data?.extra?.[issueId]?.versions[versionNumber].tags?.map(tag => (
+            <BranchBadge key={tag} tag={tag} />
```
MarceloRobert, line 130: IMO, there could be a greater font-size for the tags in the subtitle, maybe pass a `className` prop to BranchBadge and see if it looks ok. Currently we are using `text-xs` for all badges but here I don't think there's a need for it to be that small.

murilx: I uploaded a version where I'm using `text-sm` (14px instead of 12px) to the badge, the same size as the eyebrow. Tell me if you like it. IMO, I'd say I prefer the 12px more

backend/kernelCI_app/views/issueDetailsView.py
```diff
@@ -61,11 +61,17 @@ def get(self, _request, issue_id: Optional[str]) -> Response:
                 )
             version = version_row["version"]
 
-        issue_data = self._fetch_issue(issue_id=parsed_params.issue_id, version=version)
-        process_issues_extra_details(
-            issue_key_list=[(issue_id, version)],
-            processed_issues_table=self.processed_issue_extras,
-        )
+        try:
+            version = int(version)
```
WilsonNet, line 65: version should be a pydantic validation maybe add to IssueDetailsPathParameters and rename to `IssueDetailsRequestParameters`

backend/kernelCI_app/views/issueDetailsView.py
```diff
@@ -61,11 +61,17 @@ def get(self, _request, issue_id: Optional[str]) -> Response:
...
+            issue_data = self._fetch_issue(issue_id=parsed_params.issue_id, version=version)
+            process_issues_extra_details(
+                issue_key_list=[(issue_id, version)],
+                processed_issues_table=self.processed_issue_extras,
+            )
+        except ValueError:
+            return create_api_error_response(
+                error_message="The 'version' parameter must be an integer",
+            )
```
WilsonNet, line 74: won't need this since it will be inside an pydantic validation

backend/kernelCI_app/views/issueDetailsView.py
```diff
@@ -61,11 +61,17 @@ def get(self, _request, issue_id: Optional[str]) -> Response:
                 )
             version = version_row["version"]
 
-        issue_data = self._fetch_issue(issue_id=parsed_params.issue_id, version=version)
-        process_issues_extra_details(
-            issue_key_list=[(issue_id, version)],
-            processed_issues_table=self.processed_issue_extras,
-        )
+        try:
+            valid_request = IssueDetailsQueryParameters(version=version)
```
WilsonNet, line 65: you can move this validation to the top of the endpoint, that way will only need one try catch

murilx: This is here because I didn't feel safe of adding a query call to the pydantic model (if the query is `None` than it should call the query to get the version) but I can do that

WilsonNet (approved):
looks good

PR #958, opened 2025-02-18
fix: update new issue icon

### LINKED ISSUE #941: Update new label indication. Use a small circle dot

murilx:
Closes #941 

Based on this generated code: https://css-generators.com/starburst-shape/

**Important:** This solution uses [clip-path](https://developer.mozilla.org/en-US/docs/Web/CSS/clip-path) if there's any need to support older browsers, than the implementation might need to change

## Visual Reference
image

### DIFF
--- a/dashboard/src/components/Cards/IssuesList.tsx
+++ b/dashboard/src/components/Cards/IssuesList.tsx
@@ -132,9 +132,7 @@ const IssuesList = ({
           tagPills?.unshift(
             <Tooltip>
               <TooltipTrigger className="cursor-default">
-                <Badge variant="greenTag">
-                  {formatMessage({ id: 'global.new' })}
-                </Badge>
+                <div className="starburst bg-red aspect-square w-[24px]"></div>
               </TooltipTrigger>
               <TooltipContent>
                 <FormattedMessage id="issue.newIssue" />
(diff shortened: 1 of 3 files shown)


dashboard/src/components/Cards/IssuesList.tsx
```diff
@@ -132,9 +132,7 @@ const IssuesList = ({
           tagPills?.unshift(
             <Tooltip>
               <TooltipTrigger className="cursor-default">
-                <Badge variant="greenTag">
-                  {formatMessage({ id: 'global.new' })}
-                </Badge>
+                <div className="starburst bg-red aspect-square w-[20px]"></div>
```
MarceloRobert, line 135: nit to make it even more visible
```suggestion
                <div className="starburst bg-red aspect-square w-[24px]"></div>
```

MarceloRobert (approved):
LGTM

dashboard/src/index.css
```diff
@@ -4,6 +4,19 @@
 
 @custom-variant dark (&:is(.dark *));
 
+@layer utilities {
```
WilsonNet, line 7:
```suggestion
@layer components {
```
this is not an utility

WilsonNet (approved):
looks good, a minor fix, pre approving

PR #961, opened 2025-02-18
Add single letter url pattern for builds, tests and issues details

padovan:
For example, we want to access a build just by:

    d.kernelci.org/b/<build_id>


dashboard/src/routeTree.gen.ts
```diff
@@ -181,11 +158,14 @@ const HardwareHardwareIdBootIndexRoute =
...
   } as any)
 
+<<<<<<< HEAD
 const alternativesTTreeIdIndexRoute = alternativesTTreeIdIndexImport.update({
   path: '/',
   getParentRoute: () => alternativesTTreeIdRouteRoute,
 } as any)
 
+=======
+>>>>>>> parent of 694e806 (feat(route): add alternative route for /tree)
```
murilx, line 168: There are "merge noise" throughout this whole file. Since this is automatically generated you probably can get away with it by removing the file and recreating it (running the project with `pnpm dev` is sufficient to create this file). I can do this on my end considering you don't have the project setup yet

murilx (commented):
It seems that the routes removed from `dashboard/src/routes/(alternatives)` are necessary for the frontend pages. Without these files, pages like `d,kernelci.org/t/<$treeId>` aren't acessible.

If the goal is to add these shortcuts/alternatives only to the API endpoints than just adding these to `urls.py` is sufficient, but for the dashboard itself the other excluded files are also required.

Francisco2002:
This PR are closed by #971

Issue #938, opened 2025-02-17
Create Issue Listing Page


joaobertacchi:

MarceloRobert:
Link just for reference: https://grafana.kernelci.org/d/febo66mpd9atcf/new-issues

MarceloRobert:
Bring issues from the last x days and show the tree/branch of the first incident of that issue

PR #965, opened 2025-02-19
Feat: issue listing backend

### LINKED ISSUE #963: IssueListing: Backend

MarceloRobert:
Splits the previous issuesView.py into buildIssuesView and testIssuesView so it doesn't get confused or conflicting with the new issueView
Adds the issueView endpoint for the issueListing, retrieving the issues and their first incident for all issues on a specific origin with a limit of x days ago

## How to test
This PR only affects the backend, you can test the backend by running the `issues-listing-get.sh` script
You can check the queries by using the `export DEBUG_SQL_QUERY=True` command

Part of #938
Closes #963


backend/kernelCI_app/helpers/errorHandling.py
```diff
@@ -14,6 +14,9 @@ def __init__(self, message, status_code=400):
     def getJsonResponse(self):
         return self.json_response
 
+    def getRestResponse(self):
+        return Response(data={"error": self.message}, status=self.status_code)
+
```
Francisco2002, line 19: We can use this insteand of `getJsonResponse`

Francisco2002 (approved):
LGTM

backend/kernelCI_app/helpers/errorHandling.py
MarceloRobert: I believe some endpoints are still using `getJsonResponse`

backend/kernelCI_app/views/issueView.py
```diff
@@ -0,0 +1,95 @@
...
+        ) in self.processed_extra_issue_details.items():
+            self.first_incidents[issue_extras_id] = issue_extras_data["first_incident"]
+
+    @extend_schema(
+        parameters=[ListingQueryParameters],
+        responses=IssueListingResponse,
+        methods=["GET"],
+    )
+    def get(self, _request) -> Response:
+        origin_param = _request.GET.get("origin", DEFAULT_ORIGIN)
```
murilx, line 58: This line makes `origin` a optional parameter for issueView. If that's the case for every other view that uses the `ListingQueryParameters`, please add this to the model. It would be something like

```python
     origin: str = DEFAULT_MODEL
```

backend/kernelCI_app/views/issueView.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            interval_in_days = parseIntervalInDaysGetParameter(
+                _request.GET.get("intervalInDays", 7)
```
murilx, line 61: This line makes `intervalInDays` a optional parameter for issueView. If that's the case for every other view that uses the `ListingQueryParameters`, please add this to the model. It would be something like

```python
     intervalInDays: int = 7
```

If you are able to add both of these modifications (this one and the one from the other comment) to `ListingQueryParameters` or create another model for the query parameters of the issueView with these features, you'd be able to get both of these, validated, using the model itself. But consider if this change is worth it

backend/kernelCI_app/typeModels/commonListing.py
```diff
@@ -0,0 +1,15 @@
...
+
+class ListingQueryParameters(BaseModel):
+    origin: Annotated[
+        str, BeforeValidator(lambda o: DEFAULT_ORIGIN if o is None else o)
+    ]
+    interval_in_days: Annotated[
+        int,
+        Field(gt=0),
+        BeforeValidator(lambda i: DEFAULT_INTERVAL_IN_DAYS if i is None else i),
+    ]
```
murilx, line 15: nit: I believe that if you add `default` to Field, even though it does not do what we though it would, the values will show in the documentation

backend/kernelCI_app/typeModels/hardwareListing.py
```diff
@@ -28,12 +28,16 @@ class HardwareResponse(BaseModel):
 # documentation purposes. This model is not used in the code.
 # TODO Remove timestamp from the api and this model
 class HardwareQueryParamsDocumentationOnly(BaseModel):
-    origin: str = DEFAULT_ORIGIN
+    origin: Annotated[
+        str, BeforeValidator(lambda o: DEFAULT_ORIGIN if o is None else o)
+    ]
```
murilx, line 33: nit: No need to change this since it's only for documentation purposes

murilx (approved):
Working very well!

PR #989, opened 2025-02-20
feat/python lint

WilsonNet:
- **feat(hardware-listing-backend): optimize hardware view query Refactor the hardware view to use raw SQL queries for fetching test data, replacing the ORM-based approach. This change improves performance by reducing the number of database queries and leveraging SQL's capabilities for complex data aggregation.**
- **feat: add black formatting check to CI pipeline**


backend/pre-commit
```diff
@@ -1,3 +1,17 @@
...
+fi
+
 poetry run flake8
+FLAKE8_STATUS=$?
+
+if [ $FLAKE8_STATUS -ne 0 ]; then
+  exit 1
+fi
+
+exit 0
```
murilx, line 17: nit: AFAIK pre-commit will block commit if the exit status is anything other than 0, if that's true we can do

```suggestion
poetry run black --check .
BLACK_STATUS=$?

poetry run flake8
FLAKE8_STATUS=$?

exit $BLACK_STATUS || $FLAKE8_STATUS
```

murilx (approved):
Worked well on my tests. Very glad we added this to CI!

backend/pre-commit
WilsonNet: that works but you will get to wait for flake8 to perform even if black already failed, having separated ifs allow for an early exit

MarceloRobert (approved):
It's still working on my tests

PR #991, opened 2025-02-20
Feat: issue listing frontend

MarceloRobert:


dashboard/src/components/Issue/IssueCulprit.tsx
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return result;
+  }, [culprit_code, culprit_harness, culprit_tool, formatMessage]);
+
+  if (culprits.length === 0) {
+    return <span>-</span>;
+  }
+
+  return <span>{culprits.join(', ')}</span>;
```
Francisco2002, line 39: You can use `valueOrEmpty` function here.

dashboard/src/components/IssueTable/IssueTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+  },
+];
+
+interface IIssueTable {
+  issueListing?: IssueListingResponse;
+}
+
+export const IssueTable = ({ issueListing }: IIssueTable): JSX.Element => {
+  const { listingSize } = useSearch({ strict: false });
+  const navigate = useNavigate({ from: '/hardware' });
```
Francisco2002, line 166:
```suggestion
  const navigate = useNavigate({ from: '/issue' });
```

dashboard/src/lib/date.ts
```diff
@@ -0,0 +1,14 @@
+import { differenceInDays } from 'date-fns';
+
+const RELATIVE_DAYS_EDGE = 3;
+
+export const shouldShowRelativeDate = (
+  date: string,
+  limitInDays?: number,
```
Francisco2002, line 7:
```suggestion
export const shouldShowRelativeDate = (
  date: string,
  limitInDays?: number = RELATIVE_DAYS_EDGE,
```

dashboard/src/pages/IssueListing/IssueListing.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { IssueListingPage } from './IssueListingPage';
+
+const IssueListing = (): JSX.Element => {
+  const { issueSearch: unsafeIssueSearch } = useSearch({
+    strict: false,
+  });
+
+  const issueSearch = z.string().catch('').parse(unsafeIssueSearch);
+
+  const navigate = useNavigate({ from: '/' });
```
Francisco2002, line 21:
```suggestion
  const navigate = useNavigate({ from: '/issue' });
```

dashboard/src/routes/issue/index.tsx
Francisco2002, line 1: don't forget to add a alternative route to this route

dashboard/src/routes/issue/route.tsx
```diff
@@ -0,0 +1,30 @@
...
+  intervalInDays: DEFAULT_TIME_SEARCH,
+  issueSearch: '',
+  listingSize: DEFAULT_LISTING_ITEMS,
+};
+
+const zIssueSearchSchema = z.object({
+  intervalInDays: makeZIntervalInDays(DEFAULT_TIME_SEARCH),
+  issueSearch: z.string().catch(''),
+  listingSize: zListingSize,
+} satisfies SearchSchema);
```
Francisco2002, line 25:
```suggestion
const issuesListingDefaultValues = {
  intervalInDays: DEFAULT_TIME_SEARCH,
  issueSearch: '',
  listingSize: DEFAULT_LISTING_ITEMS,
};

const issuesListingSearchSchema = z.object({
  intervalInDays: makeZIntervalInDays(DEFAULT_TIME_SEARCH),
  issueSearch: z.string().catch(''),
  listingSize: zListingSize,
} satisfies SearchSchema);
```

to maintain consistent with other routes files (see build or test route files).

dashboard/src/routes/issue/index.tsx
murilx: Since `/tree` and `/hardware` don't have alternative routes, I'd say it might not be necessary to add one to `/issue` for now. Although, we could future-proof this by adding one to `/issue` now. I'm not sure...

dashboard/src/routes/issue/route.tsx
murilx: I'd say we only change these variable names if we are exporting them, otherwise there's no need. (All other routes besides build or test have these named as `defaultValues` and `z<name>SearchSchema

dashboard/src/lib/date.ts
murilx: I think it doesn't need `?` if you are passing a default value but I agree

dashboard/src/routes/issue/index.tsx
MarceloRobert: It's true that tree and hardware don't have alternatives, but neither does treeDetails and hardwareDetails, but issueDetails does have it. So I'll add it just for completion of the issue routes

dashboard/src/routes/issue/route.tsx
MarceloRobert: Since I'll make the shortcut route, I'll export them, so I will rename the variables too

dashboard/src/routes/issue/index.tsx
Francisco2002: And `/i` is a shortcut to `/issue`. Since that I can search by `/i/$issueId`, I would expect that exists a `/i` route.

dashboard/src/lib/date.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const shouldShowRelativeDate = (
+  date: string,
+  limitInDays: number = RELATIVE_DAYS_EDGE,
+): boolean => {
+  const treeDate = new Date(date);
+  const currentDate = new Date();
+  const days = differenceInDays(currentDate, treeDate);
+
+  return days <= limitInDays;
```
Francisco2002, line 13:
```suggestion
  return days < limitInDays;
```

some rows are showing the label `4 days ago` (we define 3 days as maximum)

Francisco2002 (approved):
Nice job! Just a nit

dashboard/src/components/IssueTable/IssueTable.tsx
```diff
@@ -0,0 +1,288 @@
...
+    },
+    header: ({ column }): JSX.Element => (
+      <TableHeader
+        column={column}
+        intlKey="global.treeBranch"
+        tooltipId="issueListing.treeBranchTooltip"
+      />
+    ),
+  },
+];
```
murilx, line 158: I'd add some indication that clicking this column goes to tree details instead of issue details. Maybe make the text underlined or add an (i) icon, or maybe even better: add this information in the tooltip

murilx (commented):
Besides these comments I'd still think in the future it would be good to add the tags and number of incidents to this page. But since it's not a necessity or demand right now, It's working great, there's just the question of Log Excerpt showing two times as well

murilx (approved):
Considering the duplication of log excerpt is only happening in local, this looks good to me

dashboard/src/components/IssueTable/IssueTable.tsx
MarceloRobert: I just added the information to the tooltip

PR #1000, opened 2025-02-21
refactor: apply PEP8 naming conventions

### LINKED ISSUE #493: Improve python lint rules
- Try add a rule that force the PEP8 naming rules, for example, variables should have snake_case names
- Add a rule that enforce the usage of single quotes (')

WilsonNet:
- Renamed functions and variables to follow PEP8 naming conventions.
- Updated imports to match the new function names.
- Removed unused ExceptionWithJsonResponse class.
- Added pep8-naming to the dependencies in pyproject.toml.
- Updated poetry.lock to reflect the new dependency.

Closes #493

### DIFF
--- a/backend/kernelCI_app/helpers/filters.py
+++ b/backend/kernelCI_app/helpers/filters.py
@@ -4,7 +4,7 @@ import re
 from kernelCI_app.constants.general import UNCATEGORIZED_STRING
 from kernelCI_app.helpers.commonDetails import PossibleTabs
 from kernelCI_app.typeModels.databases import PASS_STATUS, failure_status_list
-from kernelCI_app.utils import getErrorResponseBody
+from kernelCI_app.utils import get_error_body_response
 from kernelCI_app.constants.general import UNKNOWN_STRING
 
 NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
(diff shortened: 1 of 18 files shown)


backend/kernelCI_app/helpers/filters.py
```diff
@@ -404,15 +404,15 @@ def _handle_platforms(self, current_filter: ParsedFilter) -> None:
...
         try:
             for current_filter in self.filters:
                 field = current_filter["field"]
                 # Delegate to the appropriate handler based on the field
                 if field in self.filter_handlers:
                     self.filter_handlers[field](current_filter)
-        except InvalidComparisonOP as e:
-            return HttpResponseBadRequest(getErrorResponseBody(str(e)))
+        except InvalidComparisonOPError as e:
+            return HttpResponseBadRequest(get_error_body_response(str(e)))
```
murilx, line 415: Do we use this response structure actively? If not, I'd argue it's safe to make this change
```suggestion
            return create_error_response(error_message=str(e), status_code=HTTPStatus.BAD_REQUEST)
```

or using `create_api_error_response`. Maybe not for this PR, but definitely something we could do

murilx:
Didn't know our classes attributes were following PEP8 to be honest. That's neat

murilx (approved):
Seems to be working fine. Since this point might not be added to this PR but left to another I'm pre-approving

MarceloRobert (approved):
It would be ideal to rename the responses too, so that we don't have to ignore the errors there, but I know that that will be a high effort task, so let's leave if for another time. LGTM.

Issue #974, opened 2025-02-20
Unit tests for Tests


murilx:
- [x] testDetailsView
- [ ] testIssuesView

PR #1004, opened 2025-02-24
Test: test details and issues endpoints

### LINKED ISSUE #973: Unit tests for Issues
- [x] IssueDetailsBuildsView
- [x] IssueDetailsTestsView
- [x] IssueDetailsView
- [x] IssueExtrasView
- [x] IssueView

Francisco2002:
- refactor default tests asserts
- create test details tests
- create issue view, issue details, issue test, issue build tests

Close #973, #974

### DIFF
--- a/backend/kernelCI_app/unitTests/issues_test.py
+++ b/backend/kernelCI_app/unitTests/issues_test.py
@@ -2,7 +2,14 @@ from kernelCI_app.unitTests.utils.healthCheck import online
 from kernelCI_app.unitTests.utils.issueClient import IssueClient
 from kernelCI_app.unitTests.utils.asserts import (
     assert_status_code_and_error_response,
+    assert_has_fields_in_response_content,
 )
+from kernelCI_app.unitTests.utils.fields.issues import (
+    issues_expected_fields,
+    issues_listing_fields,
+)
+from kernelCI_app.unitTests.utils.fields.tests import test_expected_fields
+from kernelCI_app.unitTests.utils.fields.builds import build_expected_fields
 from kernelCI_app.utils import string_to_json
 import pytest
 from http import HTTPStatus
@@ -35,9 +42,8 @@ def test_list(origin, interval_in_day, status_code, has_error_body):
             fields=issues_listing_fields, response_content=content
         )
 
-    if "issues" in content:
         assert_has_fields_in_response_content(
-            fields=issues_fields, response_content=content["issues"][0]
+            fields=issues_expected_fields, response_content=content["issues"][0]
         )
 
 
@@ -64,6 +70,11 @@ def test_details(issue_id, issue_version, status_code, has_error_body):
         status_code=status_code,
         should_error=has_error_body,
     )
+
+    if not has_error_body:
+        assert_has_fields_in_response_content(
+            fields=issues_expected_fields, response_content=content
+        )
 
 
 @online
@@ -101,6 +112,11 @@ def test_extra_details(issues_list, status_code, has_error_body):
         should_error=has_error_body,
     )
 
+    if not has_error_body:
+        assert_has_fields_in_response_content(
+            fields=["issues"], response_content=content
+        )
+
 
 @online
 @pytest.mark.parametrize(
@@ -123,7 +139,7 @@ def test_issue_tests(issue_id, issue_version, status_code, has_error_body):
 
     if not has_error_body:
         assert_has_fields_in_response_content(
-            fields=test_fields, response_content=content[0]
+            fields=test_expected_fields, response_content=content[0]
         )
 
 
@@ -153,5 +169,5 @@ def test_issue_builds(issue_id, issue_version, status_code, has_error_body):
 
     if not has_error_body:
         assert_has_fields_in_response_content(
-            fields=build_field, response_content=content[0]
+            fields=build_expected_fields, response_content=content[0]
         )
--- a/backend/kernelCI_app/unitTests/issues_test.py
+++ b/backend/kernelCI_app/unitTests/issues_test.py
@@ -1,0 +1,157 @@
+from kernelCI_app.unitTests.utils.healthCheck import online
+from kernelCI_app.unitTests.utils.issueClient import IssueClient
+from kernelCI_app.unitTests.utils.asserts import (
+    assert_status_code_and_error_response,
+)
+from kernelCI_app.utils import string_to_json
+import pytest
+from http import HTTPStatus
+
+
+client = IssueClient()
+
+
+@online
+@pytest.mark.parametrize(
+    "origin, interval_in_day, status_code, has_error_body",
+    [
+        ("maestro", 7, HTTPStatus.OK, False),
+        ("maestro", -5, HTTPStatus.BAD_REQUEST, True),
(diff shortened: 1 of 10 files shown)


backend/kernelCI_app/unitTests/utils/issueClient.py
```diff
@@ -0,0 +1,43 @@
...
+import json
+from kernelCI_app.unitTests.utils.baseClient import BaseClient
+
+"""
+    IssueDetailsBuildsView
+    IssueDetailsTestsView
+    IssueDetailsView
+    IssueExtrasView
+    IssueView
+"""
```
murilx, line 12: I think that you accidentally left this comment on the code

backend/kernelCI_app/unitTests/utils/issueClient.py
```diff
@@ -0,0 +1,43 @@
...
+class IssueClient(BaseClient):
+    def get_issues_list(self, *, origin: str | None, interval_in_days: int | None):
+        path = reverse("issue")
+        query = {"origin": origin, "intervalInDays": interval_in_days}
+        url = self.get_endpoint(path=path, query=query)
+        return requests.get(url)
+
+    def get_issues_details(self, *, issue_id, issue_version):
+        path = reverse("issueDetails", kwargs={"issue_id": issue_id})
+        query = None if issue_version is None else {"version": issue_version}
```
murilx, line 24: IssueDetailsView should be able to handle `{"version": None}` so this conditional shouldn't be necessary

backend/kernelCI_app/unitTests/utils/defaultTestAsserts.py
```diff
@@ -0,0 +1,10 @@
+from kernelCI_app.utils import string_to_json
+
+
+def defaultTestAsserts(response, status_code, has_error_body):
```
murilx, line 4: nit: Not every test will be doing these assertions so I'd call this function something different

backend/kernelCI_app/unitTests/utils/defaultTestAsserts.py
```diff
@@ -0,0 +1,10 @@
...
+    content = string_to_json(response.content.decode())
+    assert (
+        response.status_code == status_code
+    ), f"This call give status code {response.status_code}, when it should be {status_code}."
+    if has_error_body:
+        assert "error" in content, "This call must return a error message."
```
murilx, line 10: Really like the messages

backend/kernelCI_app/unitTests/issues_test.py
murilx, line 1: Besides the "default" asserts, I think it's also recommended to assert if every excepted response field is present, at least the ones "at the top". For example, for issue details we could add

```python
assert "field_timestamp" in content
assert "id" in content
assert "version" in content
...
```

Do this at least for the more broad tests, or for tests that does not interact with these fields in any way. I believe that in your case this would be true for everyone of them, since the views being tested are simpler

backend/kernelCI_app/unitTests/utils/defaultTestAsserts.py
```diff
@@ -0,0 +1,10 @@
+from kernelCI_app.utils import string_to_json
+
+
+def defaultTestAsserts(response, status_code, has_error_body):
+    content = string_to_json(response.content.decode())
```
murilx, line 5: I think this would be better left outside the function, so that if we have more than one assertion function to pass in a single test, we just convert one time instead of having to convert inside everyfunction

backend/kernelCI_app/unitTests/issues_test.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+
+@online
+@pytest.mark.parametrize(
+    "issue_id, issue_version, status_code, has_error_body",
+    [
+        ("maestro:ae160f6f27192c3527b2e88faba35d85d27c285f", 1, 200, False),
+        ("maestro:b9856de6a9d2099f438c8946f3ba192e046bda35", 1, 200, True),
+        ("maestro:b9856de6a9d2099f438c8946f3ba192e046bda35", None, 400, True),
```
MarceloRobert, line 121: Again, even without passing the issue version, it is returning status 200 for
`http 'http://localhost:8000/api/issue/maestro:b9856de6a9d2099f438c8946f3ba192e046bda35/tests'`

backend/kernelCI_app/unitTests/issues_test.py
```diff
@@ -0,0 +1,162 @@
...
+        )
+
+
+@online
+@pytest.mark.parametrize(
+    "issue_id, issue_version, status_code, has_error_body",
+    [
+        ("maestro:2ff8fe94f6d53f39321d4a37fe15801cedc93573", 1, 200, False),
+        ("maestro:2ff8fe94f6d53f39321d4a37fe15801cedc93573", 0, 200, True),
+        ("maestro:2ff8fe94f6d53f39321d4a37fe15801cedc93573", None, 400, True),
```
MarceloRobert, line 146: ditto about getting status 200 with version None

MarceloRobert (commented):
You're missing the tests for the testIssues endpoint

backend/kernelCI_app/unitTests/issues_test.py
```diff
@@ -0,0 +1,162 @@
...
+    response = client.get_issue_builds(issue_id=issue_id, issue_version=issue_version)
+    content = string_to_json(response.content.decode())
+    assert_status_code_and_error_response(
+        response=response,
+        content=content,
+        status_code=status_code,
+        should_error=has_error_body,
+    )
+
+    if not has_error_body and len(content) > 0:
```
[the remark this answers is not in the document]
Francisco2002: You are right! I will correct this.

backend/kernelCI_app/unitTests/utils/asserts.py
murilx, line 1: add typing in this file. I believe `response: requests.Response` and `content: dict` might be sufficient. Also, every function here returns `None`

backend/kernelCI_app/unitTests/utils/fields/builds.py
murilx, line 1: nit: would change the name of this directory to `expectedFields`

backend/kernelCI_app/unitTests/issues_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+    if not has_error_body:
+        assert_has_fields_in_response_content(
+            fields=issues_listing_fields, response_content=content
+        )
+
+    if "issues" in content:
+        assert_has_fields_in_response_content(
+            fields=issues_fields, response_content=content["issues"][0]
+        )
```
murilx, line 46: As far as I understand this view correctly, every time it does not have an error it should have the "issues" field. In this case, add this 

```python
        assert_has_fields_in_response_content(
            fields=issues_fields, response_content=content["issues"][0]
        )
```
inside `if not has_error_body` and remove this check `if "issues" in content` since the other `assert_has_fields_in_response_content` should guarantee this is true

murilx (commented):
It seems to be working really fine! Just some last comments before I can approve this

murilx (approved):
Working really well

MarceloRobert (approved):
Looks really good, well done

PR #1011, opened 2025-02-27
Feat: add test status history to test details

### LINKED ISSUE #258: Show last time a test passed (in Test Details Page)
We need more info here. For a specific build? @padovan

- [x] #898
- [x] #899

### LINKED ISSUE #898: [Test Details Page - FE] Show last time a test passed (if data is not available, does not show the field)

### LINKED ISSUE #899: [Test Details Page - backend] Show last time a test passed

MarceloRobert:
Adds a new query to get the history of a test
Adds a method to tell if a given test is a regression, is fixed, or is unstable

## How to test
Fetch test endpoints with `http 'localhost:8000/api/test/<test_id>` and check if the history is getting fetched and the classification is correct. You can also check the network tab when going to a test details page to check the response.

For the frontend, go to any testDetails page and check the Status History section, some examples:
[Test with 1 pass](http://localhost:5173/test/maestro%3A67c173ff3218f15c74bc5c79)
[Test with 1 fail](http://localhost:5173/test/maestro%3A67c074b57d9799de2b4ce46d)
[Test with 1 inconclusive](http://localhost:5173/test/maestro%3A67c17c8b3218f15c74bc68bd)
[Unstable test](http://localhost:5173/test/maestro%3A678ee7b046f65f378a18d33e)
[Test with many passes](http://localhost:5173/test/maestro%3A67c0c84f7d9799de2b4d805b)

Closes #898
Closes #899
Closes #258

### DIFF
--- a/backend/kernelCI_app/queries/test.py
+++ b/backend/kernelCI_app/queries/test.py
@@ -1,0 +1,61 @@
+import datetime
+from kernelCI_app.models import Tests
+
+
+def get_test_details_data(*, test_id):
+    return (
+        Tests.objects.filter(id=test_id)
+        .values(
+            "id",
+            "build_id",
+            "status",
+            "path",
+            "log_excerpt",
+            "log_url",
+            "misc",
+            "environment_misc",
+            "start_time",
+            "environment_compatible",
+            "output_files",
+            "field_timestamp",
+            "build__compiler",
+            "build__architecture",
+            "build__config_name",
+            "build__checkout__git_commit_hash",
+            "build__checkout__git_repository_branch",
+            "build__checkout__git_repository_url",
+            "build__checkout__git_commit_tags",
+            "build__checkout__tree_name",
+            "build__checkout__origin",
+        )
+        .first()
+    )
+
+
+# TODO: combine with the test_details query
+def get_test_status_history(
+    *,
+    path: str,
+    origin: str,
+    git_repository_url: str,
+    git_repository_branch: str,
+    platform: str,
+    current_test_timestamp: datetime,
+):
+    return (
+        Tests.objects.values(
+            "field_timestamp",
+            "id",
+            "status",
+            "build__checkout__git_commit_hash",
+        )
+        .filter(
+            path=path,
+            build__checkout__origin=origin,
+            build__checkout__git_repository_url=git_repository_url,
+            build__checkout__git_repository_branch=git_repository_branch,
+            environment_misc__platform=platform,
+            field_timestamp__lte=current_test_timestamp,
+        )
+        .order_by("-field_timestamp")[:10]
+    )
--- a/backend/kernelCI_app/views/testDetailsView.py
+++ b/backend/kernelCI_app/views/testDetailsView.py
@@ -1,7 +1,11 @@
 from http import HTTPStatus
 from kernelCI_app.helpers.errorHandling import create_api_error_response
-from kernelCI_app.models import Tests
-from kernelCI_app.typeModels.testDetails import TestDetailsResponse
+from kernelCI_app.queries.test import get_test_details_data, get_test_status_history
+from kernelCI_app.typeModels.databases import FAIL_STATUS, PASS_STATUS
+from kernelCI_app.typeModels.testDetails import (
+    PossibleRegressionType,
+    TestDetailsResponse,
+)
 from drf_spectacular.utils import extend_schema
 from rest_framework.views import APIView
 from rest_framework.response import Response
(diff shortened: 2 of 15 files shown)


backend/kernelCI_app/views/testDetailsView.py
```diff
@@ -1,49 +1,86 @@
...
+                elif first_test_status == FAIL_STATUS:
+                    history_task = "regression"
+                    starting_status = FAIL_STATUS
+                    opposite_status = PASS_STATUS
+                else:
+                    return "unstable"
+                first_test_flag = False
+                continue
+
+            test_status = test["status"]
```
murilx, line 40: move `test_status` to before the first if and then you can remove the `first_test_status` variable

backend/kernelCI_app/queries/test.py
```diff
@@ -0,0 +1,60 @@
...
+    git_repository_branch: str,
+    platform: str,
+    current_test_timestamp: datetime,
+):
+    return (
+        Tests.objects.values(
+            "field_timestamp",
+            "id",
+            "status",
+            "build__checkout__git_commit_hash",
```
murilx, line 49: nit: It's still a little blurry to me the reason of selecting this field, but I'm OK with keeping it if the others agree on it

backend/kernelCI_app/queries/test.py
```diff
@@ -0,0 +1,60 @@
...
+            "build__checkout__git_repository_url",
+            "build__checkout__git_commit_tags",
+            "build__checkout__tree_name",
+            "build__checkout__origin",
+        )
+        .first()
+    )
+
+
+def get_test_status_history(
```
WilsonNet, line 35: we can merge this query, add a #TODO for this

backend/kernelCI_app/views/testDetailsView.py
```diff
@@ -1,49 +1,86 @@
...
+    def process_test_status_history(
+        self, *, status_history: list[dict]
+    ) -> PossibleRegressionType:
+        history_task: PossibleRegressionType
+        first_test_flag = True
+        status_changed = False
+
+        for test in status_history:
+            if first_test_flag:
+                first_test_status = status_history[0]["status"]
```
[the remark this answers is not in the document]
Francisco2002: nit: `latest_test_status` is a better name to this variable. The first test is the last in the list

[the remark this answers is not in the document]
MarceloRobert: With the change pointed out by @murilx this won't be needed

backend/kernelCI_app/views/testDetailsView.py
```diff
@@ -1,49 +1,85 @@
...
+        )
+
         try:
-            valid_response = TestDetailsResponse(**response)
+            valid_response = TestDetailsResponse(
+                **response,
+                status_history=[
+                    TestStatusHistoryItem(**history_item)
+                    for history_item in status_history_response
+                ],
```
murilx, line 80: Can't pydantic automatically do this "conversion"?

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -134,6 +147,49 @@ const TestDetailsSections = ({
...
+          tooltipValues={{ amount: test.status_history.length }}
+          iconClassName="size-5"
+          contentClassName="whitespace-pre-line"
+        />
+      ),
+      subtitle: (
+        <div>
+          <Badge variant="blueTag">
+            {capitalizeString(test.regression_type)}
+          </Badge>
```
murilx, line 165: Maybe it's worth adding a tooltip saying something like "regression type" just so that we can make it more clear if anyone finds this badge confusing

backend/kernelCI_app/views/testDetailsView.py
MarceloRobert: I think I was using some other variable here, but now it makes sense to use the `status_history_response` directly

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -134,6 +149,56 @@ const TestDetailsSections = ({
...
+          iconClassName="size-5"
+          contentClassName="whitespace-pre-line"
+        />
+      ),
+      subtitle: (
+        <div>
+          <Tooltip>
+            <TooltipTrigger>
+              <Badge variant="blueTag">
+                {capitalizeString(test.regression_type)}
```
murilx, line 168: You could possibly achieve the same effect just using [CSS text-transform](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform). If you don't need the actual string to be capitalized for some reason (since this only change the way its shown to the user and not the string itself), than I believe this is better than creating a function for this

```suggestion
              <Badge className="capitalize" variant="blueTag">
                {test.regression_type}
```

dashboard/src/lib/string.ts
```diff
@@ -75,3 +75,10 @@ export const includesInAnStringOrStringArray = (
   }
   return matchesRegexOrIncludes(searched, inputFilter);
 };
+
+export const capitalizeString = (text: string): string => {
+  if (!text) {
+    return '';
+  }
+  return text.charAt(0).toUpperCase() + text.slice(1);
+};
```
murilx, line 84: May be unnecessary if you only want the final text to be capitalized. Discussed more about this in the other comment

murilx (commented):
Seems to be working really well, just a comment about the `capitalizeString` function

dashboard/src/components/TestDetails/TestDetails.tsx
MarceloRobert: I was using the regression_type as a string in the field value, so that's why I made the capitalize function, but now that it is in a component I can use the css class, you're right

murilx (approved):
LGTM

WilsonNet:
There are single test satus with "regression" label
image

backend/kernelCI_app/views/testDetailsView.py
```diff
@@ -1,49 +1,85 @@
...
+                    starting_status = FAIL_STATUS
+                    opposite_status = PASS_STATUS
+                else:
+                    return "unstable"
+                first_test_flag = False
+                continue
+
+            is_inconclusive = test_status != PASS_STATUS and test_status != FAIL_STATUS
+
+            if test_status == opposite_status:
```
WilsonNet, line 41: do you have an exmaple of test with `fixed` or `regression`? 
I don't think python has access to `opposite_status` in this scope

WilsonNet: a match case would probably make things more readable, something like

```
    match (history_status, current_status):
        case (HistoryStatus.Pass, Status.Failure):
            history_status = HistoryStatus.Regression
        case (HistoryStatus.Regression | HistoryStatus.Failure | HistoryStatus.Unstable | HistoryStatus.UnstablePass, Status.Pass):
            history_status = HistoryStatus.UnstablePass
        case (_, Status.Inconclusive):
            history_status = HistoryStatus.Unstable
        case _:
            pass
```

MarceloRobert: I couldn't find those types of tests, but if you look at http://localhost:5173/test/maestro%3A678e592546f65f378a15d3c1 you'll see that it is tagging as unstable even without inconclusive tests, meaning that the `status_changed` flag is being set

WilsonNet (approved):
worked on the test cases that I could find

PR #1016, opened 2025-02-27
Test: hardware details tab views

Francisco2002:
Create tests to the endpoints:

- Hardware Details Boots View
- Hardware Details Builds View
- Hardware Details Tests View

### DIFF
--- a/backend/kernelCI_app/unitTests/hardware_test.py
+++ b/backend/kernelCI_app/unitTests/hardware_test.py
@@ -1,0 +1,213 @@
+from kernelCI_app.unitTests.utils.healthCheck import online
+from kernelCI_app.unitTests.utils.hardwareClient import HardwareClient
+from kernelCI_app.unitTests.utils.asserts import assert_status_code_and_error_response
+from kernelCI_app.utils import string_to_json
+from http import HTTPStatus
+from kernelCI_app.typeModels.hardwareDetails import HardwareDetailsPostBody
+import copy
+
+
+client = HardwareClient()
+
+UNEXISTENT_HARDWARE_ID = {
+    "id": "no hardware id",
+    "body": HardwareDetailsPostBody(
+        origin="",
+        startTimestampInSeconds=1737487800,
+        endTimestampInSeconds=1737574200,
+        selectedCommits={},
+        filter={},
+    ),
+}
+
+BAD_REQUEST_REQUEST_BODY = {
+    "id": "google,juniper",
+    "body": HardwareDetailsPostBody(
+        origin="",
+        startTimestampInSeconds="",
+        endTimestampInSeconds="",
+        selectedCommits={},
+        filter={},
+    ),
+}
+
+GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS = {
+    "id": "google,juniper",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740227400,
+        endTimestampInSeconds=1740659400,
+        selectedCommits={},
+        filter={},
+    ),
+}
+
+ARM_JUNO_HARDWARE_WITHOUT_FILTERS = {
+    "id": "arm,juno",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740232800,
+        endTimestampInSeconds=1740664800,
+        selectedCommits={},
+        filter={},
+    ),
+}
+
+GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS = copy.deepcopy(
+    GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS
+)
+GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS["body"].filter = {
+    "filter_architecture": ["arm64"],
+    "filter_boot.status": ["PASS"],
+}
+
+HARDWARE_WITH_UNEXISTENT_FILTER_VALUE = copy.deepcopy(
+    GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS
+)
+HARDWARE_WITH_UNEXISTENT_FILTER_VALUE["body"].filter = {
+    "filter_architecture": ["invalid"]
+}
+
+HARDWARE_WITH_GLOBAL_FILTER = {
+    "id": "aaeon-UPN-EHLX4RE-A10-0864",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740241800,
+        endTimestampInSeconds=1740673800,
+        selectedCommits={},
+        filter={
+            "filter_architecture": ["i386"],
+            "filter_config_name": ["defconfig"],
+            "filter_compiler": ["gcc-12"],
+            "filter_valid": ["true"],
+        },
+    ),
+}
+
+HARDWARE_WITH_LOCAL_FILTER = {
+    "id": "google,kevin-rev15",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740243600,
+        endTimestampInSeconds=1740675600,
+        selectedCommits={},
+        filter={
+            "filter_boot.issue": ["maestro:e602fca280d85d8e603f7c0aff68363bb0cd7993"],
+            "filter_boot.status": ["PASS", "MISS"],
+            "filter_test.status": ["DONE", "FAIL"],
+            "filter_test.platform": ["rk3399-gru-kevin"],
+        },
+    ),
+}
+
+SUCCESS_EXPECTED_RESPONSE = {
+    "expected_status": HTTPStatus.OK,
+    "has_error": False,
+    "check_emptiness": False,
+}
+
+ERROR_EXPECTED_RESPONSE = {
+    "expected_status": HTTPStatus.OK,
+    "has_error": True,
+    "check_emptiness": False,
+}
+
+BAD_REQUEST_EXPECTED_RESPONSE = {
+    "expected_status": HTTPStatus.BAD_REQUEST,
+    "has_error": True,
+    "check_emptiness": False,
+}
+
+EMPTY_EXPECTED_RESPONSE = {
+    "expected_status": HTTPStatus.OK,
+    "has_error": False,
+    "check_emptiness": True,
+}
+
+
+def pytest_generate_tests(metafunc):
+    base_cases = [
+        ((BAD_REQUEST_REQUEST_BODY), BAD_REQUEST_EXPECTED_RESPONSE),
+        ((GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS), SUCCESS_EXPECTED_RESPONSE),
+        ((HARDWARE_WITH_UNEXISTENT_FILTER_VALUE), EMPTY_EXPECTED_RESPONSE),
+    ]
+    extra_cases = []
+    builds_and_boots_cases = [
+        ((GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS), SUCCESS_EXPECTED_RESPONSE)
+    ]
+
+    if metafunc.config.getoption("--run-all"):
+        extra_cases = [
+            ((ARM_JUNO_HARDWARE_WITHOUT_FILTERS), SUCCESS_EXPECTED_RESPONSE),
+            ((UNEXISTENT_HARDWARE_ID), ERROR_EXPECTED_RESPONSE),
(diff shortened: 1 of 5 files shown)


backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+            base_cases + extra_cases + google_cases,
+        )
+
+    if "test_hardware_tests" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "test_hardware_tests,expected_response", base_cases + extra_cases
+        )
+
+
+@online()
```
murilx, line 162: nit:
```suggestion
@online
```

backend/kernelCI_app/views/hardwareDetailsBootsView.py
```diff
@@ -130,6 +130,12 @@ def post(self, request, hardware_id) -> Response:
             end_datetime=self.end_datetime,
         )
 
+        if len(trees) == 0:
+            return Response(
+                data={"error": "This hardware isn't associated with any commit"},
+                status=HTTPStatus.OK,
+            )
```
murilx, line 137: use `create_api_error_response`

backend/kernelCI_app/views/hardwareDetailsBuildsView.py
```diff
@@ -123,6 +123,12 @@ def post(self, request, hardware_id):
             end_datetime=self.end_datetime,
         )
 
+        if len(trees) == 0:
+            return Response(
+                data={"error": "This hardware isn't associated with any commit"},
+                status=HTTPStatus.OK,
+            )
```
murilx, line 130: ditto

backend/kernelCI_app/views/hardwareDetailsTestsView.py
```diff
@@ -128,6 +128,12 @@ def post(self, request, hardware_id) -> Response:
             end_datetime=self.end_datetime,
         )
 
+        if len(trees) == 0:
+            return Response(
+                data={"error": "This hardware isn't associated with any commit"},
+                status=HTTPStatus.OK,
+            )
```
murilx, line 135: ditto

backend/kernelCI_app/views/hardwareDetailsBootsView.py
```diff
@@ -130,6 +131,14 @@ def post(self, request, hardware_id) -> Response:
             end_datetime=self.end_datetime,
         )
 
+        if len(trees) == 0:
+            return create_api_error_response(
+                error_message={
+                    "error": "This hardware isn't associated with any commit"
+                },
```
murilx, line 138: `create_api_error_response` already returns `{"error": error_message}`. You should pass only the message, not the whole object here
```suggestion
            return create_api_error_response(
                error_message="This hardware isn't associated with any commit",
```

backend/kernelCI_app/views/hardwareDetailsBuildsView.py
```diff
@@ -123,6 +124,14 @@ def post(self, request, hardware_id):
...
         )
 
+        if len(trees) == 0:
+            return create_api_error_response(
+                error_message={
+                    "error": "This hardware isn't associated with any commit"
+                },
+                status_code=HTTPStatus.OK,
+            )
+
```
murilx, line 134: ditto

backend/kernelCI_app/views/hardwareDetailsTestsView.py
```diff
@@ -128,6 +129,14 @@ def post(self, request, hardware_id) -> Response:
...
         )
 
+        if len(trees) == 0:
+            return create_api_error_response(
+                error_message={
+                    "error": "This hardware isn't associated with any commit"
+                },
+                status_code=HTTPStatus.OK,
+            )
+
```
murilx, line 139: ditto

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def pytest_generate_tests(metafunc):
+    base_cases = [
+        ((INVALID_REQUEST_BODY), BAD_REQUEST_EXPECTED_RESPONSE),
+        ((GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS), SUCESS_EXPECTED_RESPONSE),
+        ((HARDWARE_WITH_INVALID_FILTER), EMPTY_EXPECTED_RESPONSE),
+    ]
+    extra_cases = []
+    google_cases = [((GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS), SUCESS_EXPECTED_RESPONSE)]
```
MarceloRobert, line 135: IMO since you are separating the google cases, they could go with extra cases instead of base cases since hardware requests tend to take a while to complete

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+
+GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS = {
+    "id": "google,juniper",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740227400,
+        endTimestampInSeconds=1740659400,
+        selectedCommits={},
+        filter={"filter_architecture": ["arm64"], "filter_boot.status": ["PASS"]},
+    ),
+}
```
murilx, line 61: In this case where you have two identical cases where some of the data should be always equal (like `id` and `startTimestampInSeconds` and `endTimestampInSeconds`, for example), but other parts can change you can just change that part from the first case.

```suggestion
GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS = copy.deepcopy(GOOGLE_JUNIPER_HARDWARE_WITHOUT_FILTERS)
GOOGLE_JUNIPER_HARDWARE_WITH_FILTERS["body"].filter = {"filter_architecture": ["arm64"], "filter_boot.status": ["PASS"]}
```

in this case do not forget to import `copy`

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+
+HARDWARE_WITH_INVALID_FILTER = {
+    "id": "google,juniper",
+    "body": HardwareDetailsPostBody(
+        startTimestampInSeconds=1740227400,
+        endTimestampInSeconds=1740659400,
+        selectedCommits={},
+        filter={"filter_architecture": ["invalid"]},
+    ),
+}
```
MarceloRobert, line 71: Technically your filter has an "invalid" field, but the filter itself is valid, so I think you should rename this variable and the filter to "unexistent_arch" or something like that

backend/kernelCI_app/unitTests/utils/hardwareClient.py
```diff
@@ -0,0 +1,28 @@
...
+import json
+
+
+class HardwareClient(BaseClient):
+    def get_hardware_boots(
+        self, *, hardware_id: str, body: HardwareDetailsPostBody
+    ) -> requests.Response:
+        path = reverse("hardwareDetailsBoots", kwargs={"hardware_id": hardware_id})
+        url = self.get_endpoint(path=path)
+        return requests.post(url=url, data=json.dumps({**body.model_dump()}))
```
murilx, line 14:
```suggestion
        return requests.post(url=url, data=json.dumps(body.model_dump()))
```

backend/kernelCI_app/unitTests/utils/hardwareClient.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_hardware_builds(
+        self, *, hardware_id: str, body: HardwareDetailsPostBody
+    ) -> requests.Response:
+        path = reverse("hardwareDetailsBuilds", kwargs={"hardware_id": hardware_id})
+        url = self.get_endpoint(path=path)
+        return requests.post(url, data=json.dumps({**body.model_dump()}))
```
murilx, line 21: ditto

backend/kernelCI_app/unitTests/utils/hardwareClient.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_hardware_tests(
+        self, *, hardware_id: str, body: HardwareDetailsPostBody
+    ) -> requests.Response:
+        path = reverse("hardwareDetailsTests", kwargs={"hardware_id": hardware_id})
+        url = self.get_endpoint(path=path)
+        return requests.post(url, data=json.dumps({**body.model_dump()}))
```
murilx, line 28: ditto

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+from kernelCI_app.unitTests.utils.asserts import assert_status_code_and_error_response
+from kernelCI_app.utils import string_to_json
+from http import HTTPStatus
+from kernelCI_app.typeModels.hardwareDetails import HardwareDetailsPostBody
+
+
+client = HardwareClient()
+
+INVALID_HARDWARE = {
+    "id": "no hardware id",
```
MarceloRobert, line 12: ditto about "invalid"

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+        origin="",
+        startTimestampInSeconds=1737487800,
+        endTimestampInSeconds=1737574200,
+        selectedCommits={},
+        filter={},
+    ),
+}
+
+INVALID_REQUEST_BODY = {
+    "id": "google,juniper",
```
MarceloRobert, line 23: ditto about "invalid", in this case it could even be "bad request"

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+
+    if "test_hardware_builds" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "test_hardware_builds,expected_response",
+            base_cases + extra_cases + google_cases,
+        )
+
+    if "test_hardware_tests" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "test_hardware_tests,expected_response", base_cases + extra_cases
```
MarceloRobert, line 158: it depends if you are going to put google cases on extra cases, but why not use the google case here?

backend/kernelCI_app/unitTests/hardware_test.py
murilx, line 135: nit: not exactly the google cases if there's a google case outside of this list, so I'd rename it. Maybe something like `builds_and_boots_cases`

backend/kernelCI_app/unitTests/hardware_test.py
```diff
@@ -0,0 +1,215 @@
...
+        )
+
+
+@online
+def test_get_hardware_details_boots(test_hardware_boots: dict, expected_response: dict):
```
murilx, line 163: nit: Could add `-> None` as the return type for all of these functions. Although, I don't feel the need to type the test functions as much as the source code itself

backend/kernelCI_app/views/hardwareDetailsBootsView.py
```diff
@@ -130,6 +131,14 @@ def post(self, request, hardware_id) -> Response:
             end_datetime=self.end_datetime,
         )
 
+        if len(trees) == 0:
```
MarceloRobert, line 134: can you give an example of when a hardware falls in this condition?
I tried running
```
http POST http://localhost:8000/api/hardware/unexistent \
Content-Type:application/json \
<<< '{
    "origin": "maestro",
    "startTimestampInSeconds": 1740227400,
    "endTimestampInSeconds": 1740659400,
    "selectedCommits": {},
    "filter": {}
}'
```
and it is giving an error 500

backend/kernelCI_app/unitTests/hardware_test.py
Francisco2002: I separate this test in another list because I want to run it in boots and builds only. I will rename this variable, like @murilx suggested above

backend/kernelCI_app/unitTests/hardware_test.py
Francisco2002: ditto

murilx (approved):
LGTM

backend/kernelCI_app/views/hardwareDetailsBootsView.py
MarceloRobert: I see, the other endpoints will be tackled later on other issues

MarceloRobert (approved):
Seems to be working well, LGTM

PR #1026, opened 2025-02-28
fix: remove unused cache functions

### LINKED ISSUE #999: check why pytest does not finish when tests skip
As far as was checked, it's probably due to pytest-django spawning more than 1 process when it runs, and they probably aren't killed correctly when the test is skipped, but more investigation should be done

### LINKED ISSUE #1025: Check and fix why django is not exiting after admin commands
Django is not exiting properly after running certain admin commands. One example is our `generate-schema.sh` script, which executes a Django admin command to generate an OpenAPI schema for our application.  

The issue may be caused by our code **spawning multiple threads** inside [`cache.py`](https://github.com/kernelci/dashboard/blob/a7d885cee7aa20cc5b316503f14f65bf1c4b4fcb/backend/kernelCI_app/cache.py). If these threads are not cleaned up correctly, Django might wait indefinitely, preventing the process from exiting as expected.

murilx:
Closes #999 
Closes #1025

## How to test
I think that currently the best way to test is to run our `generate-schema.sh` script and check that it exits correctly. Another option would be to create a new admin command (for example https://github.com/padovan/dashboard/commits/dev/), run it and check it's exiting correctly

### DIFF
--- a/backend/kernelCI_app/cache.py
+++ b/backend/kernelCI_app/cache.py
@@ -1,9 +1,6 @@
 from django.core.cache import cache
-from django.db import connection
-from threading import Thread
 from django.conf import settings
 
-NOTIFY_CHANNEL = "teste"
 timeout = settings.CACHE_TIMEOUT
 
 _commit_lookup = {}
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/cache.py
```diff
@@ -1,6 +1,4 @@
 from django.core.cache import cache
-from django.db import connection
-from threading import Thread
 from django.conf import settings
 
 NOTIFY_CHANNEL = "teste"
```
MarceloRobert, line 4: You can also remove this const

MarceloRobert (approved):
Seems to be working normally, pre-approving

Francisco2002 (approved):
LGTM

PR #1030, opened 2025-02-28
fix: minor details changes

### LINKED ISSUE #1024: BuildDetails small changes
Change "Defconfig" name to just "config"
Move Misc section to below Artifacts

murilx:
- Remove some duplicate message ids in favor of using the most generic one (e.g. instead of 'buildDetails.compiler' we use 'global.compiler') this includes using 'global.config' instead of 'buildDetails.defconfig' for the config field.
- For fields that are equal in both build details and test details and don't have a global message id, unified the message ids in commonDetails.* (e.g. 'buildDetails.gitCommit' and 'testDetails.gitCommitHash' are not 'commonDetails.gitCommitHash')
- Moved up the files section in build details

Closes #1024


Francisco2002:
Others suggested changes:

- `bootsTab.configs` -> `global.configs`
- `bootsTab.error` -> `global.error`
- `bootsTab.hardwareTested` and `testsTab.hardwareTested` -> `commonTab.hardwareTested`
- `bootsTab.info` -> `global.info`
- `bootsTab.success` -> `global.success`
- `buildAccordion.*` -> only `buildAccordion.logExcerpt` is used. I suggest to change to `global.logExcerpt` or something like it and remove this unused ids.
- `buildDetails.startTime` -> `commonDetails.startTime` (or `global.startTime)` 
- `buildTab.buildStatus` and `filter.buildStatus` -> `globalTable.build`
- `treeDetails.commitOrTag` -> `globalTable.commitTag`

murilx:
@Francisco2002 I agree we should change all of these, but this PR initial focus is the details page, maybe changing tab messages would make the PR harder to review. I can still add these changes if you think they are important, though (I understand that's difficult to find an opportunity to perform refactor task)

About
- `buildDetails.startTime -> global.startTime`
- `treeDetails.commitOrTag -> globalTable.commitTag`
All change these

dashboard/src/components/DetailsLink/DetailsLink.tsx
MarceloRobert, line 1: You can add an index file with the export like it is being made in other components, then you won't have to import from /DetailsLink/DetailsLink

dashboard/src/locales/messages/index.ts
MarceloRobert, line 1: There are a couple more keys that aren't being used, some from build details and a lot from buildAccordion (which we don't even have anymore), could you delete those as well?

MarceloRobert (commented):
I believe that you could use this PR to reorder the fields in the details pages, so that they match a bit better. Here's my suggestion (left is testDetails, right is buildDetails, center is both):
```
status
path / git describe

tree
git repository branch

git commit hash
git repository url

git commit tag
hardware / command

architecture
compiler

build info / config
start time / date

logs / command
test id / build id

compatibles / -
---
- / build logs
- / kernel config
```

dashboard/src/components/BuildDetails/BuildDetails.tsx
```diff
@@ -147,55 +192,37 @@ const BuildDetails = ({
...
+                linkText: formatDate(valueOrEmpty(data.start_time)),
               },
               {
                 title: 'buildDetails.buildId',
                 linkText: buildId,
               },
+              {
+                title: 'global.buildTime',
+                linkText: data.duration ? `${data.duration} sec` : '-',
+              },
```
MarceloRobert, line 225: Can you swap those? I miswrote "command" twice and forgot about the buildTime in my last comment

Francisco2002 (approved):
LGTM

MarceloRobert (approved):
LGTM, nice work

PR #1033, opened 2025-03-05
Feat: issue listing changes

murilx:


dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -73,18 +71,28 @@ const OriginSelect = ({ basePath }: { basePath: string }): JSX.Element => {
...
     case 'tree':
       return <FormattedMessage id="routes.treeMonitor" />;
     case 'hardware':
       return <FormattedMessage id="routes.hardwareMonitor" />;
+    case 'issues':
+      return params ? (
+        <FormattedMessage id="routes.issueDetails" />
+      ) : (
+        <FormattedMessage id="routes.issueMonitor" />
+      );
```
MarceloRobert, line 91: Why do you have this conditional if the URL for issue details is also /issues/?

dashboard/src/types/general.ts
```diff
@@ -361,6 +361,7 @@ export const getTargetFilter = (
 export enum RedirectFrom {
   Tree = 'tree',
   Hardware = 'hardware',
+  Issue = 'issues',
```
MarceloRobert, line 364: Add it in plural directly, so there is no confusion of the routes

MarceloRobert (commented):
Looking good, only some comments

dashboard/src/components/Issue/IssueSection.tsx
```diff
@@ -47,7 +47,7 @@ const IssueSection = ({
           <Link
             key={issue.id + issue.version}
             className="mb-16 flex not-last:mb-2 not-last:border-b not-last:pb-2"
-            to="/issue/$issueId"
+            to="/issues/$issueId"
```
padovan, line 50: Nack! individual issues URL stay as is. Only the top level one is changing to `issues`. We have sent links to dashboard issues all over the place. So we can't change that anymore.

dashboard/src/components/TopBar/TopBar.tsx
Francisco2002: This was used to differentiate the Title of issues listing and issues details. With the modification that Padovan requests, this will be changed.

dashboard/src/components/Issue/IssueSection.tsx
Francisco2002: @padovan done

dashboard/src/components/Icons/IssueCulpritInfo.tsx
MarceloRobert, line 1: There's no need for this component, please reuse `TooltipIcon`

MarceloRobert (approved):
LGTM, nice work. There could be the issue breadcrumb for test details and build details, but I'll pre-approve

PR #1035, opened 2025-03-05
Feat: schema v5 support

### LINKED ISSUE #1022: Add support for v5 db schema change
As said in [this email](https://groups.io/g/kernelci-webdashboard/message/145), the v5 of the database schema is backwards incompatible, so we should take the necessary precautions to not break the dashboard.
The main problem is the change from build valid to build status, which we are using heavily in the dashboard.

Remember to change all uses in the frontend and also in the backend.
The dashboard should support both schemas while the change is underway, then we can move to support only build status

murilx:
Schema changes that will affect the dashboard:
- [x] `DROP "waived" FROM "tests"`
- [x] `DROP "contacts" FROM "checkouts"`
- [x] `DROP "build_valid"/"test_status" FROM "issues"` (Was already dropped from the dashboard)
- [ ] `REPLACE "valid" WITH "status" IN "builds"`

Schema changes that won't affect the dashboard
- BE STRICTER ABOUT TEST PATHS 
- PROHIBIT NULL CHARACTERS ('\0') IN STRINGS
- ONLY ALLOW HTTPS URLS FOR REPOS

Closes #1022 

## How to test
- Navigate through the application - tree details, hardware details, build details - and verify that everything works correctly


backend/kernelCI_app/viewCommon.py
```diff
@@ -1,38 +1,30 @@
...
-from kernelCI_app.helpers.build import build_status_map
+from kernelCI_app.typeModels.commonDetails import BuildHistoryItem
 
 
-class BuildDict(TypedDict):
-    valid: str
-    config_name: str
-    architecture: str
-    compiler: str
+def create_default_build_status() -> dict:
```
MarceloRobert, line 4: If you use a pydantic model with a default value you won't need this function, see if you can already remove it or add a TODO to do it

backend/kernelCI_app/queries/tree.py
```diff
@@ -0,0 +1,108 @@
...
+from kernelCI_app.cache import get_query_cache, set_query_cache
+from kernelCI_app.helpers.treeDetails import create_checkouts_where_clauses
+
+
+def get_tree_details_data(request, commit_hash: str) -> list[tuple]:
+    cache_key = "treeDetails"
+
+    origin_param = request.GET.get("origin")
+    git_url_param = request.GET.get("git_url")
+    git_branch_param = request.GET.get("git_branch")
```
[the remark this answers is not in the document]
murilx: This is how it's currently done in `helpers/treeDetails.py:get_tree_details_data` as well. However, I agree this is the correct approach for this problem. I can work on improving this even though it's not the focus of this PR

backend/kernelCI_app/queries/hardware.py
```diff
@@ -0,0 +1,509 @@
...
+    hardware;
+    """
+    with connection.cursor() as cursor:
+        cursor.execute(query, params)
+        return dict_fetchall(cursor)
+
+
+@typing_extensions.deprecated(
+    "This implementation is temporary while the schema is being updated."
+)
```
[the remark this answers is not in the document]
murilx: Missplaced the notation... I'll change it to the correct function

backend/kernelCI_app/helpers/commonDetails.py
```diff
@@ -9,11 +9,11 @@ def add_unfiltered_issue(
     issue_version: int,
     should_increment: bool,
     issue_set: Set,
-    is_invalid: bool,
+    is_failed_build: bool,
```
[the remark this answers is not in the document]
murilx: Changed it to `is_failed_task`

dashboard/src/utils/utils.ts
```diff
@@ -126,20 +127,8 @@ export const sanitizePlatforms = (
...
-  build_valid?: boolean | null,
-): AccordionItemBuilds['status'] => {
-  if (build_valid === true) {
-    return 'pass';
-  }
-  if (build_valid === false) {
-    return 'fail';
-  }
-  return 'null';
+const isBuildError = (build_status: Status): number => {
```
MarceloRobert, line 130: Add a TODO to check if this is still necessary since it will become just a simple comparison

dashboard/src/utils/utils.ts
MarceloRobert, line 1: See if you can pass the other status to the mappings as well

dashboard/src/utils/utils.ts
MarceloRobert: Also, your change is inverted, it should be 1 when the status is FAIL

MarceloRobert (commented):
It seems that clicking the circles in the build status card is not filtering correctly (though filtering by the modal is normal)

dashboard/src/components/Tabs/Builds/StatusCard.tsx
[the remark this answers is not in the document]
murilx: I don't think I can combine them as quick as I'd like because of the feature of filtering by clicking in the status card that only exists in the builds status card and to remove it now would make a regression. Other than that it should be really easy to combine then. I'll add a TODO

backend/kernelCI_app/helpers/filters.py
```diff
@@ -4,15 +4,21 @@
...
 from kernelCI_app.utils import get_error_body_response
 from kernelCI_app.constants.general import UNKNOWN_STRING
 
 NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
 
 
-def is_build_invalid(build_valid: Optional[bool]) -> bool:
-    return build_valid is None or build_valid is False
+def is_build_invalid(build_status: StatusValues) -> bool:
+    return build_status == NULL_STATUS or build_status == FAIL_STATUS
```
MarceloRobert, line 21: Looking closer, add a TODO to check if this function could be the same as `is_test_failure`

murilx:
> It seems that clicking the circles in the build status card is not filtering correctly (though filtering by the modal is normal)

It seems that when you click on a circle it tries to filter by its legend ('Success', 'Failed', 'Inconclusive'). Even though that was the case for build filters before, that's no longer true. I'll take some time to fix this behavior but it should be possible

dashboard/src/components/Tabs/Builds/StatusCard.tsx
```diff
@@ -17,16 +17,17 @@ interface IStatusCard {
   ) => void;
 }
 
+// TODO: Remove this component and start using Tabs/Tests/StatusCard.tsx instead
```
MarceloRobert, line 20:
```suggestion
// TODO: Combine this component with Tabs/Tests/StatusCard.tsx instead
```

dashboard/src/utils/utils.ts
murilx: Done

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -366,12 +356,7 @@ def _process_time_range(self, request) -> None:
         except Exception as ex:
             log_message(ex)
 
-    @extend_schema(
-        responses=TreeCommitsResponse,
-        parameters=[TreeCommitsQueryParameters],
-        methods=["GET"],
-    )
-    def get(self, request, commit_hash: Optional[str]) -> Response:
+    def get_params(self, request) -> tuple[str, str, str, str, str, list[str]]:
```
MarceloRobert, line 359: Add a TODO to remove this function and use a pydantic model validation, it will raise a validation error if there are missing parameters and will make the code a bit cleaner, without the need for this method

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -396,104 +405,22 @@ def get(self, request, commit_hash: Optional[str]) -> Response:
...
-                query,
-                self.field_values,
+        if rows is None:
+            message = (
+                "This error was probably caused because the server was using"
+                "an old version of the database. Please try requesting again"
+            )
+            return create_api_error_response(
+                error_message=message,
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
```
MarceloRobert, line 422: This should be temporary, right?

MarceloRobert (commented):
I believe all those messages of "probably caused by a change in the schema" should have a comment of being temporary, otherwise future problems with the endpoint could be told as "caused by a change in schema" when they are not

MarceloRobert (approved):
All seems to be working from my tests. Great job!!

Francisco2002 (approved):
Nice job!!!

PR #1045, opened 2025-03-07
Feat: add discord webhook helpers

### LINKED ISSUE #1027: Add support to discord webhook to send notifications

MarceloRobert:
Adds helper functions to send messages using a discord webhook

## How to test
Get a discord webhook url, by going to discord server settings, creating or viewing a webhook and copying its URL
Set the environment variable `DISCORD_WEBHOOK_URL` with the url of your webhook (should look something like `https://discord.com/api/webhooks/<id>/<webhook_id>`)
Currently there is only a notification for failed builds that had tests, such as this [localhost buildDetails](http://localhost:5173/build/maestro:67c8a01118018371956c8c4b)

Closes #1027

### DIFF
--- a/backend/kernelCI_app/views/buildTestsView.py
+++ b/backend/kernelCI_app/views/buildTestsView.py
@@ -1,4 +1,6 @@
 from http import HTTPStatus
+from kernelCI_app.helpers.discordWebhook import send_discord_notification
+from kernelCI_app.helpers.logger import create_endpoint_notification
 from kernelCI_app.typeModels.buildDetails import BuildTestsResponse
 from kernelCI_app.models import Tests
 from drf_spectacular.utils import extend_schema
@@ -18,6 +20,7 @@ class BuildTests(APIView):
             "start_time",
             "environment_compatible",
             "environment_misc",
+            "build__valid",
         )
 
         if not result:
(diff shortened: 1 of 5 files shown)


murilx:
Just a quick correction, is seems that you have implemented code for both failed builds with tests but also for trees with no builds, tests or boots, or am I overlooking something?

backend/kernelCI_app/views/buildTestsView.py
```diff
@@ -26,6 +29,13 @@ def get(self, request, build_id: str) -> Response:
                 status=HTTPStatus.OK,
             )
 
+        if result[0]["build__valid"] is False:
```
WilsonNet, line 32: check if result[0] exists

murilx: I think that the last if (`if not result`) should be enough, shouldn't it?

```python
Python 3.12.7 (main, Nov 18 2024, 15:03:02) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = []
>>> not a
True
>>> a.append(1)
>>> not a
False
```

murilx: but I'd still do
```suggestion
        if result[0].get("build__valid") is False:
```

just to be sure

docker-compose.yml
```diff
@@ -45,6 +45,7 @@ services:
             - DB_DEFAULT_USER=${DB_DEFAULT_USER:-kernelci}
             - DJANGO_SECRET_KEY=${DJANGO_SECRET_KEY}
             - DEBUG=False
+            - DISCORD_WEBHOOK_URL=${DISCORD_WEBHOOK_URL}
```
WilsonNet, line 48: Add the env description to the backend README file

WilsonNet (approved):
nice job, worked on my tests

murilx (approved):
Worked well on my tests

PR #1046, opened 2025-03-07
Fix: test details without env misc

MarceloRobert:
Fixes test details pages where there is no environment misc to extract the platform, which were crashing because of the status history filter

Example: http://localhost:5173/test/redhat%3A1575003976-x86_64-kernel_upt_1
Compare it with the same page on staging: https://staging.dashboard.kernelci.org:9000/test/redhat%3A1575003976-x86_64-kernel_upt_1


murilx (approved):
Worked well on my tests

WilsonNet (approved):
worked here

PR #1061, opened 2025-03-11
feat: log-viewer- improvements

### LINKED ISSUE #996: Improve log exerpt viewer
Image

* Expand log view part to take most of the vertical space
* Add test information at the top
* Drop issues (#993)
* Add a View full log button

Beyond what is on the image: After the Highlights sections add: "Sometimes log excepts doesn't bring the most relevant data for a test case. In such case, check the full log (!add link). We also invite you to create an issue<link to KCIDB new issue> in reporting the problem.

(the new issue link should include at list a link to the test case in the description by default)

WilsonNet:
Closes #996

### DIFF
--- a/dashboard/src/components/BuildDetails/BuildDetails.tsx
+++ b/dashboard/src/components/BuildDetails/BuildDetails.tsx
@@ -49,6 +49,7 @@ import ButtonOpenLogSheet from '@/components/Button/ButtonOpenLogSheet';
 
 import MemoizedLinkItem from '@/components/DetailsLink';
 import { LinkIcon } from '@/components/Icons/Link';
+import { processLogData } from '@/hooks/useLogData';
 
 import BuildDetailsTestSection from './BuildDetailsTestSection';
 
@@ -72,6 +73,12 @@ const BuildDetails = ({
     status: issueStatus,
     error: issueError,
   } = useBuildIssues(buildId);
+  const logData = useMemo(() => {
+    if (!data) {
+      return undefined;
+    }
+    return processLogData(buildId, { type: 'build', ...data });
+  }, [buildId, data]);
 
   const { logOpen } = useSearch({ from: '/_main/build/$buildId' });
   const navigate = useNavigate({ from: '/build/$buildId' });
@@ -308,8 +315,7 @@ const BuildDetails = ({
               type={sheetType}
               jsonContent={jsonContent}
               hideIssueSection
-              logUrl={data?.log_url}
-              logExcerpt={data?.log_excerpt}
+              logData={logData}
             />
           </Sheet>
         </ErrorBoundary>
--- a/dashboard/src/components/Filter/CodeBlock.tsx
+++ b/dashboard/src/components/Filter/CodeBlock.tsx
@@ -83,7 +83,7 @@ const Code = ({
       {variant === 'log-viewer' && statsElement}
       <pre
         className={cn(
-          'w-full max-w-[100vw] overflow-x-auto rounded-md bg-[#DDDDDD] p-4 font-mono text-sm leading-4 text-[#767676]',
+          'max-h-[80%] w-full max-w-[100vw] overflow-auto rounded-md bg-[#DDDDDD] p-4 font-mono text-sm leading-4 text-[#767676]',
           className,
         )}
       >
@@ -237,7 +237,7 @@ const CodeBlock = ({
 
   return (
     <>
-      <div className="h-full">
+      <div className="min-h-[300px]">
         <div className="pl-3">
           {variant === 'log-viewer' && (
             <h3 className="py-1 text-2xl font-bold">
@@ -273,9 +273,6 @@ const CodeBlock = ({
         )}
 
         <MemoizedCode
-          className={cn('', {
-            'max-h-[425px]': variant !== 'log-viewer',
-          })}
           variant={variant}
           parsedCode={
             typeof parsedCode === 'string'
--- a/dashboard/src/components/Log/LogViewerCard.tsx
+++ b/dashboard/src/components/Log/LogViewerCard.tsx
@@ -1,21 +1,53 @@
-import { FormattedMessage } from 'react-intl';
-import { LuFileJson } from 'react-icons/lu';
+import { FormattedMessage, useIntl } from 'react-intl';
 
 import { useMemo, type JSX } from 'react';
 
 import { Link } from '@tanstack/react-router';
 
-import BaseCard from '@/components/Cards/BaseCard';
+import { SearchIcon } from '@/components/Icons/SearchIcon';
+import { StatusIcon } from '@/components/Icons/StatusIcons';
+import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
+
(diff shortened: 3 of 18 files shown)


Francisco2002 (approved):
LGTM

dashboard/src/components/BuildDetails/BuildDetails.tsx
MarceloRobert, line 1: I understand that because of the cache the useLogData hook won't take long, but you are already making a buildDetails request for this page, can't you just create a `LogData` object with the data that you already have and pass it to the LogOrJsonSheetContent? Then you won't need the `useLogData`

dashboard/src/components/Filter/CodeBlock.tsx
MarceloRobert, line 1: I'm not sure if this is the culprit but I see that the log sheet now has a weird scroll nesting in it, instead of just the code block being scrollable it's the entire log excerpt part

dashboard/src/components/Log/LogViewerCard.tsx
```diff
@@ -24,39 +56,73 @@ export const LogViewerCard = ({
...
+            />
+            <div className="flex w-10 justify-center">
+              <SearchIcon className="text-blue w-full" />
+            </div>
+          </Link>
+        </>
+      );
+    } else {
+      return <FormattedMessage id="logSheet.noLogFound" />;
+    }
```
MarceloRobert, line 87: This will say "No logs available for download" while on staging it says "No logs available". IMO the older one is better.
ref: http://localhost:5173/tree/73b8c1dbc2508188e383023080ce6a582ff5f279?o=0dayci&ti%7Cc=next-20250320&ti%7Cch=73b8c1dbc2508188e383023080ce6a582ff5f279&ti%7Cgb=master&ti%7Cgu=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fnext%2Flinux-next.git

dashboard/src/components/Log/LogViewerCard.tsx
```diff
@@ -24,39 +56,73 @@ export const LogViewerCard = ({
...
+              </span>
+              <div className="text-sm">
+                <FormattedMessage
+                  id="title.hardwareDetails"
+                  values={{
+                    hardwareName: `${hardware} (${logData?.architecture})`,
+                  }}
+                />
+              </div>
+              {variant === 'modal' && linkComponent}
```
MarceloRobert, line 113: nit: I would prefer a bit more spacing between the first two infos and this linkComponent. Specially when there is no log available it just looks like a clump of different font weights and sizes

dashboard/src/components/TestDetails/TestDetails.tsx
MarceloRobert, line 1: ditto unnecessary useLogData when you already have useTestDetails

dashboard/src/routes/log-viewer.tsx
MarceloRobert, line 1: Don't you need some catch for invalid values? Specially for the type.
ref: http://localhost:5173/log-viewer?itemId=maestro%3A67d8ed5428b1441c081b944e&type=invalid&url=https%3A%2F%2Ffiles.kernelci.org%2Fbaseline-arm64-android-67d8ed5428b1441c081b944e%2Flog.txt.gz

dashboard/src/pages/LogViewer.tsx
```diff
@@ -12,17 +14,65 @@ import { LogExcerpt } from '@/components/Log/LogExcerpt';
...
+import { LogViewerCard } from '@/components/Log/LogViewerCard';
 
 export function LogViewer(): JSX.Element {
-  const { url } = useSearch({ from: '/log-viewer' });
+  const { url, type, itemId } = useSearch({ from: '/log-viewer' });
 
   const { data, status } = useLogViewer(url);
+  const { data: logData, isLoading } = useLogData(
+    itemId ?? '',
+    type ?? 'build',
```
MarceloRobert, line 26: Are you sure you can default to build here? This log is from a test, but without the type it is trying to get the build details for it and is showing wrong data: http://localhost:5173/log-viewer?itemId=maestro%3A67d8ed5428b1441c081b944e&url=https%3A%2F%2Ffiles.kernelci.org%2Fbaseline-arm64-android-67d8ed5428b1441c081b944e%2Flog.txt.gz

dashboard/src/components/Log/LogViewerCard.tsx
murilx: I added a margin of 12 pixels (mb-3) between hardware and the link. Do you think it's better?

dashboard/src/pages/LogViewer.tsx
```diff
@@ -12,17 +12,107 @@ import { LogExcerpt } from '@/components/Log/LogExcerpt';
...
+  const { url, type, itemId } = useSearch({ from: '/log-viewer' });
+  const historyState = useRouterState({ select: s => s.location.state });
+  const previousSearch = useSearchStore(s => s.previousSearch);
+
+  let handledType: LogType =
+    type === 'build' || type === 'test' ? type : 'build';
+  let handledItemId = itemId ?? '';
+  if ((type !== 'build' && type !== 'test') || !type || !itemId) {
+    handledItemId = '';
+    handledType = 'build'; // Not used in this case, just necessary for useLogData typing
```
MarceloRobert, line 33: I think it would be better if useLogData accepted an undefined type, instead of an unused default type

dashboard/src/hooks/useLogData.ts
```diff
@@ -0,0 +1,95 @@
...
+  type: 'build';
+};
+
+type TypedTTestDetails = Partial<TTestDetails> & {
+  type: 'test';
+};
+
+export const processLogData = (
+  id: string,
+  data: TypedTBuildDetails | TypedTTestDetails,
```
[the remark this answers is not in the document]
murilx: This is a type of discriminated union (https://dev.to/darkmavis1980/what-are-typescript-discriminated-unions-5hbb). If I pass type as a different parameter rather than inside the object, typescript won't be able to correctly determine if a type should exist or not.

```typescript
const handledTitle = type === 'build' ? data.git_commit_name : git.path
```

Will throw an error because `git_commit_name` only exist in `TBuildDetails` but typescript can't determine if `data` is `TBuildDetails` or `TTestDetails`. However,

```typescript
const handledTitle = data.type === 'build' ? data.git_commit_name : git.path;
```

won't throw the error. This is only one solution though, other possibility is to create a type gard inside `processLogData` and check at the start if `data` is `TBuildDetails` or `TTestDetails`. This way is only the solution I decided to go for

dashboard/src/pages/LogViewer.tsx
murilx: At first I struggled with this idea because I thought it would cause some errors. It indeed is not the best solution I believe (because it breaks the dichotomy of 'test' and 'build' and adds another option) but if we handle the hook data correctly it's much better and elegant than my initial solution

MarceloRobert (approved):
LGTM, nice work!

PR #1062, opened 2025-03-11
Chore: logic to construct tree proof file

### LINKED ISSUE #1015: Store in a file the tree identification for all trees
A lot of trees have the same name or don't have a name, and since there is no `tree_id` we could have a file that maps the trees by their relevant fields to a unique "name" (or an identifier).
This should be done to the trees of all origins.
This could be stored in a simple file in the backend, since we don't have access to the main database.

Francisco2002:
- create admin command to make a file with trees of all origin (focus in maestro trees)
- create a initial file with the trees

Close #1015

### **How to test**

- Drop `kernelCI_app/management/commands/data/tree-proof.yml` file
- Run the command `poetry run python3 manage.py treeproof`
- See that it's generate a new `tree-proof.yml`

### DIFF
--- a/backend/kernelCI_app/management/commands/treeproof.py
+++ b/backend/kernelCI_app/management/commands/treeproof.py
@@ -1,0 +1,91 @@
+from django.core.management.base import BaseCommand
+from kernelCI_app.models import Checkouts
+import re
+import yaml
+import os
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    def __init__(self):
+        self.maestro_trees = dict()
+        self.non_maestro_trees = dict()
+        self.trees: dict[str, dict] = {"trees": {}}
+
+    def _define_tree_name(self, git_url: str):
+        regex = r"([^/]+)/([^/]+)?$"
+        match = re.search(regex, git_url)
+
+        first_part = match.group(1)
+        second_part = match.group(2).split(".")[0]
+
+        new_tree_name = first_part
+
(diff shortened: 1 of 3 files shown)


backend/kernelCI_app/management/commands/data/tree-proof.yml
```diff
@@ -0,0 +1,157 @@
+trees:
```
padovan, line 1: This looks good. We can use this as starting point. A few more things:

* rename it to `trees-name.yaml`
* move the `data` dir to `backend/`

backend/kernelCI_app/management/commands/treeproof.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def handle(self, *args, **options):
+        self._get_trees_proofs()
+        self._get_trees_proofs(is_maestro=True)
+
+        merged_dict = self._merge_trees()
+        formatted_dict = self._format_trees_to_yml_format(merged_dict)
+
+        base_dir = os.path.dirname(os.path.abspath(__file__))
+        filepath = os.path.join(base_dir, "data/tree-proof.yml")
```
padovan, line 87: Let's set add the data dir as a settings in Django and pull that setting here instead of this path construction you are doing.  `BACKEND_DATA_DIR`.

padovan (changes requested):

padovan (approved):

murilx (approved):
LGTM

PR #1064, opened 2025-03-11
test: tree commit history

### LINKED ISSUE #987: Unit tests for Tree Commits History View

murilx:
Closes #987

### DIFF
--- a/backend/kernelCI_app/unitTests/treeCommitHistory_test.py
+++ b/backend/kernelCI_app/unitTests/treeCommitHistory_test.py
@@ -1,0 +1,119 @@
+import pytest
+from http import HTTPStatus
+from requests import Response
+from kernelCI_app.unitTests.utils.healthCheck import online
+from kernelCI_app.unitTests.utils.client.treeClient import TreeClient
+from kernelCI_app.unitTests.utils.asserts import assert_status_code_and_error_response
+from kernelCI_app.unitTests.utils.commonTreeAsserts import (
+    assert_tree_commit_history_fields,
+)
+from kernelCI_app.utils import string_to_json
+
+
+client = TreeClient()
+
+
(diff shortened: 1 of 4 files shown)


backend/kernelCI_app/unitTests/treeCommitHistory_test.py
```diff
@@ -0,0 +1,132 @@
...
+    pytestconfig, tree: dict, status_code: HTTPStatus, has_error_body: bool
+) -> None:
+    response, content = request_data(tree)
+    assert_status_code_and_error_response(
+        response=response,
+        content=content,
+        status_code=status_code,
+        should_error=has_error_body,
+    )
+
```
murilx, line 69: assert that `tree_id == response["git_commit_hash"]

murilx: Thought of doing this initially but realized each element should have a different git_commit_hash, so I don't think it's necessary. We could assert that at least one are equal to tree_id though

backend/kernelCI_app/unitTests/treeCommitHistory_test.py
```diff
@@ -0,0 +1,132 @@
...
+MISSING_ORIGIN_TREE = {
+    "id": "invalid_id",
+    "query": {
+        "origin": "",
+        "git_url": "",
+        "git_branch": "",
+        "start_timestamp_in_seconds": "",
+        "end_timestamp_in_seconds": "",
+    },
+}
```
MarceloRobert, line 25: You seem to not be using this variable

backend/kernelCI_app/unitTests/treeCommitHistory_test.py
```diff
@@ -0,0 +1,132 @@
...
+    return response, content
+
+
+@online
+@pytest.mark.parametrize(
+    "tree, status_code, has_error_body",
+    [
+        # Checking for invalid_id
+        ({"id": "invalid_id", "query": {"origin": "maestro"}}, HTTPStatus.OK, True),
+        # Checking for invalid query body
```
MarceloRobert, line 54: I don't think there's a need for these comments

MarceloRobert (commented):
Looking good, only some comments

MarceloRobert (approved):
Seems to be a bit slow for me but it is working fine

PR #1069, opened 2025-03-12
notifications: add arrows to highligh time direction

padovan:


MarceloRobert (approved):
I believe this is the correct arrow orientation based on comparing the notifications sent and the dashboard and also by checking [the query where the tests come from](https://github.com/kernelci/dashboard/blob/635c3b7e245b6bba92d317d5807749d505e82809/backend/kernelCI_app/management/commands/libs/kcidb.py#L390). LGTM

murilx (approved):
LGTM

PR #1070, opened 2025-03-12
notifications: fix test status query

padovan:
We should be selecting distinct id and look at data ordered by start time.

### DIFF
--- a/backend/kernelCI_app/management/commands/libs/kcidb.py
+++ b/backend/kernelCI_app/management/commands/libs/kcidb.py
@@ -396,17 +396,11 @@ def kcidb_tests_results(origin, giturl, branch, hash, path):
         "branch": branch,
         "hash": hash,
         "path": path,
-        "interval": "25 days",
+        "interval": "30 days",
     }
 
     query = """
-            WITH latest_checkout AS (
-                SELECT
-                    _timestamp
-                FROM checkouts
-                WHERE git_commit_hash = %(hash)s
-            ),
-            ranked_tests AS (
+            WITH prefiltered_tests AS (
                 SELECT
                     t.id,
                     t.path,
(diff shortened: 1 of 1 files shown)


padovan:
@MarceloRobert this is working as expected and matching the your query in the cases I checked in the dashboard.

MarceloRobert (commented):
As we are discussing in discord, it seems that your query is returning 2 results for [this kernelci_sleep test](https://staging.dashboard.kernelci.org:9000/test/maestro:67cf61dec5336ecbf369c77e), while the query that is being used in the dashboard can return the last 10 results.

I believe it is an issue with `row_number()`, if I remove the `where rn <= 9` I can see the other tests in the history

padovan:
> As we are discussing in discord, it seems that your query is returning 2 results for [this kernelci_sleep test](https://staging.dashboard.kernelci.org:9000/test/maestro:67cf61dec5336ecbf369c77e), while the query that is being used in the dashboard can return the last 10 results.
> 
> I believe it is an issue with `row_number()`, if I remove the `where rn <= 9` I can see the other tests in the history

The issue there is that it will bring way too many results for other tests for some reason.

backend/kernelCI_app/management/commands/libs/kcidb.py
```diff
@@ -416,28 +410,59 @@ def kcidb_tests_results(origin, giturl, branch, hash, path):
...
+                AND p.platform = t1.platform
+                AND p.config_name = t1.config_name
+            ),
+            ranked_tests AS (
+                SELECT *,
+                    ROW_NUMBER() OVER (
+                        PARTITION BY path, platform, config_name
+                        ORDER BY start_time DESC NULLS LAST
+                    ) as rn
+                FROM final_prefilter
```
MarceloRobert, line 460: I believe T1 will return a single row most of the time, unless there are more than one of the "same" test (as we are considering here) within a specific commit_hash and have different `path, platform, config_name` (meaning more than one test where `rn = 1 and git_commit_hash = %(hash)s`). And if it so happens that T1 has more than one row, then ranked_tests won't have rows with unique `rn`. So you will be getting, for example, rows with conditions (a, b, c) and (a, b, d) from T1, and getting the tests with `rn` 1, 2, 3, 4 for conditions (a, b, c) and also 1, 2, 3 for conditions (a, b, d). I assume this is expected, but if T1 is supposed to only return a single row, then the condition up there would have to change but in turn you won't need this last `row_number()`, just a `limit` will do.

padovan: There can only one test with rn = 1. That is why we putting row_numbers into things.

With top1, I want to make sure I only show a history if the last test belongs to current checkout.

MarceloRobert (approved):
I saw that for some tests which have different sets of path, platform, and config this query is returning more than 5 results, such as this ltp test: https://staging.dashboard.kernelci.org:9000/test/maestro%3A67d1bce0f59fb9874e7d0a47. But, as you said, this is expected, so I approve

PR #1071, opened 2025-03-12
Test: add tree details routes tests

### LINKED ISSUE #982: Unit tests for Tree Details Builds View

### LINKED ISSUE #986: Unit tests for Tree Details Full View

Francisco2002:
- Tree Details Full
- Tree Details Builds

Close #982
Close #986

### DIFF
--- a/backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
+++ b/backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
@@ -1,0 +1,71 @@
+from kernelCI_app.unitTests.utils.healthCheck import online
+from kernelCI_app.unitTests.utils.client.treeClient import TreeClient
+from kernelCI_app.unitTests.utils.asserts import (
+    assert_status_code_and_error_response,
+    assert_has_fields_in_response_content,
+    assert_build_filters,
+)
+from kernelCI_app.utils import string_to_json
+from kernelCI_app.unitTests.utils.fields.tree import tree_builds_expected_fields
+from http import HTTPStatus
+import pytest
+from kernelCI_app.unitTests.utils.treeDetailsCommonTestCases import (
+    ANDROID_MAESTRO_MAINLINE,
+    NEXT_PENDING_FIXES_BROONIE,
+    UNEXISTENT_TREE,
+    INVALID_QUERY_PARAMS,
+)
+
+client = TreeClient()
+
+
+@online
+@pytest.mark.parametrize(
+    "base_tree, status_code, has_error_body",
+    [
+        (ANDROID_MAESTRO_MAINLINE, HTTPStatus.OK, False),
+        (NEXT_PENDING_FIXES_BROONIE, HTTPStatus.OK, False),
+        (UNEXISTENT_TREE, HTTPStatus.OK, True),
+        (INVALID_QUERY_PARAMS, HTTPStatus.OK, True),
+    ],
+)
+def test_tree_details_build(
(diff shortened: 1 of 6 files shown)


backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
```diff
@@ -0,0 +1,66 @@
...
+    [
+        (ANDROID_MAESTRO_MAINLINE, HTTPStatus.OK, False),
+        (NEXT_PENDING_FIXES_BROONIE, HTTPStatus.OK, False),
+        (UNEXISTENT_TREE, HTTPStatus.OK, True),
+        (INVALID_QUERY_PARAMS, HTTPStatus.OK, True),
+    ],
+)
+def test_tree_details_build(
+    pytestconfig,
+    search_params: dict,
```
murilx, line 33: nit: instead of `search_params` I'd call this `base_tree` or something like that

backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
murilx, line 1: It would be better if we had tests that verified if the filters are actually modifying the result. For example, if you filter by a certain config_name, all builds returned by the endpoint should have that config_name. If you are filtering for status valid, any other status should return zero, etc.

murilx: Maybe just add this assertions when we have --run-all to speed up the tests on a daily basis

backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
Francisco2002: done

backend/kernelCI_app/unitTests/treeDetailsBuilds_test.py
Francisco2002: This is a good ideia. Thanks!!

backend/kernelCI_app/unitTests/utils/asserts.py
murilx, line 1: Really liked the implementation. There's room for improvement for fields that return a "status object", for example, where any status that's not in the filter should have value 0. Just to clarify

```python
# Example of status object that's present in some fields in summary. tests.status and boots.architecture are examples
status_object = {
    'PASS': 10,
    'FAIL': 15,
    'MISS': 5,
    'SKIP': 2,
    'ERROR': 0,
    'NULL': 0,
}
```

If we filter by `tests.status = 'PASS'` our status object should become

```python
status_object = {
    'PASS': 10,
    'FAIL': 0,
    'MISS': 0,
    'SKIP': 0,
    'ERROR': 0,
    'NULL': 0,
}
```

But that's just food for thought. I believe this is reasonable for now

backend/kernelCI_app/unitTests/utils/treeDetailsCommonTestCases.py
```diff
@@ -0,0 +1,62 @@
...
+            git_branch="for-kernelci",
+        ),
+        "filters": {
+            "boot.status": "PASS",
+            "config_name": "multi_v7_defconfig",
+            "test.hardware": "fsl,imx6q-sabrelite",
+            "test.status": "SKIP",
+        },
+    },
+}
```
murilx, line 62: nit: could you add the dashboard link from where you got these trees? Just so its easier for others to review as well

murilx (approved):
Working well

PR #1072, opened 2025-03-12
Feat: status history improvements

### LINKED ISSUE #1048: Test Status Improvements
- [ ] Make test history icons larger
- [ ] Show test status on Hover
- [ ] Show unstable description on embed 
- [ ] Highlight which icon is current test (test id = page details id)

### LINKED ISSUE #1065: check test status history consistency
The status history for some tests doesn't seem to be correct, it is not the same history between the dashboard and the notification system.

Example of some tests from the notification system: https://groups.io/g/kernelci-results/message/58639
Query from the notification system: https://github.com/kernelci/dashboard/blob/7cea9c7df7bd315d9427d5054901151318d386f1/backend/kernelCI_app/management/commands/notifications.py#L408

Example SQL query from the notification system:
```SQL
WITH
	LATEST_CHECKOUT AS (
		SELECT
			_TIMESTAMP
		FROM
			CHECKOUTS
		WHERE
			GIT_COMMIT_HASH = '4d872d51bc9d7b899c1f61534e3dbde72613f627'
	),
	RANKED_TESTS AS (
		SELECT
			T.ID,
			T.PATH,
			T.STATUS,
			T.START_TIME,
			T.ENVIRONMENT_MISC ->> 'platform' AS PLATFORM,
			B.ARCHITECTURE,
			B.COMPILER,
			B.CONFIG_NAME,
			C.GIT_COMMIT_HASH,
			ROW_NUMBER() OVER (
				PARTITION BY
					T.PATH
				ORDER BY
					T.START_TIME DESC NULLS LAST
			) AS RN
		FROM
			TESTS T
			JOIN BUILDS B ON T.BUILD_ID = B.ID
			JOIN CHECKOUTS C ON B.CHECKOUT_ID = C.ID
			JOIN LATEST_CHECKOUT LC ON LC._TIMESTAMP > C._TIMESTAMP
		WHERE
			T.ORIGIN = 'maestro'
			AND C.GIT_REPOSITORY_URL = 'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git'
			AND C.GIT_REPOSITORY_BRANCH = 'master'
			AND T.PATH LIKE 'kernelci_sleep'
			AND T.ENVIRONMENT_MISC ->> 'platform' != 'kubernetes'
			AND C._TIMESTAMP >= NOW() - INTERVAL '25 days'
			AND B._TIMESTAMP >= NOW() - INTERVAL '25 days'
			AND T._TIMESTAMP >= NOW() - INTERVAL '25 days'
	)
SELECT
	*
FROM
	RANKED_TESTS
WHERE
	RN <= 9
ORDER BY
	PATH,
	START_TIME DESC NULLS LAST;
```

### LINKED ISSUE #1073: Bug: test status history renders inverted
If you go to a test details link on staging, such as this [kernelci_sleep](https://staging.dashboard.kernelci.org:9000/test/maestro:67cf61dec5336ecbf369c77e), the status history might render inverted, especially if you navigate to a previous test and go back to that page

MarceloRobert:
Changes backend query to also consider the build config as a part of a test definition; also uses `start_time` instead of `_timestamp` for filtering and ordering.
Changes frontend to be more visible and clearer in the informations - increased icon size, current test indicator, explicit regression type tooltip.
Changes the way the reverse status history icons are calculated, fixing a bug where it was reverting the already reversed array.

## How to test
Go to any test details pages and compare with staging.
Localhost examples:
[Regression test](http://localhost:5173/test/maestro:67ce445218018371957db690)
[Fixed test](http://localhost:5173/test/maestro%3A67c8eddd18018371956edb3d)
[Unstable test](http://localhost:5173/test/maestro:67cf61dec5336ecbf369c77e) - This one also has a "crossed" timestamp x start time with the latest test, meaning that timestamp from A is > B but start time from A is < B, so on its [staging counterpart](https://staging.dashboard.kernelci.org:9000/test/maestro:67cf61dec5336ecbf369c77e) there is one last test missing

Closes #1048 
Closes #1065 
Closes #1073

### DIFF
--- a/backend/kernelCI_app/queries/test.py
+++ b/backend/kernelCI_app/queries/test.py
@@ -1,6 +1,13 @@
-import datetime
 from typing import Optional
 from kernelCI_app.models import Tests
+from kernelCI_app.typeModels.databases import (
+    Build__ConfigName,
+    Checkout__GitRepositoryBranch,
+    Checkout__GitRepositoryUrl,
+    Origin,
+    Test__Path,
+    Test__StartTime,
+)
 
 
 def get_test_details_data(*, test_id):
@@ -18,7 +25,6 @@ def get_test_details_data(*, test_id):
             "start_time",
             "environment_compatible",
             "output_files",
-            "field_timestamp",
             "build__compiler",
             "build__architecture",
             "build__config_name",
@@ -35,29 +41,26 @@ def get_test_details_data(*, test_id):
 
 def get_test_status_history(
     *,
-    path: str,
-    origin: str,
-    git_repository_url: str,
-    git_repository_branch: str,
+    path: Test__Path,
+    origin: Origin,
+    git_repository_url: Checkout__GitRepositoryUrl,
+    git_repository_branch: Checkout__GitRepositoryBranch,
     platform: Optional[str],
-    current_test_timestamp: datetime,
+    current_test_start_time: Test__StartTime,
+    config_name: Build__ConfigName,
 ):
     query = Tests.objects.filter(
         path=path,
         build__checkout__origin=origin,
         build__checkout__git_repository_url=git_repository_url,
         build__checkout__git_repository_branch=git_repository_branch,
-        field_timestamp__lte=current_test_timestamp,
-    ).values(
-        "field_timestamp",
-        "id",
-        "status",
-        "build__checkout__git_commit_hash",
-    )
+        start_time__lte=current_test_start_time,
+        build__config_name=config_name,
+    ).values("start_time", "id", "status")
 
     if platform is None:
         query = query.filter(environment_misc__platform__isnull=True)
     else:
         query = query.filter(environment_misc__platform=platform)
 
-    return query.order_by("-field_timestamp")[:10]
+    return query.order_by("-start_time")[:10]
--- a/dashboard/src/components/TestDetails/StatusHistoryItem.tsx
+++ b/dashboard/src/components/TestDetails/StatusHistoryItem.tsx
@@ -8,6 +8,8 @@ import {
 
 import { Link } from '@tanstack/react-router';
 
+import { TiArrowSortedDown } from 'react-icons/ti';
+
 import type { TestStatusHistoryItem } from '@/types/tree/TestDetails';
 
 import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
@@ -16,9 +18,11 @@ import { cn } from '@/lib/utils';
 export const StatusHistoryItem = ({
   historyItem,
   iconsClassName,
+  isCurrentTest = false,
 }: {
   historyItem: TestStatusHistoryItem;
   iconsClassName?: string;
+  isCurrentTest?: boolean;
 }): JSX.Element => {
   const statusIcon = useMemo(() => {
     const status = historyItem.status;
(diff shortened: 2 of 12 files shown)


backend/kernelCI_app/queries/test.py
```diff
@@ -35,29 +41,26 @@ def get_test_details_data(*, test_id):
...
+        build__config_name=config_name,
+    ).values("start_time", "id", "status")
 
     if platform is None:
         query = query.filter(environment_misc__platform__isnull=True)
     else:
         query = query.filter(environment_misc__platform=platform)
 
-    return query.order_by("-field_timestamp")[:10]
+    return query.order_by("-start_time")[:10]
```
murilx, line 66: AFAIK `config_name` could also be null

dashboard/src/components/TestDetails/StatusHistoryItem.tsx
```diff
@@ -16,33 +18,47 @@ import { cn } from '@/lib/utils';
...
           <MdOutlineCancel
-            className={cn(iconsClassName, 'text-red text-2xl')}
+            className={cn(
+              iconsClassName,
+              'text-red text-3xl',
+              !isCurrentTest && 'mt-6',
+            )}
           />
         );
       case 'PASS':
```
murilx, line 40: Could you add a constant for 'PASS' and 'FAIL' similar to what we did in the backend? I know there's a lot of places that uses the hardcoded string (and you should not change these other places in this PR) but I believe it would be better if we started transitioning from these hardcoded strings to a constant approach

dashboard/src/components/TestDetails/StatusHistoryItem.tsx
```diff
@@ -53,11 +69,19 @@ export const StatusHistoryItem = ({
           from="/test/$testId/"
           search={s => s}
           state={s => s}
+          className="flex flex-col items-center"
         >
+          {isCurrentTest && <ArrowDown />}
```
murilx, line 74: Maybe an arrow icon similar to what we use in our commit graph would suit better our application. Have you give `md/MdArrowDropDown` or `io/IoMdArrowDropdown` from react icons a try?

dashboard/src/components/TestDetails/StatusHistoryItem.tsx
MarceloRobert: This is not really hardcoded because `status` here is of the type `Status` (from /src/types/database.ts), so typescript won't allow strings other than the correct statuses

murilx: True, completely forgot about that..

backend/kernelCI_app/queries/test.py
MarceloRobert: Django does the treatment for null fields if the value is None, only platform is being treated explicitely because django is translating this comparison to a json, not to a string
https://docs.djangoproject.com/en/5.1/ref/models/querysets/#field-lookups

murilx: perfect!

dashboard/src/components/TestDetails/StatusHistoryItem.tsx
MarceloRobert: I used TiArrowSortedDown now, it looks like those dropdown arrows but it has a bit less margin, making it look bigger than those alternatives

murilx (approved):
Worked well on my tests. Nice job

Francisco2002 (approved):
Nice job!!!!

PR #1077, opened 2025-03-13
feat: send notification when server errors happen

### LINKED ISSUE #1050: Send to Discord any error 500
Let's catch Error 500 as early as possible, send to discord when that have

See if it can be done through a catch all middleware

murilx:
Closes #1050

## Visual Reference
Some examples of this happening
image

## How to test
- Set `DISCORD_WEBHOOK_URL` to the private server channel
- Force a server error (500-599). You can do it just returning a Response with HTTPStatus.INTERNAL_SERVER_ERROR in any view you'd like and check the server to see the error message showing there

### DIFF
--- a/backend/kernelCI_app/middleware/logServerErrorMiddleware.py
+++ b/backend/kernelCI_app/middleware/logServerErrorMiddleware.py
@@ -1,0 +1,51 @@
+import json
+from http import HTTPStatus
+from kernelCI_app.helpers.logger import create_endpoint_notification
+from kernelCI_app.helpers.discordWebhook import send_discord_notification
+
+REASON_MAX_LEN = 500
+
+
+class LogServerErrorMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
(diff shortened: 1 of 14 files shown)


MarceloRobert (commented):
It would be useful if the real reason for the error were shown, like a ValidationError or some other error, maybe even more specific than that

backend/kernelCI_app/middleware/logServerErrorMiddleware.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class LogServerErrorMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if HTTPStatus(response.status_code).is_server_error:
+            reason = response.content.decode()
```
Francisco2002, line 13: The `reason` is like a HTML page when you access the route by a browser

Francisco2002 (approved):
LGTM

PR #1078, opened 2025-03-13
Feat: cron job support

### LINKED ISSUE #1003: Add support for custom cron jobs
Be able to upload scripts through a Pull Request, such that Django will use them for custom functions

MarceloRobert:
Added support to cron jobs by adding the django_crontab app.
To set up cron jobs, edit the `CRON_JOBS` variable in `kernelci/settings.py`. You can run commands or tasks from there.

Added an example task that increments a counter in a separate file.
Added cron job to update the tree files once a week (every Sunday midnight)

### Locally

To start or update the cron jobs, run
`poetry run ./manage.py crontab add`
You don't need to run the backend or server for those cron jobs to work, they will be run in your machine

To stop the cron jobs, run
`poetry run ./manage.py crontab remove`

To list the cron jobs, run
`poetry run ./manage.py crontab show`

### On docker

Start up docker, which will automatically update the cron jobs
Check the cron jobs inside the docker image by running
`sudo docker exec -it dashboard-backend-1 crontab -l`

You can also use the same commands that you were using locally to remove/add/update cron jobs

Closes #1003


docker-compose.yml
```diff
@@ -45,7 +45,7 @@ services:
             - DB_DEFAULT_USER=${DB_DEFAULT_USER:-kernelci}
             - DB_DEFAULT_NAME=${DB_DEFAULT_NAME:-kcidb}
             - DJANGO_SECRET_KEY=${DJANGO_SECRET_KEY}
-            - DEBUG=False
+            - DEBUG=True
```
murilx, line 48: Why are you changing DEBUG to true on docker compose?

backend/.gitignore
```diff
@@ -175,4 +175,7 @@ cython_debug/
 runtime/
 
 # Gmail API token
-gmail_api_token.json
\ No newline at end of file
+gmail_api_token.json
+
+# Example cron task result
+kernelCI_app/counter.txt
```
murilx, line 181: This file is not being created because the example task is commented in the `CRONJOBS` setting (and it should be only a comment in my opinion)

backend/kernelCI_app/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+    try:
+        with open(counter_file, "r") as file:
+            counter = int(file.read().strip())
+    except FileNotFoundError:
+        counter = 0
+
+    counter += 1
+
+    with open(counter_file, "w") as file:
+        file.write(str(counter))
```
murilx, line 17: nit: This appears to be a temporary solution until real tasks are added to this file. I suggest adding a TODO comment to clarify this and mention that it's safe to remove once other tasks are available as examples. Additionally, creating an unnecessary file may introduce unnecessary "noise" into the system, so I think it's not the best example in this situation.

docker-compose.yml
MarceloRobert: Oops, I forgot to remove this from when I was testing the cron jobs on docker, thanks for spotting

backend/kernelCI_app/tasks.py
MarceloRobert: Counting in an external file was the easiest way to make sure that the cron job was running. Since this is just an example for testing and is not meant to be ran in production, I would say we can leave it as is. I will add a comment to remove this function once we have real tasks.

murilx:
Testing here I couldn't verify that the cron job is actually running inside the docker. Trying to run the example task it never created the `counter.txt` file. Could you verify this, please?

murilx (approved):
Worked well on my tests

PR #1082, opened 2025-03-14
fix: remove culprit column and filter by code

### LINKED ISSUE #1049: Remove culprit and filter by code and add Issue for adding culprit again
- [x] Remove culprit column from issue Listing
- [x] Filter to only display issues that the culprit is code

murilx:
Added this issue for future discussion about how to handle culprit:

Closes #1049

## Visual Reference
Before
image

After
image

## How to test
- Go to issue listing and check that you can't find any issue that does not have culprit code


Francisco2002 (approved):
Simple and elegant.
Nice job!!

PR #1092, opened 2025-03-18
Add notifications cron job

padovan:
Deploy the notifications to crontab. The current code has:

`if not is_production_instance():`

so we will be testing them in staging first and then we drop the `not` for production deploy.

### DIFF
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -12,6 +12,7 @@ from urllib.parse import quote_plus
 from django.core.management.base import BaseCommand
 
 from kernelCI_app.constants.general import DEFAULT_ORIGIN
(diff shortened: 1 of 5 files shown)


backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -270,6 +271,14 @@ def generate_issues_summary(service, email_args):
 
     send_email_report(service, report, email_args)
 
+    if not is_production_instance():
```
MarceloRobert, line 274: This is sending the first email report independent if it is running on production instance or not. Just confirming this is the intended behavior.

MarceloRobert (approved):
Seems straightforward, looks good to me

backend/kernelCI_app/management/commands/notifications.py
padovan: good catch!

PR #1105, opened 2025-03-20
Feat: drop origin from issue listing

### LINKED ISSUE #1075: Drop origin selector from Issues
Let's turn it into a column instead. Right now only maestro and redhat are sending them and I believe we will benefit from seeing them together. Later once with have filters with #1074 we can filter per origin.

Francisco2002:
Close #1075 

### **How to test**
- Navigate to issue listing page and see that origin select isn't in the header
- Navigate through the other pages to see that origin select isn't hidden in that
- Run tests

### DIFF
--- a/backend/kernelCI_app/queries/issues.py
+++ b/backend/kernelCI_app/queries/issues.py
@@ -4,6 +4,7 @@ from django.db.utils import ProgrammingError
 from kernelCI_app.helpers.database import dict_fetchall
 from kernelCI_app.helpers.logger import log_message
 import typing_extensions
+from kernelCI_app.models import Issues
 from kernelCI_app.helpers.build import valid_do_not_exist_exception, valid_status_field
 from kernelCI_app.constants.general import SCHEMA_VERSION_ENV
 from datetime import datetime
--- a/backend/kernelCI_app/queries/issues.py
+++ b/backend/kernelCI_app/queries/issues.py
@@ -6,6 +6,8 @@ from kernelCI_app.helpers.logger import log_message
 import typing_extensions
 from kernelCI_app.helpers.build import valid_do_not_exist_exception, valid_status_field
 from kernelCI_app.constants.general import SCHEMA_VERSION_ENV
+from datetime import datetime
+from django.db.models import Q
 
 
 @typing_extensions.deprecated(
@@ -84,3 +86,36 @@ def get_issue_tests(*, issue_id: str, version: int) -> list[dict]:
     with connection.cursor() as cursor:
         cursor.execute(query, params)
         return dict_fetchall(cursor)
+
+
+def get_issue_listing_data(
+    *,
+    interval_date: datetime,
+    culprit_code: bool | None,
+    culprit_harness: bool | None,
+    culprit_tool: bool | None,
+) -> list[dict]:
+    filters = Q(field_timestamp__gte=interval_date)
+
+    culprit_query = Q()
+
+    if culprit_code:
+        culprit_query |= Q(culprit_code=True)
+    if culprit_harness:
+        culprit_query |= Q(culprit_harness=True)
+    if culprit_tool:
+        culprit_query |= Q(culprit_tool=True)
+
+    filters = filters & culprit_query
+
+    issues_records = Issues.objects.values(
+        "id",
+        "field_timestamp",
+        "comment",
+        "version",
+        "origin",
+        "culprit_code",
+        "culprit_harness",
+        "culprit_tool",
+    ).filter(filters)
+    return issues_records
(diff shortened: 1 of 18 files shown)


murilx (approved):
It's much slower but worked well on my tests

MarceloRobert (commented):
I see that @murilx added the culprit_code filter in the frontend, but I believe that with the query being as slow as it got now, it might be better to move that filter to the backend query, because with your current change the backend is trying to get the first_seen value for 297 issues but it will only show the 172 issues with culprit_code == true in the frontend. Filtering in the backend via an optional query parameter and defaulting it to culprit_code in the frontend would lower the request time while keeping the option to get all culprits via the api if needed.

backend/kernelCI_app/queries/issues.py
```diff
@@ -40,3 +41,32 @@ def get_issue_builds(*, issue_id: str, version: int) -> Optional[dict]:
...
+    culprit_tool: bool | None,
+) -> list[dict]:
+    filters = {"field_timestamp__gte": interval_date}
+
+    if culprit_code:
+        filters["culprit_code"] = True
+    if culprit_harness:
+        filters["culprit_harness"] = True
+    if culprit_tool:
+        filters["culprit_tool"] = True
```
murilx, line 60: Maybe something to think in the future, but I believe there should be a way of filtering by `false` and `None` as well instead of just `true`. For example, If I want to see every issue that's not caused by tool, these filters do not provide any way of doing that

backend/kernelCI_app/typeModels/issueListing.py
```diff
@@ -8,14 +10,31 @@
...
+        Field(default=False),
+    ]
+    culprit_harness: Annotated[
+        bool | None,
+        Field(default=False),
+    ]
+    culprit_tool: Annotated[
+        bool | None,
+        Field(default=False),
+    ]
```
murilx, line 30: nit: I believe Annotated shouldn't be necessary here, but that's not a problem

backend/kernelCI_app/unitTests/issues_test.py
murilx, line 1: It would be good to add tests for the new query parameters

dashboard/src/locales/messages/index.ts
```diff
@@ -192,6 +192,8 @@ export const messages = {
     'issueDetails.firstIncidentData': 'First Incident Data',
     'issueDetails.id': 'Issue Id',
     'issueDetails.issueDetails': 'Issue Details',
+    'issueDetails.issueListingInfo':
+      'The issues listed below are caused by the execution code',
```
murilx, line 196: I'm not sure this is better but I'd like to see the word 'culprit' explicitly used here, something like

```suggestion
      'The culprit for all issues listed is code',
```

dashboard/src/types/general.ts
```diff
@@ -262,7 +262,10 @@ export type SearchParamsKeys =
   | 'startTimestampInSeconds'
   | 'endTimestampInSeconds'
   | 'issueVersion'
-  | 'logOpen';
+  | 'logOpen'
+  | 'culpritCode'
+  | 'culpritTool'
+  | 'culpritHarness';
```
murilx, line 268: add minification for these new query parameters

backend/kernelCI_app/queries/issues.py
murilx: @MarceloRobert would like to hear your thoughts on this

dashboard/src/components/IssueTable/IssueTable.tsx
murilx, line 1: By filtering in the backend there's no need to also filter in the frontend, so you can also remove the code that's filtering by culprit_code in the frontend (it's around line 209, I'm not sure if there's a way to mark it in this comment because it was not modified in this PR yet)

backend/kernelCI_app/typeModels/issueListing.py
MarceloRobert: I think `Optional[bool] = False` would be the same

Francisco2002: I will typing as `bool | None`, the same as in the others parts of application

backend/kernelCI_app/queries/issues.py
murilx: That needs some thought later indeed. For now we can keep it this way

murilx (commented):
It's working well, but you should update the requests file related to this change. I believe it's `issue-listing-get.sh`

MarceloRobert (approved):
Worked well on my tests, I also think the request file should be updated, but I'll approve already

backend/requests/issue-listing-get.sh
```diff
@@ -1,4 +1,4 @@
-http 'localhost:8000/api/issue/?origin=maestro&intervalInDays=1'
+http 'localhost:8000/api/issue/?&intervalInDays=1&culpritCode=true'
```
murilx, line 1: This does not necessarily need to come with `culpritCode` but I like the addition

murilx (approved):
Worked well on my tests

PR #1110, opened 2025-03-21
Feat: solidify schema change

### LINKED ISSUE #1109: Solidify schema change
Database schema has been updated to version 5, we can remove unneeded functions and comments and update the env variable for default schema version to 5.

- change default schema version to 5
- review queries where django lazy load was not triggering a schema change (such as build tests) - adding a `list(results)` is enough to trigger it
- return the correct data after changing the schema version instead of just returning an error
- refactor duplicate functions to be a single one with just the necessary areas changed
- remove temporary deprecated tags
- remove temporary error message about schema change

MarceloRobert:
## Changes
- Adds schema-version file instead of environment variable, so it can be shared between processes in the production docker instance
- Makes queries with db schema version 4 update schema and return correct data instead of just returning an error
- Refactors queries with build valid/status versions to be a single function
- Updates default db schema version to 5 and removes temporary error messages
  - Also updates some endpoints to only return a `Response`, removing the need for `create_error_response`

## How to test
To test the update from db schema version 4 to 5, change the file "schema-version.yaml" to set the variable as '4' and check endpoints which can change the db schema
Also test running on docker
Check if the updated queries are still returning correct data

Closes #1109

### DIFF
--- a/backend/kernelCI_app/helpers/environment.py
+++ b/backend/kernelCI_app/helpers/environment.py
@@ -6,7 +6,7 @@ from kernelCI_app.constants.general import (
 )
 from kernelCI_app.helpers.logger import log_message
 
-DEFAULT_SCHEMA_VERSION = "4"
+DEFAULT_SCHEMA_VERSION = "5"
 
 
 def get_yaml_data(*, file: str):
--- a/backend/kernelCI_app/helpers/environment.py
+++ b/backend/kernelCI_app/helpers/environment.py
@@ -1,0 +1,44 @@
+import yaml
+
+from kernelCI_app.constants.general import (
+    SCHEMA_VERSION_ENV_FILE,
+    SCHEMA_VERSION_ENV_NAME,
+)
+from kernelCI_app.helpers.logger import log_message
+
+DEFAULT_SCHEMA_VERSION = "4"
+
+
+def get_yaml_data(*, file: str):
+    try:
+        with open(file, "r") as f:
+            data = yaml.safe_load(f)
+            return data
+    except FileNotFoundError as e:
+        log_message(f"File not found at {file}\n Error: {e}")
+        return None
+    except yaml.YAMLError as e:
+        log_message(f"Error parsing YAML: {e}")
+        return None
+
+
+def set_yaml_data(*, file: str, data) -> None:
+    try:
+        with open(file, "w") as f:
+            yaml.safe_dump(data, f, default_flow_style=False)
+    except (yaml.YAMLError, IOError) as e:
+        log_message(f"Error writing YAML: {e}")
+
+
+def get_schema_version() -> str:
+    data = get_yaml_data(file=SCHEMA_VERSION_ENV_FILE)
+    schema_version = data.get(SCHEMA_VERSION_ENV_NAME, DEFAULT_SCHEMA_VERSION)
+    return schema_version
+
+
+def set_schema_version(*, version: str) -> None:
+    if version is not None and version != "":
+        set_yaml_data(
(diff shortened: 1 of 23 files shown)


backend/kernelCI_app/queries/hardware.py
```diff
@@ -422,8 +306,8 @@ def query_records(
 
             return dict_fetchall(cursor)
     except ProgrammingError as e:
-        if valid_do_not_exist_exception(e):
-            os.environ[SCHEMA_VERSION_ENV] = "5"
+        if is_valid_does_not_exist_exception(e):
+            set_schema_version("5")
```
murilx, line 310: won't you call `query_records` again in this case as you've been doing to the other queries?

MarceloRobert: I forgot that one

murilx (approved):
Worked well on my tests

backend/kernelCI_app/queries/hardware.py
Francisco2002:
```suggestion
        if is_valid_does_not_exist_exception(e):
            set_schema_version(version="5")
```

backend/kernelCI_app/queries/issues.py
```diff
@@ -48,9 +46,10 @@ def get_issue_builds(*, issue_id: str, version: int) -> list[dict]:
             cursor.execute(query, params)
             return dict_fetchall(cursor)
     except ProgrammingError as e:
-        if valid_do_not_exist_exception(e):
-            os.environ[SCHEMA_VERSION_ENV] = "5"
+        if is_valid_does_not_exist_exception(e):
+            set_schema_version("5")
```
Francisco2002, line 50:
```suggestion
            set_schema_version(version="5")
```

backend/kernelCI_app/queries/tree.py
```diff
@@ -398,18 +279,20 @@ def get_tree_details_data(
                 rows = cursor.fetchall()
                 set_query_cache(cache_key, params, rows)
         except ProgrammingError as e:
-            if valid_do_not_exist_exception(e):
-                os.environ[SCHEMA_VERSION_ENV] = "5"
+            if is_valid_does_not_exist_exception(e):
+                set_schema_version("5")
```
Francisco2002, line 283:
```suggestion
                set_schema_version(version="5")
```

backend/kernelCI_app/queries/tree.py
```diff
@@ -512,8 +395,14 @@ def get_tree_commit_history(
             )
             return cursor.fetchall()
     except ProgrammingError as e:
-        if valid_do_not_exist_exception(e):
-            os.environ[SCHEMA_VERSION_ENV] = "5"
+        if is_valid_does_not_exist_exception(e):
+            set_schema_version("5")
```
Francisco2002, line 399:
```suggestion
            set_schema_version(version="5")
```

backend/kernelCI_app/helpers/environment.py
```diff
@@ -0,0 +1,42 @@
...
+
+def set_yaml_data(*, file: str, data) -> None:
+    try:
+        with open(file, "w") as f:
+            yaml.safe_dump(data, f, default_flow_style=False)
+    except (yaml.YAMLError, IOError) as e:
+        log_message(f"Error writing YAML: {e}")
+
+
+def get_schema_version(default="4") -> str:
```
Francisco2002, line 31: I think that the default value should be a constant value, not a parameter

Francisco2002 (approved):
Nice job!!!!

PR #1111, opened 2025-03-21
feat: add cache database

### LINKED ISSUE #1094: Store current tree listing data to a sqlite file
The sqlite database should contain the basic information from the tree listing, and have a Django model that describes the table, being able to be expanded with data that is not necessarily in KCIDB

murilx:
Closes #1094 

- Call the TreeView endpoint and check if the data get by the request is also stored in the `cache.sqlite3` file 
- Check if the same behavior persists in the docker

**IMPORTANT:** In this task I took security measures to ensure that we won't add anything to the KCIDB database. Please review them with extra caution to make sure that's the case


backend/kernelCI_cache/checkouts.py
```diff
@@ -0,0 +1,126 @@
...
+from datetime import datetime, timezone, timedelta
+from kernelCI_app.typeModels.treeListing import Checkout
+from kernelCI_app.typeModels.databases import Origin
+from kernelCI_app.helpers.logger import log_message
+from kernelCI_cache.models import Checkouts
+
+
+# This function could be called in another thread if we face performance issues
+# delay in endpoints in the future
+def populate_cache_db(data: list[Checkout], origin: Origin) -> None:
```
MarceloRobert, line 11: You are not receiving a `list[Checkout]` here, you are receiving (and using) a `list[dict]`, hence the use of ["property"] instead of .property

backend/kernelCI_cache/checkouts.py
```diff
@@ -0,0 +1,126 @@
...
+            cache_data.test_miss = checkout["test_status"]["miss"]
+            cache_data.test_skip = checkout["test_status"]["skip"]
+            cache_data.test_error = checkout["test_status"]["error"]
+            cache_data.test_null = checkout["test_status"]["null"]
+            cache_data.save(
+                using="cache",
+                update_fields=[
+                    "start_time",
+                    "git_commit_tags",
+                    "build_pass",
```
MarceloRobert, line 57: I think there might be a way to just say "update all fields" instead of selecting all fields and updating them

backend/kernelCI_cache/checkouts.py
```diff
@@ -0,0 +1,126 @@
...
+                    "test_error",
+                    "test_null",
+                ],
+            )
+
+        except Checkouts.DoesNotExist:
+            tz_offset = timezone(timedelta(hours=-3))
+            timestamp = datetime.now(tz_offset)
+            timestamp_str = timestamp.strftime("%Y-%m-%d %H:%M:%S.%f")
+            timestamp_str = timestamp_str[:-3] + " " + timestamp.strftime("%z")
```
MarceloRobert, line 85: Make this a separate function to convert from one format to the other

backend/kernelCI_cache/models.py
```diff
@@ -0,0 +1,52 @@
+from django.db import models
+
+
+class Checkouts(models.Model):
```
MarceloRobert, line 4: Just a question, but would it be better to call this model something like "CachedCheckouts" so it doesn't get confused with the original Checkouts? Specially when using automatic imports

backend/kernelCI_cache/checkouts.py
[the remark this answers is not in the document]
murilx: Thought that having it in the "checkouts" file would be enough, but you are right

backend/kernelCI_cache/checkouts.py
murilx: I was using update_fields to force an update (https://docs.djangoproject.com/en/5.1/ref/models/instances/#specifying-which-fields-to-save) but on further search, I believe that by changing the values directly on a model instance will ensure that django will do an update even if I don't pass this parameter. I'll remove it

backend/kernelCI_cache/models.py
murilx: I thought that maybe the imports would be enough to differentiate between them, and that having `CacheCheckouts` inside the `kernelCI_cache` app would be unnecessary redundancy, but I agree that it could cause some confusion so I'll change the model name

MarceloRobert (approved):
Seems to be working really well on my tests, the file is shared between docker workers

Francisco2002 (approved):
Nice job!!!!!

PR #1116, opened 2025-03-24
fix: disable refetch on windows focus

### LINKED ISSUE #1066: [frontend] empty tables are reloaded when switching tabs in the browser
I usually open multiple tabs in my browser when I'm using the dashboard.
Every time I switch to a tab with an empty table in it, e.g. build-detail's "test results" and "issues", the table is reloaded. Sometimes, a single tab switch triggers two or three requests to the API.

The expected behavior would be to lazy-load the tables once, and switching tabs shouldn't trigger a "table refresh".

I'm quite rusted on my react, so I don't have any idea which hook is contributing to this "reload on focus".
From what I saw, `useQuery` doesn't cache by default (`staleTime=0` https://tanstack.com/query/v4/docs/framework/react/guides/caching), however it might be as simple as a missing hook wrapping it, I honestly don't remember much.

murilx:
Disabled refetch on window focus for the queries where this was a problem:
- useBuildTests
- useBuildIssues
- useTestIssues

Closes #1066

## How to test
Open the a build or test details, wait until the issue section and test results section (for build details) are loaded, change tabs and go back to the details page. The section should not start loading again like it does on prod. You can test the same thing for the issue section in builds/boots/tests tables on any page (tree details, for example)

### DIFF
--- a/dashboard/src/api/buildDetails.ts
+++ b/dashboard/src/api/buildDetails.ts
@@ -2,7 +2,7 @@ import type { UseQueryResult } from '@tanstack/react-query';
 import { useQuery } from '@tanstack/react-query';
 
 import type { TBuildDetails } from '@/types/tree/BuildDetails';
-
+import type { ApiUseQueryOptions } from '@/types/api';
 import type { TIssue } from '@/types/issues';
 
 import { RequestData } from './commonRequest';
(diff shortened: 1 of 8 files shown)


MarceloRobert (commented):
I would say it would also be useful if you added those conditions to hardwareListing (because most origins don't have hardware), to issueDetailsTests and Builds (because issueDetails makes both requests but only uses one), and on buildDetails and testDetails too, because they are used in the log sheets

MarceloRobert (approved):
LGTM

PR #1117, opened 2025-03-24
Feat: show checkouts without builds

### LINKED ISSUE #1036: Better message for Tree Details with No Builds
if a checkout has no build it claims that the tree does not exists

(This is a little rare to reproduce because it will only occurs in checkouts with no builds, search for that in the database if you want to reproduce)

### LINKED ISSUE #1115: Show commit history graph even if checkout has no builds
The tree details page won't show the commit history if the latest checkout has no builds. This should not happen as it won't be possible to navigate to older checkouts (that possibly have builds and tests) without changing the url.
The commit history endpoint should come independently of the summary or full endpoints

MarceloRobert:
Closes #1036 
Closes #1115


dashboard/src/components/DetailsPages/SectionError.tsx
```diff
@@ -69,9 +73,10 @@ const SectionError = ({
       <p className="text-2xl font-semibold">
         <FormattedMessage id={message} />
       </p>
-      {errorMessage && errorStatusCode !== HttpStatusCode.Ok && (
+      {((errorMessage && errorStatusCode !== HttpStatusCode.Ok) ||
+        forceErrorMessageUse) && (
         <p>
-          {globalErrorMessage[variant].label}: {errorMessage}
+          {globalErrorMessage[variant].label}: {errorResponseMessage}
```
murilx, line 79: Shouldn't this be something like

```suggestion
          {globalErrorMessage[variant].label}: {errorResponseMessage ?? errorMessage}
```

in the case `errorMessage` is sent as prop to this component?

backend/kernelCI_app/views/treeDetailsSummaryView.py
```diff
@@ -189,10 +191,23 @@ def get(self, request, commit_hash: str | None) -> Response:
...
         if len(rows) == 0:
             return create_api_error_response(
-                error_message="Tree not found",
+                error_message="Tree checkout not found",
                 status_code=HTTPStatus.OK,
             )
 
+        if len(rows) == 1:
+            row_data = get_current_row_data(rows[0])
+            if row_data["build_id"] is None:
```
murilx, line 200: since data comes asynchronously we may have checkouts with boots/tests but with no builds and if that's the case they should be shown in the dashboard instead of a error message

backend/kernelCI_app/views/treeDetailsView.py
```diff
@@ -201,10 +203,23 @@ def get(self, request, commit_hash: str | None):
...
         if len(rows) == 0:
             return create_api_error_response(
-                error_message="Tree not found",
+                error_message="Tree checkout not found",
                 status_code=HTTPStatus.OK,
             )
 
+        if len(rows) == 1:
+            row_data = get_current_row_data(current_row=rows[0])
+            if row_data["build_id"] is None:
```
murilx, line 212: since data comes asynchronously we may have checkouts with boots/tests but with no builds and if that's the case they should be shown in the dashboard instead of a error message

murilx (commented):
Seems to be working fine, but I believe this logic might be incorrect

MarceloRobert:
> Seems to be working fine, but I believe this logic might be incorrect

Because a test is not directly related to a checkout, if the checkout doesn't have any build then that means there's no way to tell which test belongs to that checkout. In other words, if a checkout doesn't have any builds, it won't have any related tests.

murilx (approved):
There's still this placeholderData bug that we need to tackle (please make an issue for that if you haven't already), but otherwise it's working well. Nice job

Francisco2002 (approved):
LGTM!
Some comments

PR #1118, opened 2025-03-25
Test/improve tests execution

### LINKED ISSUE #1103: Improve unit tests
Our tests are taking a long time to complete, there are some improvements to be made:

1. Add retry after request error.
  Sometimes the database proxy loses connection unexpectedly, this makes the test automatically fail even though it is not our code's fault and could scrap an entire test run when a simple retry could fix the issue

2. Use endpoints with faster response.
  Some tests are trying to reach endpoints with a lot of data and that makes the request slower, even though we could validate the test with less data

3. Not use unnecessary tests.
  Individual tests should probe different scenarios, make sure that we don't have unnecessary tests doing the same kind of validation and adding unnecessary slowness

Other approaches could be used to further lower the time that takes to run all tests.

Francisco2002:
Close #1103 

### **How to test**

- Run the command `time poetry run pytest`
- Run the command `time poetry run pytest -n 4 --dist loadscope --reruns 4`
- Compare the executino time


backend/README.md
```diff
@@ -35,9 +35,10 @@ poetry run python3 manage.py runserver
 The backend includes unit tests covering some parts of the source code. To run the tests, use the following command:
 
 ```sh
-poetry run pytest
+poetry run pytest -n 4 --dist loadscope --reruns 4
 ```
 
+This command runs the tests, splitted into 4 processes. Each process runs a test file. If a test fail, for some reason, it retry, at most 4 times.
```
murilx, line 41:
```suggestion
This command runs the tests, splitted into 4 processes. Each process runs a test file. If a test fails, for some reason, it's retried up to 4 times.
```

murilx:
Maybe we could benefit from adding these flags as default options, so that we don't have to remember to add them every time we run our tests. https://docs.pytest.org/en/7.1.x/reference/reference.html#confval-addopts

murilx:
Also, why are you recommending `-n 4` instead of `-n auto` for example? I'd like to point out that these questions only matter if we really add these flags to our "default", if not we can each select the flags we prefer...

Francisco2002:
@murilx the `-n auto` flag will create a number of process based on the number of cores in the machine. I put this value to give a most significant value to this parameter

backend/pyproject.toml
```diff
@@ -31,9 +31,12 @@ flake8 = "^7.1.0"
 pytest = "^8.3.4"
 pytest-django = "^4.10.0"
 pep8-naming = "^0.14.1"
+pytest-xdist = {extras = ["psutil"], version = "^3.6.1"}
+pytest-rerunfailures = "^15.0"
 
 [tool.pytest.ini_options]
 DJANGO_SETTINGS_MODULE = "kernelCI.settings"
+addopts = "-n 4 --dist loadscope --reruns 4 --maxfail=1"
```
murilx, line 39: use can use `-x` instead of `--maxfail=1`

backend/README.md
```diff
@@ -35,9 +35,10 @@ poetry run python3 manage.py runserver
 The backend includes unit tests covering some parts of the source code. To run the tests, use the following command:
 
 ```sh
-poetry run pytest
+poetry run pytest -n 4 --dist loadscope --reruns 4
```
murilx, line 38: remove these flags that are now default from the example so anyone reading understands that it can only run `poetry run pytest` but also add somewhere an explanation that we have these default flags and that's why it behaves like that

backend/README.md
```diff
@@ -55,7 +56,7 @@ subset of tests cases is run to save time, especially during push. To run all te
 `--run-all` flag:
 
 ```sh
-poetry run pytest --run-all
+poetry run pytest -n 4 --dist loadscope --reruns 4 --run-all
```
murilx, line 59: ditto about removing the flags

backend/README.md
```diff
@@ -38,6 +38,7 @@ The backend includes unit tests covering some parts of the source code. To run t
 poetry run pytest
 ```
 
+This command runs the tests, splitted into 4 processes. Each process runs a test file. If a test fails, for some reason, it's retried up to 4 times.
```
murilx, line 41:
```suggestion
This command runs the tests, splitted into 4 processes. Each process runs a test file. If a test fails, for some reason, it's retried up to 4 times. This behavior is due to the flags we are passing to `addopts` in the `pyproject.toml` settings file.
```

murilx (approved):
Very impressed by the how much time we reduced from our test suite and how simple it was! Nice job

backend/README.md
```diff
@@ -38,6 +38,8 @@ The backend includes unit tests covering some parts of the source code. To run t
...
```
MarceloRobert, line 41:
```suggestion
This command runs the tests, splitted into 4 processes. Each process runs a test file. If a test fails for any reason, it's retried up to 4 times.
```
Saying "for some reason" between commas sounds like we don't know why the test is retried

MarceloRobert (approved):
Just tested it and it did run faster, I still think the docs should have the change I commented about but I'll pre-approve

PR #1120, opened 2025-03-25
fix: update error messages for empty data

### LINKED ISSUE #1119: Finish changing error messages for 200 status "error"
There are other instances of the backend sending an error with status 200 that aren't yet treated correctly in the frontend. This includes but is not limited to build details and test details

murilx:
Closes #1119

**NOTE:** I did not change `CommitNavigationGraph` and `TestStatusHistory` messages because if the page is loaded them they should also return with no errors (since they should return at least the data for the current page)


Francisco2002 (approved):
LGTM

Issue #1132, opened 2025-03-27
Improve tree details queries


MarceloRobert:
Some trees, specially ones with a lot of tests, are taking too long to finish (>30s), such as this [tuxsuite tree in staging](https://staging.dashboard.kernelci.org:9000/tree/bddc6e9322072ac3aa15bd972c5bcbf9f447d246?o=tuxsuite&p=t&ti%7Cc=v6.6.83-167-gbddc6e932207&ti%7Cch=bddc6e9322072ac3aa15bd972c5bcbf9f447d246&ti%7Cgb=&ti%7Cgu=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fstable%2Flinux-stable-rc.git)

There should be improvements in the query or the processing of data to lower that time.

Issue #1131, opened 2025-03-27
Improve test status history query


MarceloRobert:
Some pages with tests that have generic names are taking about 40 seconds to load the status history, such as this [ltp test in staging](https://staging.dashboard.kernelci.org:9000/test/maestro:67ce429218018371957d9e59)

Other test details can take much less time, such as this [kselftest.seccomp.seccomp_seccomp_benchmark test in staging](https://staging.dashboard.kernelci.org:9000/test/maestro:67ce445218018371957db690).

Query for the status history:
```
SELECT
    "tests"."start_time",
    "tests"."id",
    "tests"."status"
FROM
    "tests"
    INNER JOIN "builds" ON ("tests"."build_id" = "builds"."id")
    INNER JOIN "checkouts" ON ("builds"."checkout_id" = "checkouts"."id")
WHERE
    (
        "checkouts"."git_repository_branch" = 'master'
        AND "checkouts"."git_repository_url" = 'https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git'
        AND "checkouts"."origin" = 'maestro'
        AND "builds"."config_name" = 'multi_v7_defconfig'
        AND "tests"."path" = 'ltp'
        AND "tests"."start_time" <= '2025-03-10 01:38:26.137000+00:00'::TIMESTAMPTZ
        AND ("tests"."environment_misc" -> 'platform') = '"beaglebone-black"'::JSONB
    )
ORDER BY
    "tests"."start_time" DESC
LIMIT
    10
```

PR #1133, opened 2025-03-27
Feat: docker performance improvements

### LINKED ISSUE #1029: Investigate performance crashes on big queries
tuxsuite origin has some 100k tests tree

It was reported that in some cases those are failing

Investigate to see if we can increase the tolerance for slow queries

eg:
https://dashboard.kernelci.org/api/tree/c0eb65494e59d9834af7cbad983629e9017b25a1/summary?origin=tuxsuite&git_url=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fnext%2Flinux-next.git&git_branch=master

MarceloRobert:
## Changes
- Adds a timeout to the dockerfile so that the default (30, as described by [gunicorn's documentation](https://docs.gunicorn.org/en/stable/settings.html#timeout)) is overwritten by a larger value (120) to allow endpoints to be loading for longer, such as this [staging test](https://staging.dashboard.kernelci.org:9000/test/maestro:67ce429218018371957d9e59) that takes about 40s to load the status history, or this [staging tree](https://staging.dashboard.kernelci.org:9000/tree/bddc6e9322072ac3aa15bd972c5bcbf9f447d246?o=tuxsuite&p=t&ti%7Cc=v6.6.83-167-gbddc6e932207&ti%7Cch=bddc6e9322072ac3aa15bd972c5bcbf9f447d246&ti%7Cgb=&ti%7Cgu=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Fstable%2Flinux-stable-rc.git) that takes about 80s to get the full data.
- Piggybacking is a small fix to a log message that doesn't need to be sent when misc is null, and a small code style problem

## How to test
Check the pages linked above, and also access them through the localhost _while running on docker_, this effect will not be seen if you run without docker because the changes only affect the docker's `gunicorn`.
Also check the api pages that they make requests to, by looking into the network tab of the page
You can try changing the timeout parameter and see that the workers will complain about it after the given time

Closes #1029

PS.: Understandably, this does not solve the issue of heavy queries, but it does prevent the error from happening until we have improvements to the query or the processing of data in the endpoints, which will be tackled in #1131 and #1132

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -43,5 +43,5 @@ EXPOSE 8000
 
 VOLUME ["/backend/data"]
 
-CMD ["--workers=5", "--forwarded-allow-ips=*", "--bind=0.0.0.0:8000"]
+CMD ["--workers=5", "--forwarded-allow-ips=*", "--bind=0.0.0.0:8000", "--timeout=240"]
 ENTRYPOINT ["/entrypoint.sh"]
--- a/backend/kernelCI_app/typeModels/commonDetails.py
+++ b/backend/kernelCI_app/typeModels/commonDetails.py
@@ -79,8 +79,10 @@ class BuildHistoryItem(BaseModel):
             return json.loads(value)
         elif isinstance(value, dict):
             return value
+        elif value is None:
+            return None
         else:
-            log_message("Invalid misc for BuildHistoryItem")
+            log_message(f"Invalid misc for BuildHistoryItem;\nType: {type(value)}")
 
     @field_validator("status", mode="before")
     @classmethod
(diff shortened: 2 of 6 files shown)


murilx (approved):
There's still occasionally some timeouts but I understand the reasoning for not increasing the threshold any further. Great start in enabling heavier pages to load on the dashboard

murilx (approved):
Reapproving after the new additions. Really glad you removed that message while the components load

PR #1134, opened 2025-03-27
fix: verify cache before making a new request

### LINKED ISSUE #1090: filtering the first time in a tree details cause a lot of requests
When first entering a tree details page (maybe hardware details also has this problem, this should be checked) and first filtering something by clicking in a card (should check if filtering by modal causes the same issue) it sends a lot of requests. This does not seem very troublesome because page is cached, but it's still unexpected behavior that should be corrected

Image

murilx:
Closes #1090

## How to test
- Enter a tree details with the network tab of inspect open
- Filter for anything in the page and check that it's doing only one request instead of many


MarceloRobert (approved):
Nice, it's much better to not make all those requests

PR #1135, opened 2025-03-28
Feat: auto-migrate sqlite cache db

### LINKED ISSUE #1127: Auto migrate sqlite db
Ignore the sqlite cache from git and make docker automatically migrate the sqlite on startup

MarceloRobert:
## Changes
- Ignores the sqlite file from git;
- Renames and update initial migration;
- Adds automatic migration to docker entrypoint;
- Adds volume use to persist data even if docker stops running;

## How to test
This PR will only affect the sqlite, which is not currently used in the dashboard.
To see the changes run the migrate-cache-db.sh and go to the tree listing page with different origins, the sqlite should update normally.
When running on docker, you should see a message about the migrations when the backend starts

The persistent sqlite from the volume will be stored in docker's volume path, usually at `~/var/lib/docker/volumes/dashboard_db-data/_data`

Closes #1127


murilx (approved):
Nice job

PR #1138, opened 2025-03-31
ci: add action to run tests of backend

### LINKED ISSUE #1089: Use unit tests in github
Currently unit tests are being run pre-push, but they should be running in the github ci so they don't block other things

Francisco2002:
Close #1089


backend/Dockerfile
```diff
@@ -25,7 +25,8 @@ COPY . /backend
 COPY ./utils/docker/entrypoint.sh /entrypoint.sh
 WORKDIR /backend
 
-RUN poetry install --no-interaction --only main
+# RUN poetry install --no-interaction --only main
+RUN poetry install --no-interaction --only main,dev
```
MarceloRobert, line 29: Maybe you don't need all dev dependencies here, since this will be going into the same dockerfile that runs on staging and production anyway, you could add the packages you need for testing on main dependencies

.github/workflows/ci.yaml
```diff
@@ -134,6 +134,42 @@ jobs:
...
+          chmod 777 application_default_credentials.json
+          mkdir -p backend/runtime/secrets
+          echo "${{ secrets.PG_PASSWORD }}" > backend/runtime/secrets/postgres_password_secret
+          echo "DB_DEFAULT_USER=${{ secrets.PG_USERNAME }}" >> $GITHUB_ENV
+          echo "DJANGO_SECRET_KEY=$(openssl rand -base64 22)" >> $GITHUB_ENV
+          echo "DB_DEFAULT_NAME=kcidb" >> $GITHUB_ENV
+
+      - name: Run database proxy and backend docker services
+        run: |
+          docker compose up proxy -d --build
```
MarceloRobert, line 149: Are you running both the database proxy and backend services with this single command? If not, either update the name or the command.

MarceloRobert: Also, isn't there a difference between the `cloudsql-proxy` and the `proxy` services in the docker-compose file?

MarceloRobert (commented):
Seems alright, but there are some comments in the code that are maybe unnecessary

.github/workflows/ci.yaml
murilx: `proxy` service is the one needed to run the backend and server correctly. Also, `proxy` has both `backend` and `collect_static` as dependencies, so I don't need to specify anything other than this service to run everything (since `backend` is a dependency it'll be run and since `cloudsql-proxy` is a dependency of `backed` it'll also run)

backend/Dockerfile
murilx: There's no more need to install dev dependencies in this container because I was able to use pytest outside of the container to run the tests

MarceloRobert (approved):
Awesome that it is now running on github, great job!
Only the message for when there is no backend connection that is quite long, but it's good that its giving an error instead of just skipping tests

PR #1139, opened 2025-03-31
feat(test-details): add link to hardware details

### LINKED ISSUE #960: add missing links to tests and build details
Currently test details has a link to the tree details it's from. Build details should have a link to its tree as well. Also, add links to hardware details of both of them

murilx:
Build has no direct relation with hardware so that's why I didn't add any link to it as well. Even 'platform' that's inside the `misc` field in the `builds` table has no direct relation with the available hardware details pages, since we only consider platform and compatibles of tests in this case

Closes #960

## How to test
- Enter a page with environment compatible: http://localhost:5173/test/maestro%3A67ea51dc67593f2aa0417959
- Enter a page with platform and no environment compatible: http://localhost:5173/test/maestro%3A67e9d6b667593f2aa0405adb
- Enter a page with none: http://localhost:5173/test/redhat%3A1739009469-aarch64-kernel-automotive_upt_aws_5?o=redhat

### DIFF
--- a/dashboard/src/components/TestDetails/TestDetails.tsx
+++ b/dashboard/src/components/TestDetails/TestDetails.tsx
@@ -1,4 +1,5 @@
 import { FormattedMessage, useIntl } from 'react-intl';
+import { roundToNearestMinutes } from 'date-fns';
 
 import { Fragment, useCallback, useMemo, useState } from 'react';
 import type { Dispatch, SetStateAction, JSX } from 'react';
@@ -15,7 +16,12 @@ import { MdChevronRight } from 'react-icons/md';
 
 import type { UseQueryResult } from '@tanstack/react-query';
 
-import { shouldTruncate, truncateBigText, valueOrEmpty } from '@/lib/string';
+import {
+  EMPTY_VALUE,
+  shouldTruncate,
+  truncateBigText,
+  valueOrEmpty,
+} from '@/lib/string';
 import type { TTestDetails, TestStatusHistory } from '@/types/tree/TestDetails';
 import { Sheet } from '@/components/Sheet';
 import {
@@ -59,6 +65,9 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@/components/Tooltip';
 
 import MemoizedLinkItem from '@/components/DetailsLink';
 import { processLogData } from '@/hooks/useLogData';
+
+import { dateObjectToTimestampInSeconds, daysToSeconds } from '@/utils/date';
+import { REDUCED_TIME_SEARCH } from '@/utils/constants/general';
 
 import { StatusHistoryItem } from './StatusHistoryItem';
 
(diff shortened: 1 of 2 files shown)


dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -138,6 +141,49 @@ const TestDetailsSections = ({
     test.tree_name,
   ]);
 
+  const hardwareDetailsLink = useMemo(() => {
+    // eslint-disable-next-line no-magic-numbers
+    const LAST_7_DAYS = 7 * 24 * 60 * 60;
```
MarceloRobert, line 146: - This could be a const in another file
- The variable doesn't exactly relate to the _last_ 7 days, but rather to 7 days in seconds
- The default time interval for hardware is 5 days, not 7

MarceloRobert (commented):
I think it would be better if each hardware or platform would direct to the respective page, so the "hardware" field would direct to the platform, if it exists, or to the first compatible. And the compatibles field would direct to each compatible page for each string, if that's possible

murilx:
> I think it would be better if each hardware or platform would direct to the respective page, so the "hardware" field would direct to the platform, if it exists, or to the first compatible. And the compatibles field would direct to each compatible page for each string, if that's possible

I can't think of a way to guarantee that a hardware details page for a platform exists if the same page has a non null environment_compatible. That is because the logic for a hardware details page is environment_compatible first. So, if for example every test for the platform `A` also has all valid environment_compatibles, than a page for `A` won't exist

murilx:
I did misremember. This logic I described applies to shown hardware in hardware listing but there still exist pages for all test platforms

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -138,6 +142,76 @@ const TestDetailsSections = ({
...
+    hardware,
+    formatMessage,
+    startTimestampInSeconds,
+    endTimestampInSeconds,
+  ]);
+
+  const compatiblesLink = useMemo(() => {
+    if (!test.environment_compatible) {
+      return <span>{valueOrEmpty(test.environment_compatible, '-')}</span>;
+    }
```
MarceloRobert, line 184: if test.environment_compatible doesn't exist, you don't need to pass it to valueOrEmpty

dashboard/src/components/TestDetails/TestDetails.tsx
```diff
@@ -138,6 +142,76 @@ const TestDetailsSections = ({
...
+
+    return (
+      <>
+        {test.environment_compatible.map(
+          (compatible: string): JSX.Element => (
```
MarceloRobert, line 189: IMO it would be better to leave them in a single line
```suggestion
    return (
      <div className="flex flex-wrap gap2">
        {test.environment_compatible.map(
          (compatible: string): JSX.Element => (
```

dashboard/src/components/TestDetails/TestDetails.tsx
murilx: I wanted to use the default value of 'valueOrEmpty' function, so that if we - for some reason - change it in the future, we won't need to change it manually here as well. I did make it wrong though by passing '-' directly to the function

MarceloRobert (approved):
LGTM

PR #1143, opened 2025-03-31
fix: share backend cache between workers

### LINKED ISSUE #1141: Verify discord notification cooldown
The discord notification should only show the same message every x seconds (600 for now), but it seems to not be using this cooldown correctly

murilx:
Uses a memcached server instead of local memory

Closes #1141

- Run the docker and try to throw an error that will cause a discord message. Note that they don't repeat
- Make sure to run all of our tests
- Try to open pages with large amounts of data (e.g. one tuxsuite tree details page) and check that it's working properly


MarceloRobert (approved):
Great, it's working well, just took me a bit to take a grasp on redis at first

PR #1145, opened 2025-03-31
fix: disable retry if status is 200

### LINKED ISSUE #1108: Requests with error and status 200 shouldn't retry
In a lot endpoints we are returning status 200 when there is no data (for example, no tests for a given build in the endpoint /api/build/<build_id>/tests).
Those requests are currently raising an error because there is no data, triggering a new request.
Requests may not retry when the status is 200.

murilx:
Creates a custom function to handle any retries, checking if they sent a status 200 and only disallowing retry if so

Closes #1108

## How to test
- Check that any "200 error" does not retry
- Check that any other situation (actual errors) has retries

### DIFF
--- a/dashboard/src/api/issueDetails.ts
+++ b/dashboard/src/api/issueDetails.ts
@@ -3,11 +3,7 @@ import { useQuery } from '@tanstack/react-query';
 
 import type { TIssueDetails } from '@/types/issueDetails';
 
-import type {
-  BuildsTableBuild,
-  TErrorWithStatus,
-  TestHistory,
-} from '@/types/general';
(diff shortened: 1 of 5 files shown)


dashboard/src/api/issueDetails.ts
MarceloRobert, line 1: For these endpoints I think you can simply remove the retry here, I think the purpose of them was to avoid all those retrys before having this new function

murilx: Yeah, I believe so. The other hook (useLogFiles) does not seem to be used in the application anymore

MarceloRobert (approved):
LGTM, it feels much faster without those unnecessary retries

PR #1146, opened 2025-03-31
Feat: update checkout cache

### LINKED ISSUE #1128: Update checkout cache
Make a separate task that will be run on a cron job, which will:
- select the last checkout of each tree between now and 90 days (similar to treeViewFast)
- select all data from sqlite cache
- iterate over the kcidb results, check if they are stable, check if they already exist in cache; if exists and are stable, continue; if exists and unstable, add to update_list; if doesn't exist, add to update_list
- select from kcidb every checkout that was listed for update, with the counts (similar to treeView)
- iterate over this kcidb result, define if they are unstable, check if they exist in sqlite; if exists, update sqlite; if doesn't exist, add to sqlite

The sqlite will now also store checkout_id (not as a primary key) and a new "unstable" column, which should be boolean

MarceloRobert:
Closes #1128


backend/kernelCI/settings.py
```diff
@@ -132,8 +132,7 @@ def get_json_env_var(name, default):
 # poetry run ./manage.py crontab arg
 # where "arg" is add, remove or show
 CRONJOBS = [
-    # Example for running tasks that are not commands, uncomment to run for testing
-    # ('* * * * *', 'kernelCI_app.tasks.example_task'),
+    ("0 * * * *", "kernelCI_app.tasks.update_checkout_cache"),
```
murilx, line 135: Just to clarify. This means it's going to update every hour, is that correct?

backend/kernelCI_app/queries/build.py
```diff
@@ -52,8 +52,8 @@ def get_build_details(build_id: str) -> Optional[list[dict]]:
         return query.select()
     except ProgrammingError as e:
         if is_valid_does_not_exist_exception(e):
-            set_schema_version(version="5")
-            log_message("Build Details -- Schema version updated to 5")
+            set_schema_version()
+            log_message("Build Details -- Schema version updated")
```
murilx, line 56: nit: You could add the DEFAULT_SCHEMA_VERSION in this message. And the same goes to the other instances of `log_message`

```suggestion
            log_message(f"Build Details -- Schema version updated to {DEFAULT_SCHEMA_VERSION}")
```

backend/kernelCI_app/queries/tree.py
```diff
@@ -66,11 +108,13 @@ def get_tree_listing_data(
     # potentially causing the tree listing page show the same tree multiple times
     query = f"""
             SELECT
-                1 as id,
+                MAX(checkouts.id) as id,
```
murilx, line 111: Just out of curiosity, what is the reasoning behind this?

backend/kernelCI_app/tasks.py
```diff
@@ -1,20 +1,70 @@
...
+            checkout_ids_for_update.append(checkout.id)
+
+    updated_checkouts_data = get_tree_listing_data_by_checkout_id(
+        checkout_ids=checkout_ids_for_update,
+    )
+
+    for checkout in updated_checkouts_data:
+        checkout["unstable"] = _is_checkout_unstable(checkout=checkout)
+
+    populate_checkouts_cache_db(data=updated_checkouts_data, field_schema="raw")
```
murilx, line 70: `populate_checkouts_cache_db` does not have a `field_schema` parameter

backend/kernelCI_app/tasks.py
```diff
@@ -1,20 +1,70 @@
...
+    for checkout in all_kcidb_checkouts:
+        if not isinstance(checkout, Checkouts):
+            continue
+
+        is_unstable = _is_checkout_unstable(checkout=checkout)
+
+        if checkout.id not in all_sqlite_checkout_ids:
+            checkout_ids_for_update.append(checkout.id)
+            continue
+
```
murilx, line 59: nit: move this up to before `is_unstable = ...`

backend/kernelCI_app/tasks.py
```diff
@@ -1,20 +1,70 @@
...
+def _is_checkout_unstable(*, checkout: dict | Checkouts) -> bool:
+    three_days_ago = now() - timedelta(days=3)
+    if isinstance(checkout, Checkouts):
+        start_time = checkout.start_time
+    else:
+        start_time = checkout["start_time"]
+
+    is_old = start_time is not None and start_time < three_days_ago
+    is_done = _is_checkout_done(checkout=checkout)
+    if is_old or (not is_old and is_done):
```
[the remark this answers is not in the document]
MarceloRobert: I believe I was debugging some issue and thought that this could be it, but `is_old or is_done` should work

backend/kernelCI_app/tasks.py
MarceloRobert: Also from a previous way that I was doing stuff, since you called `populate_checkouts_cache_db` from treeView, the field names were different, but I was keeping both. Since I removed it from treeView I removed the field_schema parameter from the function but forgot to remove it here

murilx (approved):
Worked well on my tests. Nice job

PR #1152, opened 2025-04-02
Feat: use checkout cache

### LINKED ISSUE #1017: Cache checkouts when they finish testing
The latest checkout for a tree should be cached with relevant information, including the count for builds/boots/tests status
This cache could be done in a sqlite file in the backend
The primary key for that data will be the tree identifier from the new file in #1015

### LINKED ISSUE #1018: Use the cached checkouts in tree listing
Since the latest checkout is what we show in treeListing, we will be able to show the latest checkout for every tree that was cached by #1017 
The filter for "last x days" won't be necessary anymore
The query for tree listing fast and tree listing "slow" won't be necessary for the frontend anymore

Blocked by #1128

MarceloRobert:
Closes #1018 
Closes #1017


MarceloRobert:
Seems to be working correctly from my tests, all trees have the correct count and the rows are correct (mostly the same, except for the ones that shouldn't even be in staging)

PR #1153, opened 2025-04-02
feat: define origins dynamically

### LINKED ISSUE #1144: Add "arm" to origin selector
"arm" origin is sending data but the dashboard is not allowing the visualization of that data since the origins are hardcoded

### LINKED ISSUE #1149: Add origins dinamically
Origins should not be hardcoded, but we shouldn't lose the specificity of them, maybe adding a file in the backend where we can update it every so often, or an api call to it with a long cache ttl

murilx:
## Changes
- Adds origins endpoint to retrieve all origins from checkouts;
- Uses that endpoint in the frontend instead of having the origins hardcoded;
  - The origin search parameter is now a simple string, and not an enum anymore; however, there is no way to change it from the dashboard unless the user changes it in the url and the fields that are an enum in the backend (all status fields) are also just a simple string in the frontend search parameters

## How to test
Go to the tree listing and see that the origin selector now loads for a second, getting the origins from the backend
All endpoints that require origins should run normally

Closes #1149
Closes #1144

### DIFF
--- a/backend/kernelCI_app/views/originsView.py
+++ b/backend/kernelCI_app/views/originsView.py
@@ -1,0 +1,19 @@
+from http import HTTPStatus
+from kernelCI_app.helpers.errorHandling import create_api_error_response
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from kernelCI_app.queries.checkout import get_origins
+from drf_spectacular.utils import extend_schema
+
+
+class OriginsView(APIView):
+    @extend_schema(
+        responses={HTTPStatus.OK: list[str], HTTPStatus.BAD_REQUEST: dict[str, str]},
+        methods=["GET"],
+    )
+    def get(self, _request) -> Response:
+        origins = get_origins()
+        if len(origins) == 0:
+            return create_api_error_response(error_message="No origins found")
+
+        return Response(origins)
--- a/dashboard/src/api/origin.ts
+++ b/dashboard/src/api/origin.ts
@@ -1,0 +1,21 @@
+import { useQuery, type UseQueryResult } from '@tanstack/react-query';
+
+import { MILLISECONDS_IN_ONE_HOUR } from '@/utils/date';
+
+import { RequestData } from './commonRequest';
+
+const fetchOrigins = async (): Promise<string[]> => {
+  const data = await RequestData.get<string[]>('/api/origins/');
+  return data;
+};
+
+const ORIGIN_CACHE_DURATION = 2 * MILLISECONDS_IN_ONE_HOUR;
+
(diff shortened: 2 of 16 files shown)


dashboard/src/api/origin.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { RequestData } from './commonRequest';
+
+const fetchOrigins = async (): Promise<string[]> => {
+  const data = await RequestData.get<string[]>('/api/origins/');
+  return data;
+};
+
+// eslint-disable-next-line no-magic-numbers
+const TWO_HOURS_IN_MILISSECONDS = 2 * 60 * 60 * 1000;
```
MarceloRobert, line 11: This seems too specific to be made into a separate const. Maybe move this to a "milliseconds in an hour", and then "origin cache timeout" with 2 * that

backend/kernelCI_app/views/originsView.py
```diff
@@ -0,0 +1,22 @@
+from django.core.cache import cache
+from rest_framework.views import APIView
+from rest_framework.response import Response
+from kernelCI_app.queries.checkout import fetch_origins
+
+
+class OriginsView(APIView):
+    def __init__(self) -> None:
+        self.key = "origins"
+        self.duration = 12 * 60 * 60  # 12 hours
```
MarceloRobert, line 10: Just "duration" sounds like a generic variable type, in this context it's more specific to the cache duration

dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -43,22 +43,29 @@ const OriginSelect = ({ basePath }: { basePath: string }): JSX.Element => {
...
+    if (origin === undefined) {
       navigate({
-        search: previousSearch => ({ ...previousSearch, origin: origin }),
+        search: previousSearch => ({
+          ...previousSearch,
+          origin: DEFAULT_ORIGIN,
+        }),
       });
     }
   });
```
MarceloRobert, line 63: I don't think this is necessary since zod will already default the parameter to the default origin

backend/kernelCI_app/views/originsView.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def get(self, _request) -> Response:
+        origins = cache.get(self.key)
+        if origins is None or len(origins) == 0:
+            origins = fetch_origins()
+            cache.set(
+                self.key,
+                origins,
+                self.duration,
+            )
```
MarceloRobert, line 20: Considering the other queries that set caches, I would say that this cache get/set should be in the query file

MarceloRobert (approved):
Working well on my tests

PR #1163, opened 2025-04-08
Feat: use tree names file

### LINKED ISSUE #1148: Use treenames.yaml on tree listing
The data for treenames.yaml is meant to consolidate the names on treeListing, we should use it

### LINKED ISSUE #1162: use trees-name.yaml on hardware pages
A lot of hardware endpoints also return/show the tree_name, which should now be retrieved by the trees-name.yaml

MarceloRobert:
## Changes
Overrides the tree_name on all endpoints that return tree_name to use the tree_name on the trees-name.yaml file, which is retrieved by the git_repository_url

Special notes to the treeLatest endpoint (or the shortcut to a tree):
First, it will now retry the query in case the kcidb tree name is null but is not null on the file, allowing to search for a tree that doesn't have tree_name on kcidb (such as /tree/mainline/master?o=arm)
Secondly, it will also return the name from the file even if it exists on kcidb but is different. For example: if a tree has name A on kcidb but name B on trees-name.yaml, I can fetch /tree/A/branch and it will return me the treeDetails page but with tree name B; a breaking scenario could be if a tree _has_ the B name on kcidb when it shouldn't, in which case the shortcut /tree/A/branch will show the same _name_ as /tree/B/branch, but the pages will be different

There are also some trees, such as some chromiumos, which had their name on kcidb but could be changed to a new one following the trees-name.yaml file.

## How to test
Go to any tree page and check the tree name, especially for some trees which don't have tree_name, such as on origins other than maestro. Altering the trees-name.yaml file is also a way to better see the changes.

## Examples
[localhost 0dayci treeListing](http://localhost:5173/tree?o=0dayci) - [staging equivalent](https://staging.dashboard.kernelci.org:9000/tree?o=0dayci)
image
image

[localhost chromiumos](http://localhost:5173/tree?ts=chromiumos) - staging equivalent
image
image

Closes #1148 
Closes #1162

### DIFF
--- a/backend/kernelCI_app/helpers/hardwareDetails.py
+++ b/backend/kernelCI_app/helpers/hardwareDetails.py
@@ -331,8 +331,14 @@ def handle_test_history(
     *,
     record: Dict,
     task: List[HardwareTestHistoryItem],
+    tree_url_to_name: dict[str, str],
 ) -> None:
     create_record_test_platform(record=record)
+
+    defined_tree_name = tree_url_to_name.get(
+        record["build__checkout__git_repository_url"],
+        record["build__checkout__tree_name"],
+    )
 
     test_history_item = HardwareTestHistoryItem(
         id=record["id"],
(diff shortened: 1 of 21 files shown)


MarceloRobert:
From my testing, everything seems alright. @padovan would you like to check if those changes are correct?

MarceloRobert:
I'll be merging this PR since it got no more comments and there will be virtually only visual changes. Further comments can be made into issues for this area.

PR #1172, opened 2025-04-11
Add more trees notification

padovan:
Add omap and stable-rt


MarceloRobert:
It looks like there might be some connection issues from github's side, you can check if everything is alright locally by going to the backend folder and executing `poetry run pytest`. If everything is passing, try just re-running github's CI or bumping it with another push.

padovan:
It is also failing on the main branch for me. My change is clearly not related to it as I only touched some Yaml files in the notifications side.

PR #1173, opened 2025-04-12
trees-name: update some trees name

padovan:
For the chromeos/chromiumos duplication, it will eventually get resolved as the CI system are submitting under the same URL now.


MarceloRobert (approved):
LGTM. I see that GitHub is not being able to set up the credentials properly (hence the failing CI). I would say that for now it's ok and you can merge it, but if you had code which would affect the tests, it would be good to check how to use the secrets from this repository.

padovan:
We won't pursue maintaining tree names on the dashboard anymore.

PR #1179, opened 2025-04-16
Feat: order treeListing with pinned trees first

### LINKED ISSUE #998: pin mainline, next and stable branches at the top of the tree list
these are our most important trees. Let's pin a the top of the tree list view.

MarceloRobert:
## Changes
- Adds a list to specify priority trees;
- Uses that list to order the treelisting data, putting the prioritized trees first.

This means that when there is _no sorting_, those trees will appear at the top. In the case of any sorting, the sorting will override the ordering, but it is possible to reset the sorting by clicking at the header (sorting goes as "unsorted" -> "sort ascending" -> "sort descending" -> "unsorted")
The priority trees are defined by a regex that describes the tree/branch key of a tree (the git_repository_url was not used since the tree_name is now retrieved by the trees_name file which is unique for each url already). Being a regex it allows for a tree with any branch name.
The current priority trees are mainline/master, next/master, and stable trees with any branch.

## How to test
Go to the tree listing page and check the ordering of the data, try sorting with the other columns, filtering by path and checking other origins.
You can also change the list of pinned trees to add/remove them.

## Examples

New first page of tree listing
image

Second page
image

Closes #998

### DIFF
--- a/dashboard/src/components/TreeListingPage/TreeTable.tsx
+++ b/dashboard/src/components/TreeListingPage/TreeTable.tsx
@@ -63,6 +63,8 @@ import { statusCountToRequiredStatusCount } from '@/utils/status';
 import { MemoizedInputTime } from '@/components/InputTime';
 import { shouldShowRelativeDate } from '@/lib/date';
 import { valueOrEmpty } from '@/lib/string';
+import { PinnedTrees } from '@/utils/constants/tables';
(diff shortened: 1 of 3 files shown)


MarceloRobert:
@padovan can you confirm to me that these are the tree/branches that you want pinned?

padovan:
Correct

MarceloRobert:
All seems to be working fine on my tests

PR #1187, opened 2025-04-22
Feat: use trees name in shortcut endpoint

### LINKED ISSUE #1186: Make the tree latest endpoint be based on the trees-name file first
Currently, when using the shortcut /tree/tree_name/tree_branch, that tree_name is queried in the database as is, meaning that the tree_name in the search parameter is directly applied as a kcidb filter (in other words, we are searching for that tree_name as a kcidb tree_name).
However, since what is showed to the user is the tree_name from the trees-name.yaml file, the endpoint should first look which git_url is related to that tree_name in the file, and query for that url ignoring the kcidb tree_name.

Example:
If in the file I have:
```yaml
my_own_tree:
    https://my_tree_url.git
```
and in the kcidb there is no tree named "my_own_tree",
I should still be able to go to /tree/my_own_tree/my_branch and find the result where the url is `https://my_tree_url.git
`

This issue comes from #1185

MarceloRobert:
## Changes
- Alters the query for tree latest to optionally consider the git url and turn tree_name optional too
- Alters the endpoint to first try to fetch the result where the url matches the name in the file
- Adds the expanded frontend route /checkout as an alternative to /c

## How to test
Get a tree where the name has been altered in the trees-name.yaml such that the new name does not exist in kcidb (such as agross, or you can modify by hand) and go to the tree_latest endpoint with the altered name and with the kcidb name, both should point to the same checkout
In the frontend, routes /tree, /checkout and /c should have the same behavior as to shortcut to the latest tree checkout

Examples: http://localhost:5173/tree/agross/for-next and http://localhost:5173/tree/qcom/for-next are the same, even though there is no agross in kcidb and no qcom in the trees-name file

Closes #1186


dashboard/src/pages/TreeLatest.tsx
```diff
@@ -16,20 +16,15 @@ import { DEFAULT_DIFF_FILTER } from '@/types/general';
...
-}): JSX.Element => {
-  const searchParams = useSearch({ from: urlFrom });
+export const TreeLatest = (): JSX.Element => {
+  const searchParams = useSearch({
+    strict: false,
+  });
   const navigate = useNavigate();
   const { treeName, branch } = useParams({
-    from: urlFrom,
+    strict: false,
```
WilsonNet, line 25: please add type checking if you gonna remove the strict

MarceloRobert: I went back to the urlFrom parameter, not using `strict: false` anymore :+1:

PR #1192, opened 2025-04-24
Feat: change hardware listing to be platform based

### LINKED ISSUE #1184: Hardware list: refactor to base list on platform names
We had a feedback session with users and the felt the current UI is confusing. Let's reshape the List to be based on the platform name and then turn the compatible in tags that we can show in front of each platform name.

MarceloRobert:
## Changes
- Reformats and refactors hardware listing query to group by platform first and not unnest hardware;
- Changes typing accordingly to consider hardware as optional and platform as single string;
- Adds compatibles to frontend table as badges;
- Removes unnecessary filters and card from boots/tests tab for hardware/platform details.

Note: the change in the query for the hardware listing data was made such that the hardware are not unnested now, and from business requirements a platform should always be associated with the same array of compatibles, and in that way it is possible to group by platform and hardware without having more than one row repeating a platform.

## How to test
Go to the hardware listing page and see the listing based on platforms now.
- Check the search and sorting of the hardware table to see that they still work normally;
- Check if the counting between listing and details to see if they match;
- Compare with the staging hardware listing to see if the list of compatibles makes sense, you can also mannually go to a hardware details page on staging with the platform name and access the "PlatformDetails" from there (just remember to still have the startTimestamp and endTimestamp) in the search parameters

New hardware listing:
image

Old hardware listing:
image

Closes #1184

### DIFF
--- a/backend/kernelCI_app/queries/hardware.py
+++ b/backend/kernelCI_app/queries/hardware.py
@@ -21,39 +21,37 @@ from kernelCI_app.cache import get_query_cache, set_query_cache
 from kernelCI_app.typeModels.hardwareDetails import CommitHead, Tree
 
 
-def get_hardware_listing_data(
-    start_date: datetime, end_date: datetime, origin: str
-) -> list[dict]:
+def _get_hardware_listing_count_clauses() -> str:
     build_valid_count_clause = """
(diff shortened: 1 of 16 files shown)


MarceloRobert:
All good. No external objections so I'll merge with this design.

PR #1194, opened 2025-04-24
Refactor: make commit hash required in tree latest routes and endpoint

### LINKED ISSUE #1188: Turn gitCommitHash into a required searchParam
Change the routes so there are additional ones where the gitCommitHash param is required instead of a optional search param (remove the optional params and add the required routes), as in:
have /tree/mainline/master/12345 instead of /tree/mainline/master?gch=12345

MarceloRobert:
## Changes
- Adds new routes pointing to the same TreeLatest component;
- Adds endpoint pointing to the same view;
- Removes old search parameter.

Replaces the old method of using `?gitCommitHash=<hash>` (or `?gch`) with a real route `/<hash>` in the shortcuts for a specific tree checkout.
For example, the old:
`http://localhost:5173/tree/android/android13-5.15?gch=b00a7c5da3e82ced55aed32dce15a7188ac012c8`
Now becomes
`http://localhost:5173/tree/android/android13-5.15/b00a7c5da3e82ced55aed32dce15a7188ac012c8`

Also adds an endpoint for this purpose, before:
`http://localhost:8000/api/tree/android/android13-5.15?gch=b00a7c5da3e82ced55aed32dce15a7188ac012c8`
Now becomes
`http://localhost:8000/api/tree/android/android13-5.15/b00a7c5da3e82ced55aed32dce15a7188ac012c8`

## How to test
Test going to a checkout with a specific hash from a tree using the shortcut routes.
All shortcut endpoints should point to the same tree
Also try getting from the API

Closes #1188


padovan:
For clarity, if the user types:
* `/c/mainline/master/` it takes him to the lastest results
* `/c/mainline/master/<hash>` It takes them to the tree view of that specific checkout

PR #1196, opened 2025-04-25
notifications: proper bail out if issue has broken info

padovan:


backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -355,6 +355,10 @@ def generate_issue_report(service, issue_id, email_args, ask_ignore=False):
             incident["last_pass_commit"] = last_test[0]["git_commit_hash"]
             incident["last_pass_id"] = last_test[0]["id"]
         report = generate_boot_issue_report(issue, incidents)
+    else:
+        print(f"unable to generate issue report for {issue["id"]}", file=sys.stderr)
+        sys.exit(-1)
```
MarceloRobert, line 360: nit: I would be a bit more explicit for why it was unable to generate the issue report. In this case it seems like it is for not having any incidents (or simply the test and build ids are missing). But it also seems pretty easy to spot in the code where the problem is coming from, so no big deal

PR #1197, opened 2025-04-25
Move notifications helpers files out of the commands folder

padovan:
Take one more step to integrate our notifications with the rest of the Django code.


MarceloRobert (approved):
We can definitely join some queries and improve them in the future; for now it's a small step but there's no harm in it.
I ran the tests locally again and it's all fine, so you can ignore the CI this time again.

PR #1203, opened 2025-04-29
Feat: store when a checkout notification was sent

### LINKED ISSUE #1198: Store in Checkout model if a notification was sent or not
The Dashboard is sending notifications regarding the [status](https://groups.io/g/kernelci-results/message/58812) of the lastest checkout run of a given tree in a given origin.

Right now, it doesn't store any info if an email notification was sent or not. That limits what we can, so we want add that information to the checkout object we store in the Django db:

we want two fields:

* `notification_sent`(bool): true if it was sent
* `notification_message_id` (text): to store the message id of the email sent. (ex: "5e610cabe469732582afb752a0155ccf0e0d84fb.camel@collabora.com")

With this change, we can throw out the code that store this info in json files.

MarceloRobert:
## Changes
- Creates a new table in our sqlite to store the notifications sent for checkouts
  - Instead of using a simple boolean to check if a notification was sent, we can store the date it was sent, adding more information while keeping the possibility to check if a notification was sent or not (simply if it exists in the database or not)
  - The fields of that table are not unique, because a notification for the same checkout could be sent more than once
  - The checkout_id is not a foreign key to any other table, since the table for the cache for checkouts is unstable
- Adds simple logic to mark a checkout as sent, following what was done for issues (getting the `msg_id` from the `send_email_report`)

Note: the issue calls for an addition of new fields in the checkouts_cache table, but here there was an addition of a new table instead; this was done because the checkouts_cache is only meant to be used as an optional, faster source of data compared to kcidb. Since it is optional, it should be able to be removed without problem, but if we were to store the notification data there we would be losing information when dropping the checkouts_cache table. Therefore, we can use a separate table to store the notification data.

## How to test
These changes do not affect the dashboard, but the email notification system.

To test it, go to the backend folder and firstly update the sqlite schema with
`bash migrate-cache-db.sh`(or just `./migrate-cache-db.sh` if you're already using bash)

You must have the cloud-sql-proxy running in the background, then export the backend environment variables and in the same terminal run the following notification command
```
poetry run ./manage.py notifications --ignore-recipients --update-storage --action="summary"
```

A Gmail API key would be necessary, however, you can simply mock the msg_id with `randint` and assign None to the service at the beginning, avoiding sending emails entirely.

Closes #1198


joaobertacchi (approved):
LGTM. I didn't test it. Please make sure to test it before merging.

backend/kernelCI_cache/migrations/0008_notificationscheckout.py
```diff
@@ -0,0 +1,33 @@
...
+                (
+                    "id",
+                    models.BigAutoField(
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("notification_id", models.TextField()),
```
padovan, line 25: We should really name this `notification_message_id`.

PR #1204, opened 2025-04-30
wip

### LINKED ISSUE #1199: Store wether an issue email notification was sent or not
Similarly to #1198, we want to store the same info when we send email notifications for new issues.

MarceloRobert:
Closes #1199


MarceloRobert:
@padovan can you check the code? I made tests and everything seems to be working fine, but I would appreciate if you could take a look at it to see if the logic still follows what you expect

backend/kernelCI_cache/models.py
```diff
@@ -58,11 +58,38 @@ class Meta:
...
+        unique_together = (("issue_id", "issue_version"),)
+
+
+class IssuesCustom(models.Model):
+    id = models.TextField(primary_key=True)
+    version = models.IntegerField()
+    kcidb_timestamp = models.DateTimeField()
+    comment = models.TextField(null=True, blank=True)
+
+    type = models.TextField(null=True)  # PossibleIssueType (build/boot/test | null)
```
WilsonNet, line 89: according to this stack overflow answer django now has support for enums 
https://stackoverflow.com/questions/54802616/how-can-one-use-enums-as-a-choice-field-in-a-django-model

sqlite would need a check constraint according to this answer
https://stackoverflow.com/questions/5299267/how-to-create-enum-type-in-sqlite

Consider it a NIT because I don't know how feasible it is

backend/kernelCI_cache/queries/issues.py
```diff
@@ -0,0 +1,120 @@
...
+        FROM
+            issues_custom I
+        JOIN
+            notifications_issue N
+        ON
+            I.id = N.issue_id
+            AND I.version = N.issue_version
+        WHERE
+            I.notification_ignore = FALSE
+            AND N.notification_sent IS NULL
```
WilsonNet, line 30: shouldn't it be `IS FALSE` ? generally you don't want make the boolean field nullable

felipebergamin: `notification_sent` is a datetime. It will be null until the notification is sent.

wouldn't be better to name it as `notification_sent_at` or something like this to avoid confusion?

backend/kernelCI_cache/typeModels/issuesCustom.py
```diff
@@ -0,0 +1,13 @@
+from typing import Optional
+from pydantic import BaseModel
+
+from kernelCI_cache.typeModels.databases import PossibleIssueType
+
+
+class CustomIssue(BaseModel):
```
padovan, line 7: We don't really need CustomIssue stuff anymore. The usecase I had for these fields are no long valid, so let's throw it all away. I believe I mentioned some of that in the issue descrption.

backend/kernelCI_cache/queries/issues.py
MarceloRobert: If there was no notification for an issue, we won't store it in the database, meaning that `N.notification_sent` won't exist, therefore being null

backend/kernelCI_cache/queries/issues.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+
+    with connections["cache"].cursor() as cursor:
```
WilsonNet, line 33: I think you should probably create an issue to create another database for these long term operations, calling it `cache` will be misleading

backend/kernelCI_cache/queries/issues.py
```diff
@@ -0,0 +1,120 @@
...
+        SELECT
+            id,
+            version
+        FROM
+            issues_custom
+    """
+
+    with connections["cache"].cursor() as cursor:
+        cursor.execute(query)
+        results = cursor.fetchall()
```
WilsonNet, line 78: Shouldn't you be type checking with pydantic the result since it is a raw query?

backend/kernelCI_cache/queries/issues.py
```diff
@@ -0,0 +1,120 @@
...
+            I.id = N.issue_id
+            AND I.version = N.issue_version
+        WHERE
+            I.notification_ignore = FALSE
+            AND N.notification_sent IS NULL
+        """
+
+    with connections["cache"].cursor() as cursor:
+        cursor.execute(query)
+        return dict_fetchall(cursor)
```
WilsonNet, line 35: ditto about type checking

backend/kernelCI_cache/queries/issues.py
MarceloRobert: In this case I want the list of tuples so that I can easily compare if an issue already exists in the database via `(issue_id, issue_version)`

backend/kernelCI_cache/queries/issues.py
MarceloRobert: Makes sense, created at #1209

backend/kernelCI_cache/typeModels/issuesCustom.py
MarceloRobert: You're right, I did this in the second commit now. I had to make some changes to the notifications_issue table to allow it to store issues without notifications too.

backend/kernelCI_cache/queries/issues.py
WilsonNet: you wouldn't need to change the type, just use pydantic to check if it is right or throw an error if it is not, from what I understand this type is like a casting since it is a raw query, but maybe since you are using django migration it will work

Something like 

```
PydanticModel(results) #it does not change the type
return results
```

MarceloRobert: Done, it was fine to return it typed too

PR #1208, opened 2025-05-05
Feat: add status counts to checkout summary email notification

### LINKED ISSUE #1202: Add checkout summary to the status notification
We are not showing the "pass/fail/inconclusive" in the status notifications. We need to add that support, so the reader get a basic view of tree in the email as wel.

MarceloRobert:
In the current state, there are two queries for every tree in the checkout summary notification. This PR changes it such that there is a single query to get the basic checkout information of all trees at once, and then get the specific test data for each tree as before.
The query that replaces the current one for the checkout information is closely related to the tree listing query, which means that it will get the status count for all selected trees, and this is then added to the checkout summary email notification.

## Changes
- Replaces "latest checkout" query in the email notifications with a query similar to tree listing to get status count for all the trees in the yaml file;
- Changes the jinja template such that all fields from "checkout" are fields from kcidb, and also adds the counting for a tree;
- Refactors some tree queries and functions such that they are more reusable;

## How to test
These changes affect mostly the notifications. To test them, either set up the gmail api token or comment the lines that set up the service and send emails and mock the msg_id for the storage of checkout notifications. Then, run the command
```poetry run ./manage.py notifications --ignore-recipients --update-storage --action="summary"```

With the refactor of some tree-related functions, the tree listing and tree details should still be working normally

Closes #1202

### DIFF
--- a/backend/kernelCI_app/helpers/trees.py
+++ b/backend/kernelCI_app/helpers/trees.py
@@ -1,9 +1,12 @@
+import json
 import os
 import typing_extensions
 
 from django.conf import settings
 import yaml
 from kernelCI_app.helpers.logger import log_message
+from kernelCI_app.typeModels.common import StatusCount
+from kernelCI_app.typeModels.treeListing import Checkout
 
 
 def make_tree_identifier_key(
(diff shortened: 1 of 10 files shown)


MarceloRobert:
Eveything working on my tests

PR #1212, opened 2025-05-08
Refactor: notification files

### LINKED ISSUE #1206: Refactor how we store notifications subscription information
Today we store the recipients per tree at the `recipients.yaml` file. Then we \[have\] the `summary-signup.yaml` file for people. We have to combine it all in a single file per tree. So let's create the folder `backend/data/subscriptions` and add a file for each \[tree\] there following the yaml format below: 

```
tree:
  url: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git
  default_recipients:
    - laura.nao@collabora.com
    - shreeya.patel@collabora.com
    - broonie@kernel.org
 
  reports:
    - default:
      origin: maestro
      branch: master

    - rt_tests:
      origin: maestro
      branch: master
      path:
        - rt-tests*
      recipients:
        - linux-rt-devel@lists.linux.dev
      options:
        - ignore_default_recipients
```

`default_recipients`  has the email address that will receive issues reports and default status summary.

`reports:` section are to list the different summary reports users may want to have.

### LINKED ISSUE #1210: Just refactor current way of using tree files for notifications
This issue should enclose only the refactoring of the current functionalities, but with the new structure written in the parent issue

Part of #1206

### LINKED ISSUE #1211: Add other ways to retrieve specific test data for trees
After refactoring the current organization of files for tree storing, we then add a way to retrieve or separate specific test results for a certain tree described by the new structure in the parent issue

Depends on #1210 
Part of #1206

MarceloRobert:
Changes the way the signup files are stored, making one file for each tree and combining the recipients.
The files in the submissions folder are grouped by "tree_name" (in reality it is `git_repository_url`, but the name follows `trees-name.yaml`)

## Changes
- Converts old recipients + summary_signup to the submissions files following the example provided by issue 1206;
- Refactors some functions to uncouple them from the notifications file, to make it more readable;
- Refactors other functions in the notifications file to assure named parameters.

The refactor to the new file structure also allows for multiple different reports for the same tree, even same branch and origin, but changing path, recipients and possible options. For now, one supported option is `ignore_default_recipients`, which will make a report only use specific recipients, and not the default ones.

## How to test
This PR changes only the notification system of the dashboard. The main target for this refactor is the summary command, which you can run with
```poetry run ./manage.py notifications --ignore-recipients --update-storage --action="summary"```
Also try adding the argument `--summary-origins` with a comma-separated list of origins to restrain the checkout summary to those origins.

Since there was a refactor in the entire notifications file and many functions, it would be ideal if all of them were tested to check if they are working just as before.

Closes #1210 
Closes #1211
Closes #1206

### CODE REVIEW

File: backend/kernelCI_app/management/commands/data/old-recipients.yaml
Comment (padovan) [inline backend/kernelCI_app/management/commands/data/old-recipients.yaml:1]: Why do we still have this old-recipients file?

File: backend/kernelCI_app/management/commands/notifications.py [hunk 1]
```diff
@@ -43,13 +46,15 @@
 
 from kernelCI_app.utils import group_status
 
-STORAGE_FILE = "found_issues.json"
-SUMMARY_SIGNUP_FILE = "data/summary-signup.yaml"
+
+SUMMARY_SIGNUP_FOLDER = "data/subscriptions/"
```
padovan, line 50: Let's move them to the main backend data folder and delete this one

backend/kernelCI_app/management/commands/data/old-recipients.yaml
MarceloRobert: There were some emails in the recipients.yaml that didn't relate to any trees in the summary_signup.yaml file, so they were never really used but I was unsure if I could erase them, so I left them in that old-recipients file. Do you think it's ok to delete them too?

backend/kernelCI_app/management/commands/notifications.py
MarceloRobert: Yeah, they are not really related to the entire backend, only to the notifications command; but since they are the only data stored in that commands folder, makes sense to move it over to the main data folder

padovan:
These changes are not picking up the cc correctly anymore:

If I run:

`poetry run python3 manage.py notifications --action="issue_report" --id maestro:b7656501ace97abe52e3fb7260c982b6353f638f --send --add-mailing-lists`

in the main branch, I get:

```
Subject: [REGRESSION] chromiumos/chromeos-6.6: (build) common.h: No such file or directory in drivers/firmware/arm_scmi/t...   
To: kernelci-results@groups.io
Cc: laura.nao@collabora.com, shreeya.patel@collabora.com  
```

but on yours, the code fails to pull the cc information in:

```
Subject: [REGRESSION] chromiumos/chromeos-6.6: (build) common.h: No such file or directory in drivers/firmware/arm_scmi/t...
To: kernelci-results@groups.io  

```

MarceloRobert:
@padovan With the change I had made, it was done such that the git_url was the key to look for recipients, and in that issue that you tested the url ended with .git while in the `chromeos` file it ends without .git. I just pushed a change to make the tree_name priority over git_url, but since the url was different we could check if this is the right approach, do you think we can change the url in the files to match the ".git at the end"?

padovan:
The chromeos case was a corner case really. So let's keep using giturl as our main key

padovan:
I guess we change the chromeos file to have the right url in there

backend/kernelCI_app/management/commands/data/old-recipients.yaml
MarceloRobert: I have removed those recipients that weren't used; but I kept that change in a separate commit in case there's a need to go back to/check them

padovan (approved):

PR #1218, opened 2025-05-12
Fix: new issues ignoring stored issues

### LINKED ISSUE #1213: Regression: new issues are ignoring already stored issues
When running the new_issues action in the notifications command, issues that were already found are ignored and they are counted as new again.

This is caused because of the typing in the query that returns the already found issue keys, instead of returning a list of tuples, it is returning a list of objects IssueKeyTuple, and because of that the comparison to see if a new tuple was already found always fails (a tuple will never exist in a list of IssueKeyTuple, so the program thinks that any issue was never found/is new)

MarceloRobert:
When adding a type validation to the query, the return became `[IssueKeyTuple(str, int)]` instead of a simple list of tuples, so when comparing a tuple with the result, it was comparing a tuple with an IssueKeyTuple object, which was always false and therefore new issues were skipping the stored issues.

This PR also has a small change to fix the default gmail api file, such that we have a default fallback if we don't explicitly pass it to run the notification command.

## How to test
Check your local sqlite database, then erase the data to make it simpler to compare
Run the new issues command,
```
poetry run ./manage.py notifications --ignore-recipients --update-storage --send --action="new_issues"
```
Check the database and see that the issues are stored there as unsent
Run the new issues command again. There should be no new issues (previously there would be the same new issues again)

Closes #1213

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -173,7 +173,7 @@ CRONJOBS = [
     ),
 ]
 
-GMAIL_API_TOKEN = get_json_env_var("GMAIL_API_TOKEN", "")
+GMAIL_API_TOKEN = get_json_env_var("GMAIL_API_TOKEN", "gmail_api_token.json")
 
 # Database
 # https://docs.djangoproject.com/en/5.0/ref/settings/#databases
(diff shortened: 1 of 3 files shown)


MarceloRobert:
All working on my tests

PR #1225, opened 2025-05-13
Feat: update docker volume files from local

### LINKED ISSUE #1222: Update volume files when local files are added
Currently, the volume that we have for backend data is mounted over the local files, meaning that those local files are inaccessible when the project run in docker.
We could make it such that the local files are copied to the volume before starting the backend, this way we have them synced.

MarceloRobert:
## Changes
Adds the behavior that when a file is added locally to backend/data it is also synced to the docker volume. This is made because the docker volume is mounted on top of backend/data, making local changes inaccessible. The solution made was to copy local files to a temporary folder, then mount the volume, then sync back files from the temporary folder to the volume (adds files that didn't exist, ignores existing ones), the temporary folder is then removed.

Also moves the `gunicorn` comand to the dockerfile, such that when running backend notification commands from docker the arguments are passed only to the notification command, and not gunicorn.

Reiterating that existing files are skipped, meaning that _changes_ to local files are still not passed to the volume. This was kept this way because the cache that exists on staging and production will always be larger to the local due to the length of time that the cron job has been running there.

## How to test
Run docker compose to up the project, then add a local file or folder to backend/data.
Before this PR, the file would not appear in the volume, now it does.

Closes #1222

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -16,6 +16,7 @@ RUN apk update \
         libc-dev \
         libpq-dev \
         postgresql \
+        rsync \
     && python3 -m venv $POETRY_HOME \
     && $POETRY_HOME/bin/pip install poetry==1.8.3 \
     && ln -s $POETRY_HOME/bin/poetry /bin/poetry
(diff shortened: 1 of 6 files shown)


backend/Dockerfile
```diff
@@ -25,6 +26,14 @@ COPY . /backend
 COPY ./utils/docker/entrypoint.sh /entrypoint.sh
 WORKDIR /backend
 
+# Since the volume mount will overlay the local folder,
+# if there is any local change that we want to update we have to copy it to
+# a temporary folder that we then will sync back again to the docker volume
+# (that is, copy the local files that don't exist in the volume
+# and keep intact the ones that are already present in the volume)
+RUN mkdir -p /tmp/backend_data_backup && \
+    cp -r /backend/data/. /tmp/backend_data_backup/
```
WilsonNet, line 35: I'm pretty sure when you specify files it doesn't copy hidden files, give it a try to 
https://superuser.com/questions/61611/how-to-copy-with-cp-to-include-hidden-files-and-hidden-directories-and-their-con

make some tests

MarceloRobert:
As we re-planned, manually added files should not be in the volume, so I'm closing this PR for its main purpose and moving some of the changes to another PR

PR #1227, opened 2025-05-15
Refactor: docker volume changes

### LINKED ISSUE #1226: Set docker volume only for automatically generated data
Currently, the docker volume "db-data" (which should be renamed) is mounted over /backend/data, which not only confuses with the local folder /backend/data, but also may feel like it should include static files.

The volume that we use should store all and only data that is generated dynamically and that we don't want to lose information from.
This includes the sqlite database that we currently have, and the trees-name.yaml (generated with a command and cron jobs).
Those files should also not be included in the git tracking.

- [ ] Change the path where the volume is mounted
    - [ ] Change where the command/database points to with it
- [ ] Ignore local files and changes to files that should be stored in the volume

MarceloRobert:
## Changes

- Moves the volume folder from the container /backend/data to the container /volume_data. This is done to not confuse with the local /backend/data and also to keep volume files very distinct from local, statically generated files
- Updates some docs
- Fixes a problem with running notification commands with `docker compose run`, where the command argument was being passed to gunicorn

## How to test
Run the project with `docker compose up --build`, you should have the environment variables set in a .env, the volume folder is not necessary to be set since it has a default value.
Check if the volume was created and if the files are in it.
Run commands from `docker compose run dashboard_backend_service poetry run ./manage.py <command>` to check that they work normally

Closes #1226


WilsonNet (approved):
haven't test but looks good

PR #1232, opened 2025-05-20
Refactor: allow more filters for issue listing

### LINKED ISSUE #1074: Add filter to Issues page
It would be great to be able to filter issues per type, architecture, device stuff. Something we have to think about and plan.

### LINKED ISSUE #1235: Refactor issue page to accept any filter
Currently, the Issue Listing page accepts only arguments regarding the culprit, but we should refactor it to accept any filter, following the structure made for other pages. This will allow future improvements to add other filter options.

MarceloRobert:
Refactors issue listing to remove the old culprit system and allow for a more expansive approach with `diffFilter`, allowing future changes to add more filters.
The issue culprits are separate fields in the database, but since they are about the same idea, they can be grouped inside of diffFilter, which sends groups of filters to the backend.

## Changes
- Swaps django query constructor for SQL raw query in issueListing endpoint
- Changes filtering from query layer to python layer
- Changes frontend query parameter structure from individual culprit fields to `diffFilter`
- (piggybacking) fixing a one-line problem that the cache database was always being created as an unstaged file in the backend root directory when running locally

## How to test
The default hardcoded filter by culprit code is still being applied, so going to the issue listing page should show the same result as the production issue listing page.
It is possible to filter by other culprits by adding `?df|icu|tool=true` or `?df|icu|harness=true` in the url or through the api by adding `filter_issue.culprit=tool` or `filter_issue.culprit=harness`. It is also possible to combine those filters.

Closes #1235 
Part of #1074

### DIFF
--- a/backend/kernelCI_app/views/issueView.py
+++ b/backend/kernelCI_app/views/issueView.py
@@ -1,5 +1,6 @@
-from datetime import datetime, timedelta, timezone
 from http import HTTPStatus
+from typing import Optional
+from django.http import HttpRequest
 from drf_spectacular.utils import extend_schema
 from rest_framework.views import APIView
 from rest_framework.response import Response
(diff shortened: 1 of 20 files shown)


backend/kernelCI_app/views/issueView.py
```diff
@@ -8,21 +9,62 @@
...
+
+    def _should_discard_issue_by_culprit(
+        self, *, culprit_filters: set[PossibleIssueCulprits], record: dict
+    ) -> bool:
+        """Returns true if the record should be discarted from the resulting set of records"""
+
+        if len(culprit_filters) == 0:
+            return False
+
+        if record["culprit_code"] is True and CULPRIT_CODE in culprit_filters:
```
joaovitorgit, line 44: It's up to you, but you could create an enum for issueCulprits. This would change the need to check each case specifically, and you could simply iterate over the cases.

PR #1233, opened 2025-05-20
notifications: add info for all Maestro trees

padovan:
We need to set the recipients for all trees, so notifications can reach to the right people.

### DIFF
--- a/backend/data/notifications/subscriptions/amlogic.yaml
+++ b/backend/data/notifications/subscriptions/amlogic.yaml
@@ -1,0 +1,5 @@
+amlogic:
+  url: https://git.kernel.org/pub/scm/linux/kernel/git/amlogic/linux.git
+  default_recipients:
+    - linux-amlogic@lists.infradead.org
+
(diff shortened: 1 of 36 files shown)


MarceloRobert:
The notification command doesn't work when the files don't have any reports. It would be useful if those files were skipped btw, just adding a
```python
if tree_values.get("reports") is None:
    continue
```
before iterating in those reports in `process_submission_files()`; currently the command just breaks.

If your intent is to add new reports to those trees, specify the report with the branch that you want the report to be for.

padovan:
> The notification command doesn't work when the files don't have any reports. It would be useful if those files were skipped btw, just adding a
> 
> ```python
> if tree_values.get("reports") is None:
>     continue
> ```
> 
> before iterating in those reports in `process_submission_files()`; currently the command just breaks.

Thanks for this. I do want to have files without any report information, so I just added this if clause there.

MarceloRobert:
This PR will be closed in favor of PR #1234

PR #662, opened 2024-12-10
Feat: add tree filter in modal hardware

Francisco2002:
- create tree section filter
- use treeIndexes query params to filter hardware

Close #447

PR #1240, opened 2025-05-23
Feat: add filters in issue listing page

### LINKED ISSUE #1084: bring back culprit to issue listing table

### LINKED ISSUE #1236: Add filters and filter UI to issue listing
After issue listing supports more filters, we should add a UI to be able to select other fields, such that it doesn't stay only locked to culprit code.

- [x] Refactor current component
- [x] Populate modal with response data
- [ ] Add handlers and filters in the backend for new fields
- [x] Edit current checkbox filter section to show a manual option

MarceloRobert:
WIP

Closes #1236 
Closes #1084


dashboard/src/pages/IssueListing/IssueListingFilter.tsx
```diff
@@ -0,0 +1,155 @@
...
+      }
+      if (issue.culprit_tool) {
+        filters.issueCulprits[CULPRIT_TOOL] = false;
+      }
+
+      if (issue.categories) {
+        filters.issueCategories = {};
+        for (let j = 0; j < issue.categories.length; j += 1) {
+          const category = issue.categories[j];
+          filters.issueCategories[category] = false;
```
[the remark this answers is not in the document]
MarceloRobert: Makes sense, I though about initializing the categories filter only if some issue had some categories in order to not show that section without any category, but I can initialize the variable with the others and it still won't show up in the filter modal since it is an empty dict. Done.

dashboard/src/utils/search.ts
```diff
@@ -188,6 +189,8 @@ const diffFilterMinifiedParams: Record<TFilterKeys, string> = {
   bootIssue: 'bti',
   testIssue: 'ti',
   issueCulprits: 'icu',
+  issueCategories: 'ica',
+  issueOptions: 'io',
```
WilsonNet, line 193: nit: `as const satisfies Record<TFilterKeys, string>`

backend/kernelCI_app/helpers/issueListing.py
```diff
@@ -0,0 +1,96 @@
...
+    CULPRIT_CODE,
+    CULPRIT_HARNESS,
+    CULPRIT_TOOL,
+    HAS_INCIDENT_OPTION,
+    IssueFilterOptions,
+    PossibleIssueCulprits,
+)
+
+
+def should_discard_issue_by_culprit(
```
WilsonNet, line 14: nice that you are creating helper functions, but I would unit test these, so my brain doesn't have to keep track of the boolean logics

dashboard/src/components/Filter/CheckboxSection.tsx
```diff
@@ -68,10 +71,26 @@ const CheckboxSectionItem = ({
     () => onClickItem(value),
     [value, onClickItem],
   );
+
+  const { formatMessage } = useIntl();
+
+  // Checks if the value to be shown is different than the key used in diffFilter
+  // This happens for filters that are manual options instead of values from received data
+  const shownValue = useMemo(() => {
+    for (const page of Object.values(OptionFilters)) {
```
WilsonNet, line 80: this is not wrong, but people generally goes with `Obejct.values(OptionFileters).forEach` since you are not using break or continue

dashboard/src/components/Filter/CheckboxSection.tsx
```diff
@@ -68,10 +71,26 @@ const CheckboxSectionItem = ({
     () => onClickItem(value),
     [value, onClickItem],
   );
+
+  const { formatMessage } = useIntl();
+
+  // Checks if the value to be shown is different than the key used in diffFilter
+  // This happens for filters that are manual options instead of values from received data
+  const shownValue = useMemo(() => {
```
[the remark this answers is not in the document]
MarceloRobert: I also got a bit confused here but that assertion is not really an assertion, _sometimes_ `value` will be of the type `SplitFilterString` (for the tree filters in hardwareDetails), so that `as SplitFilterString` is acting as a way to ensure that the string will be able to be passed to that function. This was made in [PR 662](https://github.com/kernelci/dashboard/pull/662).

backend/kernelCI_app/helpers/issueListing.py
MarceloRobert: Makes sense, I thought that saying "should discard" would make more sense since it has a clearer meaning than the other filter functions, but I created some unit tests for it now

WilsonNet: Nice, keep in mind if the logic bugs one day, don't forget to cover that bug case with a unit test also

WilsonNet (approved):
looks good

PR #1254, opened 2025-05-29
Fix: i route conflict

### LINKED ISSUE #1253: Check issue i alternative route
It seems that the alternative route /i is always redirecting to issue listing instead of issue details.

When using /i/<issue_id> it should redirect to the issue details page, such as
https://d.kernelci.org/i/maestro:b27d3eb21d20e4c98fbf2773a2abf6ea009f73bb ->
https://d.kernelci.org/issue/maestro:b27d3eb21d20e4c98fbf2773a2abf6ea009f73bb

MarceloRobert:
Fixes a bug where going to /i/someid would redirect to /issues instead of /issue/someid

References:
https://tanstack.com/router/latest/docs/framework/react/routing/routing-concepts
https://tanstack.com/router/latest/docs/framework/react/routing/route-matching

## Changes
- Removed route.tsx for /i (this matched routes with children, while index.tsx only matches exact route with no children)
- Moved search validation from route.tsx to index.tsx for /i -- since the index for this route also defines a file route it can have the search validation in it, it is not exclusive to the route file

## How to test
Try going to /i and /i/$issueId and check if the redirection is working accordingly;
Also check if the search parameters are correct

Closes #1253


WilsonNet (approved):
makes sense

PR #1255, opened 2025-05-29
messages: improve explanation when we have no results

padovan:
We need to give clear message to the user if there is no results for a given tree/branch/commit.

### DIFF
--- a/dashboard/src/locales/messages/index.ts
+++ b/dashboard/src/locales/messages/index.ts
@@ -311,7 +311,7 @@ export const messages = {
     'treeDetails.inconclusiveBuilds': 'Inconclusive Builds',
     'treeDetails.inconclusiveTests': 'Inconclusive Tests',
     'treeDetails.invalidBuilds': 'Failed builds',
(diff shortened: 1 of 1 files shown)


MarceloRobert:
@padovan I checked here a bit more and it seems that if a checkout exists but has no tests, it would fall under "No tests found for this tree checkout" ([this message](https://github.com/kernelci/dashboard/blob/96d8160fa8ac356c81fe26938a6dc3c6e946bb6e/backend/kernelCI_app/views/treeDetailsTestsView.py#L91)). If the response that he got was "Tree checkout not found" then that means that the checkout itself did not exist in kcidb, as the message suggests.

I tested with the [android/android15-6.6](https://dashboard.kernelci.org/api/tree/8b5d78fb5c56fd82d8e621644ff073a375fa83d6/tests?git_branch=android15-6.6&git_url=https:%2F%2Fandroid.googlesource.com%2Fkernel%2Fcommon&origin=maestro) which _at the time of writing_ has 0 builds/boots/tests. You may in the dashboard any other tree with 0 tests and go to the endpoint to see the message it returns if you like.

PR #1271, opened 2025-06-04
Refactor: move tree details origin

### LINKED ISSUE #1243: Drop Origin Selector for Tree Details and add it to top details
Let's drop the top select for origins and add the `Checkout Origin` info to the  Tree, Branch, etc. one-line table.

Image

MarceloRobert:
Also slightly changes the top bar positioning in the code

Closes #1243


dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -98,18 +99,29 @@ const TitleName = ({ basePath }: { basePath: string }): JSX.Element => {
...
-  const { pathname } = useLocation();
-  const redirectFrom = useRouterState({ select: s => s.location.state.from });
-  const splitPath = pathname.split('/')[1] ?? '';
-  const basePath = redirectFrom ?? splitPath;
+  const matches = useMatches();
+  const redirectStateFrom = useRouterState({
+    select: s => s.location.state.from,
+  });
+
+  const pathname = matches[matches.length - 1].pathname;
```
LucasAzLima, line 107: Really nit: define a lastmatch const since it is reused in another place

LucasAzLima (approved):
Just a nit but lgtm

dashboard/src/components/TopBar/TopBar.tsx
```diff
@@ -98,18 +99,29 @@ const TitleName = ({ basePath }: { basePath: string }): JSX.Element => {
...
+  const firstUrlLocation = pathname.split('/')[1] ?? '';
+
+  const isTreeListing = useMemo(() => {
+    const lastMatch = matches[matches.length - 1];
+    const fullPath = lastMatch.fullPath;
+    return fullPath === '/tree' || fullPath === '/tree/';
```
giubrocchi, line 113: nit: could we do something like `fullPath.replaceAll('/', '') === 'tree'`?

mari1912 (approved):

dashboard/src/components/TopBar/TopBar.tsx
MarceloRobert: Ah, `fullPath` here is like a Literal containing only the routes names, so I didn't think that a general string method would work. But you're right, I changed it.

LucasAzLima (approved):

PR #1299, opened 2025-06-18
Feat: short tree details url

### LINKED ISSUE #1281: Shorten treeDetails url further
A lot of trees have a long git_repository_url, and that data is being kept in the dashboard url, making it very long. We can explore other ways to show the tree url.

- [x] Discuss possible options for making the tree details url
  - Using the shortcut as the real url (/tree/name/branch/hash/) -- this would require all trees to have at least tree_name and branch, which is not happening right now, but we could bring the trees_name.yaml file back in order to set the tree_name
  - Using the checkout id

Since we already have the short url as the shortcut, we can switch their use and when the short url is not possible we can keep the long one

- [x] Change current tree details endpoint to accept the other format (instead of information in the request, the information will come directly from the endpoint). Also make it possible to access that endpoint with only tree name, branch and hash data.
- [x] Shorten treeDetailsSummary
- [ ] Shorten treeDetails specific endpoints (build/boots/tests)
- [x] Change the frontend to make the url use either the long or the short url in the treeDetails page

MarceloRobert:
## Description
This PR changes the way that the data is used in order to get the treeDetails page. Instead of relying on the `git_url` of a tree, we now only need the `tree_name`, `git_repository_branch`, and `git_commit_hash`. When all of them are available, the url for the treeDetails page should be `/tree/<tree_name>/<branch>/<hash>`.

This has implications of changing how the backend uses those variables to query the database, and implications on the frontend subcomponents that get information from the url and navigate between pages. The frontend page can now be accessed through both urls (no redirecting, just using the same component), and the tree listing page will send the user to the new url for treeDetails. The backend views can also be accessed through both endpoints.

One issue with accessing the endpoint through two endpoints: there are some trees that don't have the url information (even though they should have), and in those cases it was useful to query by `git_repository_url IS NULL` but since the endpoints can be accessed without passing the git_url param (using only tree_name, branch, and hash to get the data), it's unclear if an empty git_url means that a tree has `NULL` git_url or if the endpoint was just accessed without that information. This means that the where clause for `git_url IS NULL` is only used if `tree_name` is also None, meaning that the user did access the endpoint from a place where they should pass `git_url`, but didn't.

The url `/tree/<tree_name>/<branch>` still leads to the shortcut of latest commit hash, and transfers the user to the old, longer, url for now.
Other endpoints such as treeDetailsBuilds/Boots/Tests were not changed.

## Changes
- [x] The backend treeDetails full and summary endpoints accepts both kinds of access (using the information in the `treeInfo` request parameters or using the `tree_name`, `branch` and `hash` from the url itself) to return the data.
- [x] Alters the frontend route of /tree/name/branch/hash to copy treeDetails search schema
- [x] Refactors treeDetails subcomponents to accept both urls in the `useSearch` and `navigate` `from` parameters.
- [x] Alters the treeListing table to direct to the direct page when possible;
- [x] Changes treeCommitHistory to be able to return data without passing a git_url for the query
  - [x] Uses `tree_name` instead of `git_url`
  - [x] Alter frontend api to call the endpoint based on which treeDetails url is being used

## How to test
Check the treeDetails page through the frontend, including cases of empty tree name or branch or url.
Check the treeDetails components, navigation and data, comparing with the production instance, all should be the same.
Check the commitHistory and treeDetailsSummary endpoints

Closes #1281

### DIFF
--- a/dashboard/src/api/commitHistory.ts
+++ b/dashboard/src/api/commitHistory.ts
@@ -1,7 +1,9 @@
 import type { UseQueryResult } from '@tanstack/react-query';
 import { useQuery } from '@tanstack/react-query';
 
+import { treeDetailsDirectRouteName } from '@/types/tree/TreeDetails';
 import type {
+  TreeDetailsRouteFrom,
   TTreeCommitHistoryResponse,
   TTreeDetailsFilter,
 } from '@/types/tree/TreeDetails';
@@ -21,6 +23,8 @@ const fetchCommitHistory = async (
   filters: TTreeDetailsFilter,
   startTimestampInSeconds: number | undefined,
   endTimestampInSeconds: number | undefined,
+  treeName?: string,
+  treeUrlFrom?: TreeDetailsRouteFrom,
 ): Promise<TTreeCommitHistoryResponse> => {
   const filtersFormatted = mapFiltersKeysToBackendCompatible(filters);
 
@@ -33,8 +37,13 @@ const fetchCommitHistory = async (
     ...filtersFormatted,
   };
 
+  const baseUrl =
+    treeUrlFrom === treeDetailsDirectRouteName
+      ? `/api/tree/${treeName}/${gitBranch}/${commitHash}`
+      : `/api/tree/${commitHash}`;
+
   const data = await RequestData.get<TTreeCommitHistoryResponse>(
-    `/api/tree/${commitHash}/commits`,
+    `${baseUrl}/commits`,
     {
       params,
     },
@@ -51,6 +60,8 @@ export const useCommitHistory = ({
   filter,
   endTimestampInSeconds,
   startTimestampInSeconds,
+  treeName,
+  treeUrlFrom,
 }: {
   commitHash: string;
   origin: string;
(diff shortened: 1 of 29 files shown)


dashboard/src/api/commitHistory.ts
```diff
@@ -33,8 +36,13 @@ const fetchCommitHistory = async (
     ...filtersFormatted,
   };
 
+  const baseUrl =
+    treeUrlFrom === '/_main/tree/$treeName/$branch/$hash'
```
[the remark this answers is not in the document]
MarceloRobert: done. I tried to export a variable starting directly from the route definition, but that would cause a circular dependency, which would require changing the file where the new map is being defined, and then changing all the imports from it. So I chose to just define them "independently", by the map side but not directly related to the real route name. This variable won't change much though, so I think this will be enough

dashboard/src/pages/TreeDetails/Tabs/Boots/BootsTab.tsx
```diff
@@ -32,25 +34,33 @@ import QuerySwitcher from '@/components/QuerySwitcher/QuerySwitcher';
...
+const BootsTab = ({
+  treeDetailsLazyLoaded,
+  urlFrom,
+}: BootsTabProps): JSX.Element => {
+  const params = useParams({
+    from: urlFrom,
   });
+  const treeId =
+    getStringParam(params, 'treeId') || getStringParam(params, 'hash');
+  const treeName = getStringParam(params, 'treeName');
```
giubrocchi, line 53: Since this is used on multiple components, do you think it's worth creating a hook to extract these values to make it easier to maintain?

MarceloRobert: I made a new helper function for it, but only in the cases where multiple fields are "sanitized", for the cases with only one or two variables I left it as is because they are not sanitizing the whole treeInfo object

giubrocchi (approved):

WilsonNet (approved):
looks good, but @padovan needs to be aware that for this to work they need to be strict with the tree names belonging to a single repo

PR #1305, opened 2025-06-24
fix:issue details log excerpt has no height limit

### LINKED ISSUE #1147: IssueDetails log excerpt has no height limit
The issue details page shows its log excerpt but doesn't have a sub-scroll, showing the entire log excerpt which can be very big,
example: https://staging.dashboard.kernelci.org:9000/issue/redhat%3A5489bb749eafbe4f9233f4a1e660c6d234b16408?iv=1

danieldiasbarbo:
## Description
This PR updates the component's maximum height constraint from max-h-[80%] to max-h-[80vh] to ensure correct rendering and scrolling behavior.
Using a percentage-based max height (80%) caused issues when the component was rendered inside containers with dynamic height, preventing overflow-auto from working correctly. By switching to a viewport-based height (80vh), the component now correctly limits its height relative to the viewport, allowing scrollbars to appear as expected when content overflows.

## Changes
- [x] Replaced max-h-[80%] with max-h-[80vh] in the component's class

## How to test
Access a issue detail page as in the following URL: /issue/redhat%3A5489bb749eafbe4f9233f4a1e660c6d234b16408?iv=1
Verify that the element previously overflowing now respects a height limit.
Confirm that it is scrollable when the content exceeds the visible area.
Ensure the scrollbar appears and functions correctly across different screen sizes.

## Visual Reference

image

Closes #1147

### DIFF
--- a/dashboard/src/components/Filter/CodeBlock.tsx
+++ b/dashboard/src/components/Filter/CodeBlock.tsx
@@ -232,6 +232,7 @@ const CodeBlock = ({
   code,
   highlightsClassnames,
   variant = 'default',
+  className,
 }: TCodeBlockProps): JSX.Element => {
   const disableHighlight = code.length >= MAX_HIGHLIGHT_CODE_LENGTH;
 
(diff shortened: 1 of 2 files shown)


giubrocchi (approved):
LGTM

PR #1307, opened 2025-06-24
Feat: Add kci-dev information

### LINKED ISSUE #1247: Add kci-dev information in the web dashboard
People might be interested in learning that can access the data through kci-dev as well, So we should have some tooltips or boxes with copy and paste commands.

MarceloRobert:
## Changes
- Adds a component that can be used as a footer to show a CTA for kci-dev, as well as a tooltip with the equivalent tooltip for that specific page
- Uses that component in all pages, showing the command for the pages that have an equivalent command for them
  - This includes the specific "builds/boots/tests" for treeDetails and hardwareDetails
  - It does not show any command for treeDetails without git_url
  - It does not show any command for issueListing and issueDetails pages since there are no commands for it there
  - All commands are subcommands of `kci-dev results`
- In order to use that component, the `sanitizeTreeInfo` was also used in the treeDetails tabs to get the required data and unify the processing.

## How to test
Go to any page and check the footer at the bottom of the page
Check the different commands, including the difference for each tab under treeDetails and hardwareDetails

## Visual reference
Isolated component while hovering in the icon:
image

Component with dynamic command:
In treeDetails:
image
In buildDetails:
image

In the treeListing page:
image

Closes #1247

### DIFF
--- a/backend/kernelCI_app/tests/integrationTests/treeListing_test.py
+++ b/backend/kernelCI_app/tests/integrationTests/treeListing_test.py
@@ -58,7 +58,7 @@ def pytest_generate_tests(metafunc):
                 False,
             ),
             (
-                {"origin": "tuxsuite", "interval_in_days": "7"},
+                {"origin": "linaro", "interval_in_days": "7"},
                 HTTPStatus.OK,
                 False,
             ),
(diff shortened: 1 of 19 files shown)


aliceinwire:
looks like we are in the right path

aliceinwire (approved):

PR #1315, opened 2025-06-26
Refactor: Move all backend strings to a single file

khanheric2003:
Replace harcoded string in endpoint return in view kernelCI_app/views to kernelCI_app/constants/localization.py

### DIFF
--- a/backend/kernelCI_app/views/kciSummaryView.py
+++ b/backend/kernelCI_app/views/kciSummaryView.py
@@ -33,7 +33,7 @@ class KciSummary(APIView):
                 origin=request.GET.get("origin", DEFAULT_ORIGIN),
                 git_branch=request.GET.get("git_branch"),
                 git_url=request.GET.get("git_url"),
-                path=request.GET.get("path", DEFAULT_PATH_SEARCH),
+                path=request.GET.getlist("path", DEFAULT_PATH_SEARCH),
                 group_size=request.GET.get("group_size", DEFAULT_GROUP_SIZE),
             )
             origin = params.origin


MarceloRobert:
Hi Khanh! Thanks for the contribution, I will review the changes when possible.
For now, please check the commit organization, I don't think there is a need for the first one, and change the message of the second one to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (you can see examples in the previous commits of the repository), in the case for this task it should start with `refactor`.

khanheric2003:
sure thing let me change it in a bit.

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,76 @@ class ClientStrings:
...
+
+    # buildIssuesView.py
+    BUILD_ISSUES_NOT_FOUND = "No issues were found for this build"
+
+    # buildDetailsView.py
+    BUILD_DETAILS_NOT_FOUND = "Build not found"
+
+    # treeLatest.py 
+    # treeView.py
+    # treeViewFast.py
```
danieldiasbarbo, line 86: Thank you for the contribution, it's great to see all the strings centralized in a single file!
Just a small note: there's no need to include comments indicating where each string is used. This adds an extra place to update whenever the usage changes, which can become a maintenance burden and is potentially unreliable if the usage is deprecated or moved. The usage locations already reference the string, there is no need to maintain the reverse mapping manually.

khanheric2003: sure let me remove it

khanheric2003: I have resolved it

MarceloRobert:
Please check linting, we use Black formatter and flake8 for the backend, you can run the same command as the githubCI to check for problems, or install an extension to help you with that.

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
...
+    HARDWARE_INVALID_JSON = "Invalid body, request body must be a valid json string"
+    HARDWARE_INVALID_TIMESTAMP = "startTimeStamp and endTimeStamp must be a Unix Timestamp"
+    HARDWARE_NO_COMMITS = "This hardware isn't associated with any commit"
+    HARDWARE_TEST_NOT_FOUND = "No tests found for this hardware"
+    HARDWARE_COMMMIT_HISTORY_INVALID_TIMESTAMP = "startTimestampInSeconds and endTimestampInSeconds must be a Unix Timestamp"
+    HARDWARE_COMMIT_HISTORY_NOT_FOUND = "Commit history not found"
+    HARDWARE_BUILDS_NOT_FOUND = "No builds found for this hardware"
+    HARDWARE_BOOTS_NOT_FOUND = "No boots found for this hardware"
+    BUILDS_TESTS_NOT_FOUND = "No tests found for this build"
+    BUILD_ISSUES_NOT_FOUND = "No issues were found for this build"
```
MarceloRobert, line 44: You could also make the strings more similar here. Also, keep the keys consistent, in this case there are tests for a single build (not build**s**)
```suggestion
    BUILD_TESTS_NOT_FOUND = "No tests found for this build"
    BUILD_ISSUES_NOT_FOUND = "No issues found for this build"
```

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
...
+    LOG_TABLE_BODY_NOT_FOUND = "Table body not found"
+    LOG_INVALID_TABLE_FORMAT = "Invalid number of columns in table row (probably not a log website)"
+    LOG_NO_FILES_FOUND = "No log files found"    
+    ISSUE_NOT_FOUND = "No issues found"
+    ISSUE_INVALID_JSON = "Invalid body, request body must be a valid json string"
+    ISSUE_EMPTY_LIST = "Invalid body, the issue list must not be empty"
+    ISSUE_NO_EXTRA_DETAILS = "No extra details found"
+    ISSUE_TEST_NOT_FOUND = "No tests found for this issue"
+    ISSUE_DETAILS_BUILD_NOT_FOUND = "No builds found for this issue"
+    HARDWARE_NOT_FOUND = "No hardwares found"
```
MarceloRobert, line 34: Since you are here, please fix the typo
```suggestion
    HARDWARE_NOT_FOUND = "No hardware found"
```

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -299,7 +299,7 @@ def post(self, request, hardware_id) -> Response:
 
         if len(records) == 0:
             return create_api_error_response(
-                error_message="Hardware not found",
+                error_message=ClientStrings.HARDWARE_NOT_FOUND,
```
MarceloRobert, line 302: This is not quite the right string. There is a slight difference between "Hardware not found" and "No hardware found". One hints at the idea that a hardware should be there but wasn't found, while the other just states that none was found, as if it was expected to have some number of hardware as a return.

This means that the `hardwareView` (which returns a list of hardware) should use the "No hardware found", while `hardwareDetailsSummaryView` and `hardwareDetailsView` (which both return data about a single hardware) should use the "Hardware not found". Add that string to the file and use it here.

backend/kernelCI_app/views/hardwareDetailsView.py
```diff
@@ -271,7 +271,7 @@ def post(self, request, hardware_id) -> Response:
 
         if len(records) == 0:
             return create_api_error_response(
-                error_message="Hardware not found",
+                error_message=ClientStrings.HARDWARE_NOT_FOUND,
```
MarceloRobert, line 274: ditto "Hardware not found" vs "No hardware found".

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
...
+    LOG_TABLE_NOT_FOUND = "Table with id 'list' not found"
+    LOG_TABLE_BODY_NOT_FOUND = "Table body not found"
+    LOG_INVALID_TABLE_FORMAT = "Invalid number of columns in table row (probably not a log website)"
+    LOG_NO_FILES_FOUND = "No log files found"    
+    ISSUE_NOT_FOUND = "No issues found"
+    ISSUE_INVALID_JSON = "Invalid body, request body must be a valid json string"
+    ISSUE_EMPTY_LIST = "Invalid body, the issue list must not be empty"
+    ISSUE_NO_EXTRA_DETAILS = "No extra details found"
+    ISSUE_TEST_NOT_FOUND = "No tests found for this issue"
+    ISSUE_DETAILS_BUILD_NOT_FOUND = "No builds found for this issue"
```
MarceloRobert, line 33: To keep the keys consistency, you don't need to use `DETAILS`, and use the plural for tests and builds (the endpoints return multiple tests/builds for a single issue)
```suggestion
    ISSUE_TESTS_NOT_FOUND = "No tests found for this issue"
    ISSUE_BUILDS_NOT_FOUND = "No builds found for this issue"
```

backend/kernelCI_app/views/issueDetailsView.py
```diff
@@ -46,7 +46,7 @@ def get(self, _request, issue_id: Optional[str]) -> Response:
 
         if not issue_data:
             return create_api_error_response(
-                error_message="Issue not found", status_code=HTTPStatus.OK
+                error_message=ClientStrings.ISSUE_NOT_FOUND, status_code=HTTPStatus.OK
```
MarceloRobert, line 49: This is not the correct string to use, similar to the difference between `hardwareView` and `hardwareDetailsView`. Instead of using "No issues found" here, use "Issue not found"

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
...
+    PROXY_FETCH_FAILED  = "Failed to fetch resource:"
+    PROXY_INVALID_URL = "Invalid URL"
+    PROXY_ERROR_FETCH = "Error fetching the resource"
+    ORIGIN_NOT_FOUND = "No origins found"
+    LOG_TABLE_NOT_FOUND = "Table with id 'list' not found"
+    LOG_TABLE_BODY_NOT_FOUND = "Table body not found"
+    LOG_INVALID_TABLE_FORMAT = "Invalid number of columns in table row (probably not a log website)"
+    LOG_NO_FILES_FOUND = "No log files found"    
+    ISSUE_NOT_FOUND = "No issues found"
+    ISSUE_INVALID_JSON = "Invalid body, request body must be a valid json string"
```
MarceloRobert, line 29: This is not really related to the issue details, and it is also shared by the hardware endpoint, so you can remove the `ISSUE_` prefix and use it on all endpoints that use this string
```suggestion
    INVALID_JSON_BODY = "Invalid body, request body must be a valid json string"
```

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
     TREE_BOOTS_NOT_FOUND = "No boots available for this tree/branch/commit"
     TREE_NOT_FOUND = "No results available for this tree/branch/commit"
     TREE_TESTS_NOT_FOUND = "No tests available for this tree/branch/commit"
+    TREE_COMMITS_HISTORY_NOT_FOUND = "History of tree commits not found"
+    TREE_STATUS_HISTORY_NOT_FOUND = "Test status history not found"
+    TEST_NO_ISSUE_FOUND = "No issues were found for this test"
```
MarceloRobert, line 18: For consistency,
```suggestion
    TEST_ISSUES_NOT_FOUND = "No issues were found for this test"
```

backend/kernelCI_app/constants/localization.py
```diff
@@ -13,3 +13,35 @@ class ClientStrings:
     TREE_BOOTS_NOT_FOUND = "No boots available for this tree/branch/commit"
     TREE_NOT_FOUND = "No results available for this tree/branch/commit"
     TREE_TESTS_NOT_FOUND = "No tests available for this tree/branch/commit"
+    TREE_COMMITS_HISTORY_NOT_FOUND = "History of tree commits not found"
+    TREE_STATUS_HISTORY_NOT_FOUND = "Test status history not found"
```
MarceloRobert, line 17: Wrong T word :sweat_smile: 
```suggestion
    TEST_STATUS_HISTORY_NOT_FOUND = "Test status history not found"
```

backend/kernelCI_app/views/treeView.py
```diff
@@ -57,7 +57,7 @@ def get(self, request: HttpRequest) -> Response:
 
         if not cached_checkouts and not kcidb_checkouts:
             return create_api_error_response(
-                error_message="Trees not found", status_code=HTTPStatus.OK
+                error_message=ClientStrings.TREE_LATEST_NOT_FOUND, status_code=HTTPStatus.OK
```
MarceloRobert, line 60: This one is a mistake by our part, this endpoint should also follow the "X not found" vs "No X found" idea. The treeView and treeView fast return a list of trees, so in this case a new localization string should be created, for "No trees were found".

backend/kernelCI_app/views/treeViewFast.py
```diff
@@ -40,7 +40,7 @@ def get(self, request: HttpRequest) -> Response:
 
         if not checkouts:
             return create_api_error_response(
-                error_message="Trees not found", status_code=HTTPStatus.OK
+                error_message= ClientStrings.TREE_LATEST_NOT_FOUND, status_code=HTTPStatus.OK
```
MarceloRobert, line 43: ditto "No trees were found".

MarceloRobert (commented):
Thanks for the contribution, some changes though

khanheric2003:
I accidently removed the pr, let me remake it

PR #1318, opened 2025-06-27
feat(database): separate SQLite databases

### LINKED ISSUE #1223: Separate sqlite databases
The tables that we have for data caching are inherently different from tables with other information, since we can have multiple tables of cache and multiple tables of other data, we should separate them by databases instead.
This would mean 1 database for cached data, 1 database for notifications and others to come.

danieldiasbarbo:
## Description
This PR separates the SQLite databases used by the backend to better reflect the purpose and lifecycle of different types of data.
It also implements the automatic migration of notifications data from the cache database to the new notification database, as part of the multi-database separation process.

## Changes
- [x] Created a new sqlite database called notification 
- [x] Redirected the older references to cache to the new notification database
- [x] Created a migration to transfer data from cache database to notication database if it is possible

## How to test
1. Run the migration `migrate-cache-db.sh` in different scenarios:
    1.1. Legacy schema: cache DB exists with notifications data, and the notification DB does not exist.
    1.2. Partial schema: cache DB exists but is empty, and the notification DB does not exist.
    1.3. Fresh setup: neither cache nor notification tables exist.
1. Verify that `volume_data/cache.sqlite3` and `volume_data/notifications.sqlite3` databases is created
1. Start the backend server and perform a quick sanity check:

Closes #1223

### DIFF
--- a/backend/kernelCI_cache/migrations/0011_copy_data_from_cache_if_exists.py
+++ b/backend/kernelCI_cache/migrations/0011_copy_data_from_cache_if_exists.py
@@ -1,0 +1,125 @@
+from kernelCI.settings import DATABASES
+from kernelCI_app.helpers.discordWebhook import send_discord_notification
+from django.db import migrations, connections
+
+cache_path = DATABASES["cache"]["NAME"]
+
+
+def _copy_if_data_not_exists(
+    cache_cursor, notif_cursor, table_name: str, columns: list[str]
+):
+    notif_cursor.execute(f"""SELECT id FROM {table_name} LIMIT 1;""")
+    existent_data = notif_cursor.fetchone()
+    if not existent_data:
+        print(f"Copying data from {table_name}")
+        query_placeholder = ", ".join(["?"] * len(columns))
+        columns_names = ", ".join(columns)
+        cache_cursor.execute(
+            f"""SELECT {columns_names}
+            FROM {table_name};"""
+        )
+        rows = cache_cursor.fetchall()
+        if rows:
+            notif_cursor.executemany(
+                f"""INSERT INTO {table_name}
+                ({columns_names})
+                VALUES ({query_placeholder})""",
+                rows,
+            )
+    else:
+        print(
+            f"Skipping {table_name} migration: "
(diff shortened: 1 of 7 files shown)


backend/kernelCI_cache/migrations/0011_copy_data_from_cache_if_exists.py
```diff
@@ -0,0 +1,103 @@
...
+                    notif_cursor.executemany(
+                        """INSERT INTO notifications_issue
+                        (id, notification_message_id, notification_sent,
+                        issue_id, issue_version, issue_type)
+                        VALUES (?, ?, ?, ?, ?, ?)""",
+                        rows,
+                    )
+        except Exception as e:
+            print(
+                f"\nError during data transfer migration: {e}. Ignoring this migration."
```
WilsonNet, line 89: Maybe send to the discord this one using the discord hook, so we can easily know that the migration is breaking

backend/kernelCI_cache/migrations/0011_copy_data_from_cache_if_exists.py
MarceloRobert, line 1: The comments are a bit all over the place IMO, I'm ok with leaving them to explain the logic but be clear when it is talking about tables or dbs, and I think you can combine some of them

backend/kernelCI_cache/migrations/0011_copy_data_from_cache_if_exists.py
```diff
@@ -0,0 +1,103 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("kernelCI_cache", "0010_notificationsissue"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            copy_from_cache_to_notification_tables, hints={"run_always": True}
```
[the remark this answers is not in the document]
MarceloRobert: It has to do with the `allow_migrate`, which Daniel changed to block migrations that don't specify a model and where it should go. This flag is then needed so that the router knows that this migration should run even if it doesn't specify models or dbs.

PR #1323, opened 2025-07-01
workflows: Add separate staging deployment (manual)

nuclearcat:
As we might develop breaking features, we have separate database, dashboard and kcidb-ng instance running on staging server. We need also suitable way to deploy dashboard there using github workflows.


MarceloRobert (approved):
Seems good to me

PR #1324, opened 2025-07-01
Feat: add checkout summary endpoint

### LINKED ISSUE #1321: Make summary notification an API endpoint
The same data that is generated in the checkout summary notification could also be returned as a new endpoint.

It should be planned how this would integrate/interact with the other structures and endpoints that we already have.

In this endpoint, it could also allow for more variables, such as amount of tests in a history, time limit, and maybe others.

MarceloRobert:
The endpoint is meant to return the same data as the checkout summary notification command. However, this endpoint returns the data for a single tree.

## Changes
- Added new endpoint that queries the same data as the notification would query.

## How to test
This endpoint is backend-only, meaning that it is not called by the frontend of the dashboard. To test it, you can use the swagger UI or send requests directly to the url.
Test with some trees (branch + url), as well as different paths and origins.
A tree that currently returns inconclusive tests is branch `for-kernelci` url `https://git.kernel.org/pub/scm/linux/kernel/git/arm64/linux.git`. An example of the response can be seen in the requests folder.

Closes #1321

### DIFF
--- a/backend/kernelCI_app/constants/localization.py
+++ b/backend/kernelCI_app/constants/localization.py
@@ -27,6 +27,7 @@ class DocStrings:
 
     PROXY_URL_DESCRIPTION = "URL to the proxy"
 
+    CHECKOUT_START_TIME_DESCRIPTION = "Start time of the checkout"
     COMMIT_HASH_PATH_DESCRIPTION = "Commit hash of the tree"
     TREE_NAME_PATH_DESCRIPTION = "Name of the tree"
     BUILD_ID_PATH_DESCRIPTION = "ID of the build"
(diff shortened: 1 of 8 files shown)


backend/kernelCI_app/views/kciSummaryView.py
```diff
@@ -0,0 +1,94 @@
...
+
+        dashboard_url = f"""https://d.kernelci.org/tree/{commit_hash}
+            ?ti%7Cc={checkout.git_commit_name}
+            &ti%7Cch={commit_hash}
+            &ti%7Cgb={branch}
+            &ti%7Cgu={git_url_safe}
+            &ti%7Ct={tree_name}
+            &o={origin}"""
+        if tree_name and branch:
+            dashboard_url = f"https://d.kernelci.org/tree/{tree_name}/{branch}/{commit_hash}?o={origin}"
```
giubrocchi, line 66: What do you think of swaping these to make clear the condition to assign either values?

```suggestion
         if tree_name and branch:
            dashboard_url = f"https://d.kernelci.org/tree/{tree_name}/{branch}/{commit_hash}?o={origin}"
        else:
              dashboard_url = f"""https://d.kernelci.org/tree/{commit_hash}
                  ?ti%7Cc={checkout.git_commit_name}
                  &ti%7Cch={commit_hash}
                  &ti%7Cgb={branch}
                  &ti%7Cgu={git_url_safe}
                  &ti%7Ct={tree_name}
                  &o={origin}"""
```

backend/kernelCI_app/constants/localization.py
```diff
@@ -77,3 +78,25 @@ class DocStrings:
...
+    BOOT_STATUS_SUMMARY_DESCRIPTION = "Summary of boot test statuses"
+    TEST_STATUS_SUMMARY_DESCRIPTION = "Summary of test statuses"
+
+    # KCI Summary related descriptions
+    REGRESSIONS_GROUP = "Regressions are grouped by hardware, config, and path."
+    KCI_SUMMARY_PATH_DESCRIPTION = (
+        "The test path to query for. SQL Wildcard can be used."
+    )
+    KCI_SUMMARY_DASHBOARD_URL_DESCRIPTION = (
+        "The url of this checkout accessible in the dashboard"
```
giubrocchi, line 92: This might sound a bit confusing, how about `The dashboard url for the checkout`?

backend/kernelCI_app/views/kciSummaryView.py
MarceloRobert: I usually prefer "default + if" instead of "if/else" but it works too. I guess I'm thinking about variable scope too much :sweat_smile:

PR #1325, opened 2025-07-03
Feat: multiple path report

### LINKED ISSUE #1217: Add support for multiple paths in a checkout report
Currently a checkout report can only show the results for one path.

It should be possible to check on multiple paths of tests in the same reports.
This could be done by adding those paths directly in sql, splitting them into different parts in python

MarceloRobert:
The approach used for querying for different like clauses was using `LIKE ANY (ARRAY[])`, I'm open to other approaches.

## Changes
- Changes the use of paths from a single string to an array of strings, allowing for multiple paths to be used _in a single report_
- Changed the path parameter of kciSummary endpoint to receive an array of paths as well

## How to test
For the notification command, change some tree file to use an array of paths (isolating the file is ideal for debugging) and run the notification command. Be aware of the intervals in the queries as well.
Also check the kciSummary endpoint, passing a tree and multiple paths and checking the return.

Closes #1217

### DIFF
--- a/backend/kernelCI_app/queries/notifications.py
+++ b/backend/kernelCI_app/queries/notifications.py
@@ -451,27 +451,62 @@ def get_checkout_summary_data(
         return dict_fetchall(cursor=cursor)
 
 
-def kcidb_tests_results(origin, giturl, branch, hash, path, interval, group_size):
-    """Fetches the 5-test history of searched tests in a specific tree.
+def kcidb_tests_results(
+    *,
+    origin: str,
+    giturl: str,
+    branch: str,
+    hash: str,
+    paths: list[str],
+    interval: str,
+    group_size: int,
+):
+    """Fetches the n-test history of searched tests in a specific tree.
 
     This query:
         - Gathers data for the tree (origin, url, branch) at every checkout prior to the searched hash;
         - Adds a row number to the data grouping by test path, test platform and build config_name,
           ordering by the test start_time;
-        - Limits to 5 tests on that history grouping;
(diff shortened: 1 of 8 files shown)


backend/kernelCI_app/queries/notifications.py
```diff
@@ -451,27 +451,63 @@ def get_checkout_summary_data(
...
+    *,
+    origin: str,
+    giturl: str,
+    branch: str,
+    hash: str,
+    paths: list[str],
+    interval: str,
+    group_size: int,
+):
     """Fetches the 5-test history of searched tests in a specific tree.
```
giubrocchi, line 464:
```suggestion
    """Fetches the x-test history of searched tests in a specific tree.
```

backend/kernelCI_app/queries/notifications.py
```diff
@@ -451,27 +451,62 @@ def get_checkout_summary_data(
...
+        path_params["path"] = paths[0]
+    else:
+        path_clause_keys = []
+        for idx, path in enumerate(paths):
+            query_key = f"path_{idx}"
+            path_params[query_key] = path
+            path_clause_keys.append(f"%({query_key})s")
+        path_clause_keys_full = " OR t.path LIKE ".join(path_clause_keys)
+
+        path_clause = "AND (t.path LIKE " + path_clause_keys_full + ")"
```
danieldiasbarbo, line 497: Nit: Do you see this kind of functionality being used in other endpoints in the future? If so, it might be worth extracting it to a helper or something similar for reuse.

danieldiasbarbo (approved):
LGTM!

PR #1327, opened 2025-07-04
Fix: avoid backend exit

### LINKED ISSUE #1328: Avoid backend exit and return correct error
The kcidb_execute_query has a `sys.exit` in it, but the tests query is now being used in a backend endpoint. If some error occurs in the query, it should be returned to the user instead of simply exiting the backend.

MarceloRobert:
The kcidb_execute_query allows for a sys.exit since it was only used in a command context. Since it is now used in the full backend context, it should be avoided in order to allow for an error 500 instead of exiting the backend

## Changes
- Replaced `kcidb_execute_query` with the usual cursor and dictfetchall

## How to test
- Go to the kciSummary endpoint and compare queries on staging and locally, they should both still work
- Also compare the notifications command before and after the change

Closes #1328

### DIFF
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -484,15 +484,21 @@ def run_checkout_summary(
         for tree_report in tree_report_list:
             path = tree_report["path"] if "path" in tree_report else "%"
 
-            new_issues, fixed_issues, unstable_tests = evaluate_test_results(
-                origin=origin,
-                giturl=giturl,
-                branch=branch,
-                commit_hash=commit_hash,
-                path=path,
-                interval="7 days",
-                group_size=5,
-            )
+            # In case an error happens in the query, we don't want to send an empty report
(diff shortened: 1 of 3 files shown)


gabrielBatistella (approved):

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -431,6 +431,22 @@ def evaluate_test_results(
...
+    *,
+    default_recipients: list[str],
+    specific_recipients: list[str],
+    options: list[PossibleReportOptions],
+) -> list[str]:
+    recipients = default_recipients + specific_recipients
+
+    for option in options:
+        match option:
+            case "ignore_default_recipients":
```
giubrocchi, line 444: Can we simplify this to:
```suggestion
    if "ignore_default_recipients" in options:
```

MarceloRobert: Done

PR #1343, opened 2025-07-10
Refactor: sync django models

### LINKED ISSUE #1342: Sync django models with kcidb
Some columns are missing, and the status column is not well defined. In order to move to a django-owned schema, first we should sync it completely with the kcidb schema.

MarceloRobert:
Most models are not following the kcidb schema 1-1. Some fields were not being used at all, while others needed an update with their type.

One asterisk is still the issue_version, which in kcidb has a primary_key constraint, but the django version used in the dashboard doesn't support it. We could either update django and update the model, or keep using only the 'unique_together' constraint until the issue version gets a refactor.

This PR removes the build valid field, removing all the conditional triggering between 'valid' and 'status' as well.

## Changes
- Updated all models to follow kcidb
  - Added `git_commit_message` and `git_repository_branch_tip` to Checkouts
  - Removed `valid` from Builds
  - Added `number_prefix`, `number_unit` and `number_value` to Tests
  - Renamed `issue`, `build` and `test` to `issue_id`, `build_id` and `test_id` on Incidents
- Removed build valid conditionals on all queries that used them
- Replaced build_status_map with other ways to define the NULL value to a null status

## How to test
Go around the entire dashboard, checking specially for null values on status.

Closes #1342


backend/kernelCI_app/models.py
```diff
@@ -115,10 +137,14 @@ class Incidents(models.Model):
...
+    issue_id = models.ForeignKey(Issues, on_delete=models.DO_NOTHING)
     issue_version = models.IntegerField()
-    build = models.ForeignKey(Builds, on_delete=models.DO_NOTHING)
-    test = models.ForeignKey(Tests, on_delete=models.DO_NOTHING)
+    build_id = models.ForeignKey(
+        Builds, on_delete=models.DO_NOTHING, null=True, blank=True
+    )
+    test_id = models.ForeignKey(
+        Tests, on_delete=models.DO_NOTHING, null=True, blank=True
+    )
```
danieldiasbarbo, line 148: If the Copilot comment applies, ditto here.

backend/kernelCI_app/helpers/hardwareDetails.py
```diff
@@ -262,7 +261,7 @@ def handle_tree_status_summary(
     tree_status_summary[tree_index][test_type_key][record["status"]] += 1
 
     if record["build_id"] not in processed_builds:
-        build_status = build_status_map(record["build__status"])
+        build_status = record.get("build__status", NULL_STATUS)
```
danieldiasbarbo, line 264: Question: The "build__status" is with double underscores?

backend/kernelCI_app/models.py
```diff
@@ -72,37 +83,49 @@ class Builds(models.Model):
...
     start_time = models.DateTimeField(blank=True, null=True)
     duration = models.FloatField(blank=True, null=True)
     output_files = models.JSONField(blank=True, null=True)
-    input_files = models.JSONField(blank=True, null=True)
     misc = models.JSONField(blank=True, null=True)
+    number_value = models.FloatField(blank=True, null=True)
+    environment_compatible = ArrayField(
+        models.TextField(), blank=True, default=list, null=True
+    )
```
danieldiasbarbo, line 123: Since we had problem with default before. Here it will behave as expected? It seems odd that it has a default value list() and can be nullable at the same time.

backend/kernelCI_app/queries/build.py
```diff
@@ -48,38 +41,18 @@ def get_build_details(build_id: str) -> Optional[list[dict]]:
...
-            raise
+    result = Tests.objects.filter(build_id=build_id).values(
+        "id",
+        "duration",
+        "status",
+        "path",
+        "start_time",
+        "environment_compatible",
+        "environment_misc",
+        "build__status",
```
danieldiasbarbo, line 56: Ditto here about the double underscores. Just checking if it is expected to be this way.

danieldiasbarbo (approved):
Found no defects, only some questions.

backend/kernelCI_app/models.py
MarceloRobert: We don't use the Incidents model directly, when we query for it we are using direct raw sql

backend/kernelCI_app/models.py
MarceloRobert: Huh, I just moved this field so I didn't notice it, but you're right, there shouldn't be a default here, I'll make the change

PR #1348, opened 2025-07-15
backend: notifications: Replace Gmail API with SMTP email backend

padovan:
Migrate email functionality from Gmail API to Django's SMTP backend for improved maintainability and simpler deployment. Remove Google API dependencies and custom line wrapping functionality.

It adds SMTP configuration settings (EMAIL_HOST, EMAIL_PORT, etc.).


docs/notifications.md
```diff
@@ -40,7 +40,43 @@ The command supports four primary actions:
...
+The notification system uses Django's SMTP backend to send emails. Configure the following environment variables:
+
+### SMTP Settings
+
+```bash
+# Email backend configuration
+export EMAIL_BACKEND="django.core.mail.backends.smtp.EmailBackend"
+export EMAIL_HOST="smtp.gmail.com"
+export EMAIL_PORT=587
+export EMAIL_USE_TLS=True
```
MarceloRobert, line 56: If I understood correctly, these first four arguments don't need to be changed (email_backend, email_host, email_port, email_use_tls), specially considering that they will be assigned a default value in `settings.py`. Perhaps you could indicate that they are optional in this doc.

MarceloRobert (approved):
I noticed that when the email arguments are not properly set (such as missing password), the command will fail only when it tries to send the email, in contrast to previously where it would fail on the command start. Either way, there is a print and a raise for such problems, I don't think they will be hard to spot, so I'll approve.

MarceloRobert:
BTW, much easier setup now, really cool

PR #1355, opened 2025-07-17
feat(treedetails): allow lookup by commit tag or hash

### LINKED ISSUE #1320: Use commit tag along commit hash
We use commit hash for the treeDetails page and endpoint now; it would be useful if it was also possible to use the tags as a search field.

It would be possible to do a /tree/treeName/gitBranch/gitIdentifier, where the identifier would be either hash or id. Since they usually have different looks, it would hardly happen a conflict. However, a conflict would still be possible, and in that case some possibilities can be studied, one of them being just giving priority to one or the other.

The commit hash is indeed useful for some cases such as the commit history, that should be accounted for.

danieldiasbarbo:
## Description
This PR updates the TreeDetails endpoint to support resolving trees not only by commit hash but also by commit tag.
Previously, the URL structure required a commit hash to identify the tree. With this change, it is now possible to access the same endpoint using a tag name, improving flexibility and usability for users who rely on tags.
No conflict resolution logic between hashes and tags was introduced since they are very different from each other (I've queried in the database to confirm this).
There are cases where the checkout constraints in KCIDB yield multiple matches. In such scenarios, the data returned is based on the latest associated checkout.

## Changes
- [x] Modified TreeDetails query to resolve using either a commit hash or a tag with a matching name.
- [x] The priority logic implemented in colisions regarding the constraints is that the information about the latest checkout will be returned.

## How to test
1. Use the TreeDetails page with a known commit hash. It should behave as before. e.g. `/tree/stable/linux-5.15.y/89950c4542652dfe435f9519a5080f7d2128764c`
1. Use the same page with a known commit tag, it should resolve the associated tree and load the same page. e.g. `/tree/stable/linux-5.15.y/v5.15.188`
1. Try with an invalid identifier. It still display an No results were found
1. If a hash and a tag share the same name, the behavior is undefined (and not currently handled).

Closes #1320

### DIFF
--- a/backend/kernelCI_app/queries/tree.py
+++ b/backend/kernelCI_app/queries/tree.py
@@ -1,5 +1,6 @@
 from typing import Optional
 from django.db import connection
+from django.db.models import Q
 
 from kernelCI_app.helpers.database import dict_fetchall
 from kernelCI_app.models import Checkouts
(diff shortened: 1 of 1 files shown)


WilsonNet (approved):
looks good

giubrocchi (approved):

MarceloRobert (approved):
Working well on the backend. I personally would like to see some changes for the frontend to call those endpoints using the tags as well, but we can discuss the need for this later.

PR #1357, opened 2025-07-18
backend: notifications: Add test status history to test reports

padovan:
Add status history display to test regression reports using emoji symbols. The history shows the last 10 test results for the same test path, platform, and configuration, displayed as: ✅ → ✅ → ✅ → ✅ → ✅ → ✅ → ✅ → ✅ → ✅ → ❌

### DIFF
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -36,7 +36,7 @@ from kernelCI_app.queries.notifications import (
     kcidb_last_test_without_issue,
     kcidb_tests_results,
 )
-from kernelCI_app.queries.test import get_test_details_data
+from kernelCI_app.queries.test import get_test_details_data, get_test_status_history
 from kernelCI_cache.queries.notifications import (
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -589,6 +589,33 @@ def generate_test_report(*, service, test_id, email_args, signup_folder):
...
+        if entry["status"] == "PASS":
+            status_symbols.append("✅")
+        elif entry["status"] == "FAIL":
+            status_symbols.append("❌")
+        else:
+            status_symbols.append("⚠️")
+
+    # Reverse to show oldest first
+    status_symbols.reverse()
+    test["status_history"] = " → ".join(status_symbols)
```
MarceloRobert, line 617: nit: I think it would be better to add a new argument in the jinja template instead of adding more properties to the test dict; otherwise we could start to lose track of the original vs added properties of the test dict

MarceloRobert (approved):
Only some minor comments, I'll pre-approve

PR #1358, opened 2025-07-18
refactor(tree): separate views and parameters per endpoint

### LINKED ISSUE #1317: Refactor /tree endpoints to separate views and parameters per route
Currently, multiple `/tree` endpoints share the same view in the backend, which leads to the reuse of `typeModels` and `parameters` in a generic way. However, depending on the specific route, some parameters are required while others are not, creating inconsistencies in both logic and API documentation.

This issue proposes a refactor to:
1. Create a dedicated view for each `/tree` endpoint
1. Define explicit query and path parameters for each view
1. Ensure that all parameter parsing and validation is handled via `Pydantic`
1. Improve `Swagger` documentation generated by `drf_spectacular` to accurately reflect required and optional fields per endpoint

The goal is to improve clarity, consistency, and reliability of the API, as well as to enhance the quality and accuracy of the documentation for external consumers.

danieldiasbarbo:
## Description
This PR refactors the `/tree` endpoints to use dedicated views for each specific route, replacing the previously shared view logic.
Previously, multiple `/tree` endpoints shared the same view, which led to generic and inconsistent parameter handling. Some parameters were required in certain contexts but not in others, which complicated both logic and documentation.

## Changes
- [x] Created a dedicated view for each endpoint;
- [x] Defined route-specific `Query` and `Path` parameters
- [x] Updated schema to reflect required/optional parameters properly

## How to test
1. Explore the /tree/... endpoints and verify that:
    1.1. Parameters are clearly defined and validated
    1.2. Required fields are marked correctly
    1.3. Descriptions are present and accurate
1. Make API requests to different `/tree` endpoints to confirm they work independently and return the expected data.
1. Navigate through the website and check if no errors are returned in the back-end.

Closes #1317

### DIFF
--- a/backend/kernelCI_app/views/treeCommitsHistory.py
+++ b/backend/kernelCI_app/views/treeCommitsHistory.py
@@ -15,6 +15,11 @@ from kernelCI_app.helpers.misc import (
     build_misc_value_or_default,
     env_misc_value_or_default,
 )
+from kernelCI_app.typeModels.commonOpenApiParameters import (
+    COMMIT_HASH_PATH_PARAM,
+    GIT_BRANCH_PATH_PARAM,
+    TREE_NAME_PATH_PARAM,
+)
 from kernelCI_app.typeModels.databases import FAIL_STATUS, NULL_STATUS, StatusValues
 from kernelCI_app.utils import is_boot
 from rest_framework.views import APIView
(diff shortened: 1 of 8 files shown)


backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -387,7 +388,7 @@ def get(
         try:
             params = TreeCommitsQueryParameters(
                 origin=request.GET.get("origin"),
-                git_url=request.GET.get("git_url"),
+                git_url=request.GET.get("git_url", ""),
```
MarceloRobert, line 391: Leave a comment on why you are giving a default value here but not on the other parameters. It might be confusing to use an empty string as a default value

backend/schema.yml
MarceloRobert, line 1: Update the schema right before merging, this will cover any small changes from the reviews

MarceloRobert (approved):
LGTM!

PR #1367, opened 2025-07-25
Modification to subscribe KernelCI notifications

johnsongeorge-w:


nuclearcat (approved):
looks good

backend/data/notifications/subscriptions/stable.yaml
```diff
@@ -2,6 +2,7 @@ stable:
   url: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git
   default_recipients:
     - stable@vger.kernel.org
+    - kernelcialerts@microsoft.com
```
padovan, line 5: This will get you notifications not only from microsoft, but for all origins. Maybe we can extend the notification system to do that. I'll check on it on Monday.

johnsongeorge-w: Thank you for the information. Currently enabling it for hyperv, will enable for stable later.

padovan (approved):

PR #1370, opened 2025-07-28
Feat: Add env var to switch to new db

MarceloRobert:
## Changes
- Adds an environment variable to quick switch between using kcidb and using the new dashboard_db
- Updates de `update_db` management command to make use of this new variable
- Adds a 0-results fallback in the select methods of `update_db` so that they don't try to make the query when they don't need to

## How to test
If you are running in docker, simply start the dashboard without the `USE_NEW_DB` env var set, check that it is using kcidb, then set it to True in the .env and restart the dashboard; it should now use the local database.
If you are running locally, set the `DASH_DB_` variables as they appear in the backend service of the docker compose, then also try out the `USE_NEW_DB` variable.
Also check if the `update_db` command still works on both scenarios

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -193,43 +193,62 @@ BACKEND_VOLUME_DIR = get_json_env_var("BACKEND_VOLUME_DIR", "volume_data")
 
 DATABASE_ROUTERS = ["kernelCI_app.routers.databaseRouter.DatabaseRouter"]
 
-DATABASES = {
-    "default": get_json_env_var(
-        "DB_DEFAULT",
-        {
-            "ENGINE": "django.db.backends.postgresql",
-            "NAME": "kernelci",
-            "USER": "kernelci",
-            "PASSWORD": "kernelci-db-password",
-            "HOST": "127.0.0.1",
-            "OPTIONS": {
-                "connect_timeout": 5,
-            },
-        },
-    ),
-    "dashboard_db": {
+# Database definition configurations
+kcidb_config = get_json_env_var(
+    "DB_DEFAULT",
+    {
         "ENGINE": "django.db.backends.postgresql",
-        "NAME": "dashboard",
-        "USER": get_json_env_var("DASH_DB_USER", "dev"),
-        "PASSWORD": get_json_env_var("DASH_DB_PASSWORD", "dev"),
-        "HOST": get_json_env_var("DASH_DB_HOST", "127.0.0.1"),
-        "PORT": get_json_env_var("DASH_DB_PORT", 5434),
+        "NAME": "kernelci",
+        "USER": "kernelci",
+        "PASSWORD": "kernelci-db-password",
+        "HOST": "127.0.0.1",
         "OPTIONS": {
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -85,6 +85,7 @@ services:
             - DASH_DB_PASSWORD=${DASH_DB_PASSWORD:-admin}
             - DASH_DB_HOST=dashboard_db
             - DASH_DB_PORT=${DASH_DB_PORT:-5432}
+            - USE_NEW_DB=${USE_NEW_DB:-False}
```
giubrocchi, line 88: We can mention on the readme how to run with the new DB (settings the envs) and what changes when running with new DB vs the kcidb

giubrocchi (approved):

backend/kernelCI/settings.py
```diff
@@ -193,43 +193,62 @@ def get_json_env_var(name, default):
...
+    "NAME": os.path.join(BACKEND_VOLUME_DIR, "cache.sqlite3"),
+}
+
+notifications_db_config = {
+    "ENGINE": "django.db.backends.sqlite3",
+    "NAME": os.path.join(BACKEND_VOLUME_DIR, "notifications.sqlite3"),
+}
+
+USE_NEW_DB = os.environ.get("USE_NEW_DB", False)
+if USE_NEW_DB is True:
```
danieldiasbarbo: I think it is better this way:
```suggestion
USE_NEW_DB = os.environ.get("USE_NEW_DB", False) == "True"
if USE_NEW_DB:
```

backend/kernelCI/settings.py
```diff
@@ -193,43 +193,63 @@ def get_json_env_var(name, default):
...
+    "NAME": os.path.join(BACKEND_VOLUME_DIR, "cache.sqlite3"),
+}
+
+notifications_db_config = {
+    "ENGINE": "django.db.backends.sqlite3",
+    "NAME": os.path.join(BACKEND_VOLUME_DIR, "notifications.sqlite3"),
+}
+
+USE_DASHBOARD_DB = os.environ.get("USE_DASHBOARD_DB", False)
+if USE_DASHBOARD_DB == "True":
```
giubrocchi, line 235: nit:
```suggestion
if str(USE_DASHBOARD_DB).lower() == "true":
```

backend/kernelCI/settings.py
```diff
@@ -193,43 +193,63 @@ def get_json_env_var(name, default):
...
+    print("IS TRUE")
```
giubrocchi, line 236: Remember to remove before merging

backend/kernelCI/settings.py
giubrocchi: There's actually a function `is_boolean_or_string_true` that you can use

MarceloRobert: Done, thanks for pointing it out

backend/kernelCI/settings.py
MarceloRobert: Done, I had forgotten about this one :sweat_smile:

PR #1374, opened 2025-07-29
Feat: add min/max age to tree-report

### LINKED ISSUE #1356: Add days parameter to the tree-report endpoint
We should allow the user to change the SQL interval param (up to a max of x days). This will help with some interesting test results evaluation, even though it may come with slow replies.

MarceloRobert:
## Changes
- Exposed time limit to the tree report endpoint queries in the form of min and max age in hours. The limit for max age is 720 hours (30 days) in order to avoid responses way too long.
- Small changes to the returned strings

## How to test
Go to the [swagger ui](http://localhost:8000/api/schema/swagger-ui/#/tree-report/tree_report_retrieve) (or make requests by hand) and fetch the tree report endpoint for a tree in a certain period.

Closes #1356

### DIFF
--- a/backend/kernelCI_app/views/treeReportView.py
+++ b/backend/kernelCI_app/views/treeReportView.py
@@ -33,6 +33,8 @@ class TreeReport(APIView):
                 git_url=request.GET.get("git_url"),
                 path=request.GET.getlist("path"),
                 group_size=request.GET.get("group_size"),
+                min_age_in_hours=request.GET.get("min_age_in_hours"),
+                max_age_in_hours=request.GET.get("max_age_in_hours"),
             )
             origin = params.origin
             git_url = params.git_url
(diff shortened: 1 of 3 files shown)


backend/kernelCI_app/views/treeReportView.py
```diff
@@ -33,24 +33,35 @@ def get(self, request: HttpRequest):
...
+                min_age_in_hours=request.GET.get("min_age_in_hours"),
+                max_age_in_hours=request.GET.get("max_age_in_hours"),
             )
             origin = params.origin
             git_url = params.git_url
             git_branch = params.git_branch
         except ValidationError as e:
             return create_api_error_response(error_message=e.json())
 
+        if params.min_age_in_hours > params.max_age_in_hours:
```
giubrocchi, line 45: It seems like if `min_age_in_hours == max_age_in_hours` it still doesn't return results if the checkout was created on that exact hour. For example:

If the checkout was created 1 hour ago, querying with `min_age_in_hours = 0` and `max_age_in_hours = 1` returns the correct result.
However, querying with `min_age_in_hours = 1` and `max_age_in_hours = 1` doesn't return any data.

To make sure that user gets a valid response to why there isn't any data I think we can make this condition as `if params.min_age_in_hours >= params.max_age_in_hours:`

MarceloRobert: It will return 0 results because the query is done with `start_time >= now() - interval 'x hours'` (>= for max and <= for min), so if the max_age and min_age are the same, it will query for if the checkout started at this exact second but x hours ago. This is basically impossible to hit anyway, but I guess that forcing to always have an interval will be good to avoid potential user confusion, so I added the condition that min shouldn't be = max

PR #1380, opened 2025-08-01
Feat: sort summary cards by fail count

giubrocchi:
# Changes

- On the FE, sort the tree and hardware details cards (hardware, config, and architecture) by two conditions:
  - Fail count
  - Alphabetically by name
- On the BE, sort the compilers from the architectures for tree and hardware build details.

# How to test

- Navigate to the tree details page and for the build, boots, and tests tabs check the following cards:
  - Architecture cards should be ordered by fail count and alphabetically if the count is the same. If there are multiple compilers, they should also be ordered alphabetically
  - Configs and hardware cards should be ordered by fail count and alphabetically if the count is the same
- Navigate to the hardware details page and check if the same conditions apply
- Make a request to the `/tree/.../summary` and `/hardware/.../summary` endpoints and check if the compilers are alphabetically sorted

### DIFF
--- a/backend/kernelCI_app/helpers/hardwareDetails.py
+++ b/backend/kernelCI_app/helpers/hardwareDetails.py
@@ -412,6 +412,7 @@ def handle_build_summary(
             and compiler not in builds_summary.architectures.get(arch).compilers
         ):
             builds_summary.architectures[arch].compilers.append(compiler)
+            builds_summary.architectures[arch].compilers.sort()
 
     if origin := build.origin:
         build_origin_summary = builds_summary.origins.get(origin)
--- a/dashboard/src/components/Cards/HardwareTested.tsx
+++ b/dashboard/src/components/Cards/HardwareTested.tsx
@@ -1,4 +1,4 @@
-import { memo, type JSX } from 'react';
+import { memo, useMemo, type JSX } from 'react';
 
 import type { IBaseCard } from '@/components/Cards/BaseCard';
 import BaseCard from '@/components/Cards/BaseCard';
@@ -24,13 +24,26 @@ const HardwareTested = ({
   title,
   diffFilter,
 }: IHardwareTested): JSX.Element => {
+  const sortedEnvironmentCompatibleNames = useMemo(() => {
+    return Object.keys(environmentCompatible).sort((a, b) => {
+      const failA = environmentCompatible[a].FAIL ?? 0;
+      const failB = environmentCompatible[b].FAIL ?? 0;
+
+      if (failB !== failA) {
+        return failB - failA;
+      }
+
+      return a.localeCompare(b);
+    });
+  }, [environmentCompatible]);
+
   return (
     <BaseCard
       title={title}
       content={
         <ScrollArea className="h-[350px]">
           <DumbListingContent>
-            {Object.keys(environmentCompatible).map(hardwareTestedName => {
+            {sortedEnvironmentCompatibleNames.map(hardwareTestedName => {
               const { DONE, FAIL, ERROR, MISS, PASS, SKIP, NULL } =
                 environmentCompatible[hardwareTestedName];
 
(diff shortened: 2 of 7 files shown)


MarceloRobert (commented):
Two things:
1. I forgot to mention the issues card, that can also be sorted (for example in [this tree](https://staging.dashboard.kernelci.org:9000/tree/sashal-next/linus-next/046eee89a3d7b18b68d60d32ab4eef8b549ca774))
2. As an extra, would it be possible to always shove any "Unknown" value to the end of the list? The Unknown is a value that we are assigning by ourselves in the backend, so it's not a real value. If it's too time-consuming to do this then don't worry about it, we already have the Unknown in the middle of nowhere sometimes (like in the configs for the tests of [this checkout](https://staging.dashboard.kernelci.org:9000/tree/mainline/master/6bcdbd62bd56e6d7383f9e06d9d148935b3c9b73?p=t)). For the issues card the "Uncategorized" value is just like the "Unknown" in the other cards.

MarceloRobert (commented):
It feels like the sorting function could be extracted and reused... calling it like
```typescript
sortedConfigs = useMemo(
  () =>
    configs.sort((a, b) => 
      sortCard(
        a.text, b.text,
        a.errors ?? 0, b.errors ?? 0
      )
```
you could avoid dealing with different variable names by passing the values that you care about instead of the whole variable

dashboard/src/components/Cards/HardwareTested.tsx
```diff
@@ -24,13 +25,33 @@ const HardwareTested = ({
...
   diffFilter,
 }: IHardwareTested): JSX.Element => {
+  const sortedEnvironmentCompatibles = useMemo(() => {
+    return Object.keys(environmentCompatible).sort((a, b) => {
+      if (a === UNKNOWN_STRING) {
+        return 1;
+      }
+      if (b === UNKNOWN_STRING) {
+        return -1;
+      }
```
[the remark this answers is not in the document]
gabrielBatistella: I think it's not needed because the ordering is done on `Object.keys(environmentCompatible)`, so `a` and `b` will be different, given that they are keys

backend/kernelCI_app/helpers/hardwareDetails.py
```diff
@@ -412,6 +412,7 @@ def handle_build_summary(
             and compiler not in builds_summary.architectures.get(arch).compilers
         ):
             builds_summary.architectures[arch].compilers.append(compiler)
+            builds_summary.architectures[arch].compilers.sort()
```
tales-aparecida, line 415: If you want to be fancy: https://docs.python.org/3/library/bisect.html#bisect.insort

dashboard/src/components/Tabs/Tests/ConfigsList.tsx
```diff
@@ -20,12 +21,34 @@ const ConfigsList = ({
   title,
   diffFilter,
 }: IConfigList): JSX.Element => {
+  const sortedConfigStatusCounts = useMemo(
```
tales-aparecida, line 24: I suggest adding some short JSDoc on these like

```suggestion
  /**
   * Sorted with UNKNOWN_STRING last, then by descending error count, then by arch text alphabetically.
   */
  const sortedConfigStatusCounts = useMemo(
```

and what if we centralized them at `src/utils/utils.ts` or `src/utils/misc.tsx`, then we could use it with something like

```js
summaryBody.sort((a,b) => sortTextAndErrors(a.arch, b.arch))
configs.sort(sortTextAndErrors)
Object.entries(configStatusCounts).sort((a,b) => sortTextAndErrors({text: a[0], errors: a[1].FAIL}, {text: b[0], errors: b[1].FAIL}))
archCompilerErrors.sort((a, b) => sortTextAndErrors({text: a.arch, errors: a.status.FAIL}, {text: b.arch, errors: b.status.FAIL}))
```

backend/kernelCI_app/helpers/hardwareDetails.py
giubrocchi: Cool! Didn't know about this one

MarceloRobert:
Also, please rebase your branch with the latest changes in main

Issue #1351, opened 2025-07-16
Create more indexes on dashboard db


MarceloRobert:
Currently, the creation of the new dashboard db only accounts for basic indexes, while kcidb had many indexes. We don't need to copy all indexes one to one, but they will surely improve the efficiency of the queries when applied correctly.

The initial `CREATE` and `ALTER` statements for the dashboard db:

```
CREATE TABLE "checkouts" ("_timestamp" timestamp with time zone NULL, "id" text NOT NULL PRIMARY KEY, "origin" text NOT NULL, "tree_name" text NULL, "git_repository_url" text NULL, "git_commit_hash" text NULL, "git_commit_name" text NULL, "git_repository_branch" text NULL, "patchset_files" jsonb NULL, "patchset_hash" text NULL, "message_id" text NULL, "comment" text NULL, "start_time" timestamp with time zone NULL, "log_url" text NULL, "log_excerpt" varchar(16384) NULL, "valid" boolean NULL, "misc" jsonb NULL, "git_commit_message" text NULL, "git_repository_branch_tip" boolean NULL, "git_commit_tags" text[] NULL, "origin_builds_finish_time" timestamp with time zone NULL, "origin_tests_finish_time" timestamp with time zone NULL);
CREATE TABLE "builds" ("_timestamp" timestamp with time zone NULL, "id" text NOT NULL PRIMARY KEY, "origin" text NOT NULL, "comment" text NULL, "start_time" timestamp with time zone NULL, "duration" double precision NULL, "architecture" text NULL, "command" text NULL, "compiler" text NULL, "input_files" jsonb NULL, "output_files" jsonb NULL, "config_name" text NULL, "config_url" text NULL, "log_url" text NULL, "log_excerpt" varchar(16384) NULL, "misc" jsonb NULL, "status" varchar(10) NULL, "checkout_id" text NOT NULL);
CREATE TABLE "issues" ("_timestamp" timestamp with time zone NULL, "id" text NOT NULL PRIMARY KEY, "version" integer NOT NULL, "origin" text NOT NULL, "report_url" text NULL, "report_subject" text NULL, "culprit_code" boolean NULL, "culprit_tool" boolean NULL, "culprit_harness" boolean NULL, "comment" text NULL, "misc" jsonb NULL, "categories" text[] NULL);
[...]

PR #1385, opened 2025-08-04
Refactor: remove foreign key constraints

### LINKED ISSUE #1382: Remove foreign keys from new db
Following the new planning, we need to send data to the new db as soon as possible, and that includes adding unconstrained data.

MarceloRobert:
## Changes
- Changes the models to keep the foreign keys (keeps the relations and allows for queries with it) while removing the constraints themselves, allowing for insertions without relations

## How to test
- Have the dashboard_db running and execute the `migrate_app_db.sh` script (check the db name. If you have `USE_DASHBOARD_DB` set, you should execute the `migrate` command with "default" as the database)
- Insert some data into the db were the foreign keys point to nonexistent objects (only the tables builds, tests and incidents have foreign keys)

Closes #1382


backend/kernelCI_app/models.py
```diff
@@ -66,7 +66,7 @@ class Builds(models.Model):
     field_timestamp = models.DateTimeField(
         db_column="_timestamp", blank=True, null=True
     )
-    checkout = models.ForeignKey(Checkouts, on_delete=models.DO_NOTHING)
```
tales-aparecida, line 69: I thought NULL=True would suffice

tales-aparecida: Ah, nvm, I'm juts tired.

On a second look in the legacy db schema posted in https://github.com/kernelci/dashboard/issues/1351 
I realized that the fields are actually "text NOT NULL", the problem here is the existence check from the FK, and not the NOT NULL.

MarceloRobert: That's awesome, it did work. Thanks for pointing it out!

backend/kernelCI/settings.py
```diff
@@ -232,8 +232,8 @@ def get_json_env_var(name, default):
     "NAME": os.path.join(BACKEND_VOLUME_DIR, "notifications.sqlite3"),
 }
 
-USE_DASHBOARD_DB = os.environ.get("USE_DASHBOARD_DB", False)
-if is_boolean_or_string_true(USE_DASHBOARD_DB):
+USE_DASHBOARD_DB = is_boolean_or_string_true(os.environ.get("USE_DASHBOARD_DB", False))
```
tales-aparecida, line 235: Could you please change:

`dashboard_db_config["NAME"] = get_json_env_var("DASH_DB_NAME", "devdashboard)`

A few lines above? I had to patch that in staging

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -46,7 +44,7 @@ def __init__(self, *args, **kwargs):
         self.end_interval = None
         self.start_timestamp = None
         self.end_timestamp = None
-        if is_boolean_or_string_true(USE_DASHBOARD_DB):
+        if USE_DASHBOARD_DB:
```
tales-aparecida, line 47: Oh, please fix the import. Settings shouldn't be imported as a module, otherwise they will break the `--settings` CLI argument and `override_settings` in unittests

```diff
-from kernelCI.settings import USE_DASHBOARD_DB
+from django.conf import settings
```

```suggestion
        if settings.USE_DASHBOARD_DB:
```

backend/kernelCI/settings.py
tales-aparecida: tbh there's a disconnect between the docker-compose and the settings with regard to DB_DEFAULT as well

tales-aparecida:
Also, pls move backend/kernelCI_app/management/commands/update_db_7_days.sh to backend/scripts

backend/kernelCI/settings.py
MarceloRobert: the `DB_DEFAULT` variable and the bridge to docker could use some improvements, but that's something for [another issue](https://github.com/kernelci/dashboard/issues/1224)

backend/kernelCI_app/models.py
MarceloRobert: Ah, my bad for not seeing it. I made the changes

giubrocchi (approved):

tales-aparecida (approved):
lgtm

PR #1386, opened 2025-08-07
Refactor: correct issue extras response

### LINKED ISSUE #1335: Issue extras: be clear when an issue version doesn't exist
In the issue extras endpoint, the first_incident is independent of the version, but the trees and tags aren't.
If the issue_id passed exists, but the combination issue_id + issue_version doesn't, then the first_incident will return, but the rest of the data won't.

The response should be refactored such that the version data isn't returned if the issue_id + issue_version doesn't exist. The first_incident data should still be returned, along with the version of that first_incident.

MarceloRobert:
## Changes
- Adds issue version to the `first_incident` field
- Fixes problem where adding two issue versions with the same issue id would only return the data for the first one
- Altered the response of the endpoint, if an issue version exists but has no incidents, it will return an object with 0 trees; if it doesn't exist, it will return as null. More comments were added in the code.

## How to test
Fetch some data from the issue_extras endpoint (I recommend using the [localhost swagger ui](http://localhost:8000/api/schema/swagger-ui/#/issue/issue_extras_create)) and test with issues that exist, that don't exist, that have 0 and more incidents, and other cases.
Also test the dashboard for the issue listing and issue details pages, everything should still be working as before.

Closes #1335

### DIFF
--- a/backend/kernelCI_app/helpers/issueExtras.py
+++ b/backend/kernelCI_app/helpers/issueExtras.py
@@ -1,9 +1,10 @@
 from collections import defaultdict
-from typing import Dict, List, Set, Tuple
+from typing import List, Tuple
 
 from kernelCI_app.helpers.logger import log_message
 from kernelCI_app.queries.issues import get_issue_first_seen_data, get_issue_trees_data
 from kernelCI_app.typeModels.issues import (
+    ExtraIssuesData,
     IssueWithExtraInfo,
     ProcessedExtraDetailedIssues,
     TreeSetItem,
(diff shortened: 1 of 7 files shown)


backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -43,40 +43,40 @@ def assign_issue_first_seen(
...
+        versions_table = processed_issue_from_id.get("versions")
+        if versions_table is None:
+            processed_issue_from_id["versions"] = {}
+
         for version in versions_per_issue[record_issue_id]:
             processed_issue_from_version = processed_issue_from_id.get(version)
-            processed_issue_from_id["versions"] = {}
             if processed_issue_from_version is None:
                 processed_issue_from_id["versions"][version] = IssueWithExtraInfo(
                     id=record_issue_id, version=version, first_seen=first_seen
```
tales-aparecida, line 82: I don't know enough about the rest of the code, but shouldn't `processed_issue_from_id.get(version)` be `processed_issue_from_id["versions"].get(version)`?

Which in that case, I would suggest

```suggestion
        for version in versions_per_issue[record_issue_id]:
            if version not in processed_issue_from_id["versions"]:
                processed_issue_from_id["versions"][version] = IssueWithExtraInfo(
                    id=record_issue_id, version=version, first_seen=first_seen
```

Since `processed_issue_from_version` is not being used outside of the membership check

backend/kernelCI_app/queries/issues.py
```diff
@@ -296,7 +299,11 @@ def get_issue_trees_data(*, issue_key_list: list[tuple[str, int]]):
         params[version_key] = key[1]
     tuple_str = ", ".join(tuple_param_list)
 
-    # TODO: use '=' instead of 'IN' when the list has a single element
+    if len(tuple_param_list) == 1:
+        comparison = "="
```
tales-aparecida, line 303: Out of curiosity, are there unit tests covering these cases? (or these functions for that matter 😅)

tales-aparecida (changes requested):
nitpicking

backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -43,40 +43,40 @@ def assign_issue_first_seen(
...
+                git_repository_url=record["git_repository_url"],
+                git_repository_branch=record["git_repository_branch"],
+                git_commit_name=record["git_commit_name"],
+                tree_name=record["tree_name"],
+                issue_version=record["issue_version"],
             )
 
+        versions_table = processed_issue_from_id.get("versions")
+        if versions_table is None:
+            processed_issue_from_id["versions"] = {}
```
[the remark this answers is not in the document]
tales-aparecida: Actually, IIUC this is the only function filling `processed_issues_table`, in that case, just set `"versions"` above, when you create the `processed_issue_from_id` dict

tales-aparecida, line 76: wrt my other comment about

> just set "versions" above

```suggestion
        if processed_issue_from_id is None:
             first_incident = FirstIncident(
                first_seen = record["first_seen"]
                git_commit_hash=record["git_commit_hash"],
                git_repository_url=record["git_repository_url"],
                git_repository_branch=record["git_repository_branch"],
                git_commit_name=record["git_commit_name"],
                tree_name=record["tree_name"],
                issue_version=record["issue_version"],
            )
            # Create new processed issue and insert in the table
             processed_issues_table[record_issue_id] = processed_issue_from_id  = {
                 "versions": {},
                 "first_incident": first_incident
             }
```

tales-aparecida: On top of this, we could even

```diff
-        processed_issue_from_id = processed_issues_table.get(record_issue_id)
-        if processed_issue_from_id is None:
+        processed_issue_from_id = processed_issues_table.setdefault(record_issue_id, {})
+        if not processed_issue_from_id:
             first_incident = FirstIncident(
	                first_seen = record["first_seen"]
	                git_commit_hash=record["git_commit_hash"],
	                git_repository_url=record["git_repository_url"],
	                git_repository_branch=record["git_repository_branch"],
	                git_commit_name=record["git_commit_name"],
	                tree_name=record["tree_name"],
	                issue_version=record["issue_version"],
	            )
	            # Fill new processed issue
	             processed_issue_from_id | = {
	                 "versions": {},
	                 "first_incident": first_incident
	             }
```

backend/kernelCI_app/queries/issues.py
MarceloRobert: Not exactly, but we have an integration test that checks if the response from the issueExtrasView endpoint is correct

backend/kernelCI_app/helpers/issueExtras.py
MarceloRobert: I reviewed this piece of code, now it's much cleaner and I aimed to use the class properties instead of dict keys

backend/kernelCI_app/helpers/issueExtras.py
tales-aparecida: Yeah, there are some built-ins that we only learn from other people, unless we take a day to read the docs, and ain't nobody got time for that 😆

backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -91,27 +87,39 @@ def assign_issue_trees(
...
             continue
 
-        current_detailed_issue = processed_issue_from_id["versions"].get(issue_version)
+        issue_versions_map = processed_issue_from_id.versions
+
+        current_detailed_issue = issue_versions_map.get(issue_version)
         if current_detailed_issue is None:
-            log_message("Got tree for issue version that is not on the list")
-            continue
+            issue_versions_map[issue_version] = IssueWithExtraInfo(
```
tales-aparecida, line 111: I don't know enough to give some "value judgment" about moving this to here, so I'll trust you have seen this doing whatever it's supposed to do 😅

tales-aparecida (approved):

backend/kernelCI_app/helpers/issueExtras.py
MarceloRobert: The versions for found issues are initialized as None (see new lines 78 and 79). If the `current_detailed_issue` is None then this was the first iteration where we encountered this `issue_version`, so I assign an `IssueWithExtraInfo` before adding the trees and tags.

PR #1388, opened 2025-08-11
Feat: short specific tree endpoints

### LINKED ISSUE #1309: Shorten treeDetails specific endpoints
Following #1281, the specific treeDetails endpoints (treeDetails builds/boots/tests) could also be shortened.

This change would also fix #1191 since it would address the frontend and all endpoints.

MarceloRobert:
Since we now have direct urls for the main tree endpoints (treeDetails summary, full, and commits), we should also have a direct url to the specific endpoints for builds, boots and tests. Those endpoints are not used by the dashboard frontend, but can be very useful if someone is accessing the endpoints directly.

## Changes
- Added a Direct version for treeDetails builds, boots, and tests
- Added pydantic validation for the direct/non-direct views separetely, in order to be cleaner.

WIP: add integration tests for those endpoints

## How to test
Get a tree with non-null `tree_name`, `git_repository_branch` and `git_commit_hash` and check them under the new endpoints.
Also check the same trees under the old endpoints to see if they are still working.

Closes #1309


backend/kernelCI_app/views/treeDetailsBuildsView.py
```diff
@@ -116,3 +117,69 @@ def get(self, request, commit_hash: str | None) -> Response:
...
+    )
+    def get(
+        self,
+        request: HttpRequest,
+        tree_name: str,
+        git_branch: str,
+        commit_hash: str,
+    ) -> Response:
+        try:
+            params = DirectTreeQueryParameters(origin=request.GET.get("origin"))
```
danieldiasbarbo, line 141: This may not have the behavior you expect. Since the `DirectTreeQueryParameters` defines a default option. I think the appropriate way of doing it is using the [model_validate](https://docs.pydantic.dev/latest/concepts/models/#validating-data) method.
```suggestion
            params = DirectTreeQueryParameters.model_validate(request.GET.dict())
```

backend/kernelCI_app/views/treeDetailsBuildsView.py
```diff
@@ -65,23 +69,20 @@ def _sanitize_rows(self, rows):
...
-        origin_param = request.GET.get("origin")
-        git_url_param = request.GET.get("git_url")
-        git_branch_param = request.GET.get("git_branch")
-
+    def get(
+        self,
+        request: HttpRequest,
+        git_url: Optional[str],
+        tree_name: Optional[str],
+        git_branch: Optional[str],
```
danieldiasbarbo, line 77: This way you don't need to instantiate this parameters on the `super().get(...)`. You will need to reorganize the params to put all the default arguments at the end.
```suggestion
        git_url: Optional[str] = None,
        tree_name: Optional[str] = None,
        git_branch: Optional[str] = None,
```

backend/kernelCI_app/views/treeDetailsBuildsView.py
```diff
@@ -116,3 +117,69 @@ def get(self, request, commit_hash: str | None) -> Response:
...
+        self,
+        request: HttpRequest,
+        commit_hash: str,
+    ) -> Response:
+        try:
+            params = TreeQueryParameters(
+                git_branch=request.GET.get("git_branch"),
+                git_url=request.GET.get("git_url"),
+                origin=request.GET.get("origin"),
+            )
```
danieldiasbarbo, line 174: Ditto about model validation

backend/kernelCI_app/views/treeDetailsBuildsView.py
```diff
@@ -116,3 +117,69 @@ def get(self, request, commit_hash: str | None) -> Response:
...
+        except ValidationError as e:
+            return create_api_error_response(error_message=e.json())
+
+        return super().get(
+            request=request,
+            git_url=params.git_url,
+            tree_name=None,
```
danieldiasbarbo, line 181: This can be removed after you setted the default value on the super method definition.

danieldiasbarbo (commented):
I've commented only on BuildsView but the comments applies also to the other two views.

backend/kernelCI_app/views/treeDetailsBuildsView.py
MarceloRobert: I wanted to leave them without defaults in order to "force" the validation in the children and make it explicit when one of the fields is always expected to be None or not

backend/kernelCI_app/typeModels/treeDetails.py
```diff
@@ -56,6 +56,12 @@ class DirectTreeQueryParameters(BaseModel):
     )
 
 
+class DirectTreePathParameters(BaseModel):
+    tree_name: str
+    git_branch: str
+    commit_hash: str
+
+
```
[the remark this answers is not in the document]
MarceloRobert: It is used in the tests in order to unify which fields are expected

PR #1389, opened 2025-08-12
feat(settings): add CORS_ALLOWED_ORIGINS as env var

nuclearcat:
Make CORS_ALLOWED_ORIGINS configurable via environment variable to allow for setting up instances of the dashboard for different domains.

CORS_ALLOW_ALL_ORIGINS is now also configurable via an environment variable, defaulting to 'False'.

Updated README.md with instructions on how to use the new env var.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,20 @@ If you are going to use a database user other than `kernelci`, set it to `DB_DEF
 export DB_DEFAULT_USER=<user>
 ```
 
+If you are setting up instance different than production KernelCI dashboard, you need to define CORS_ALLOWED_ORIGINS.
+
+docker-compose example:
+```yaml
+services:
+  backend:
+    environment:
+      CORS_ALLOWED_ORIGINS: ["https://d.kernelci.org","https://dashboard.kernelci.org"]
+```
+or .env file example:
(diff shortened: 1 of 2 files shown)


README.md
MarceloRobert, line 1: Environment variables will already be sent as strings, so you don't need the `'` in both examples, just use
`CORS_ALLOWED_ORIGINS=["https://d.kernelci.org","https://dashboard.kernelci.org"]` (without the `'` around the array)

Using the apostrophes will actually break the var parsing.

MarceloRobert (commented):
The notifications command is no longer working properly since the default value of `CORS_ALLOWED_ORIGINS` contains the production instance, which will make the command always return.

README.md
MarceloRobert: Though the new default is an empty array, you can keep those examples here, but remove the `'` around the array such that future users don't get confused

MarceloRobert (approved):
It's working well, but please change the example in the readme to remove just the apostrophes

nuclearcat:
I will make sure on correct variables in README, as seems .env file and docker-compose syntax are tricky a little, and then will merge.

PR #1395, opened 2025-08-13
Feat: save sent checkout reports and skip recently sent reports

### LINKED ISSUE #1216: Save checkouts notifications once per report

MarceloRobert:
## Description
Previously we were just saving when a notification was sent, its message id and checkout id. Those fields don't define a report, so we weren't able to discern if a specific report had been sent or not. This knowledge is useful in order to be able to not query for all reports all the time, allowing us to skip heavy reports and only re-send them after some time.

## Changes
- Added more fields to the NotificationCheckouts table in order to identify a unique report
  - Updates old entries with a specific `RunPython` migration piece
- Added check to see which reports where sent in the last 12 hours (possible to change with a global variable)
- Added ability to skip reports that were already sent in that interval, defaulted to true

## How to test
- Run the migration with `poetry run python3 manage.py migrate --database=notifications`
- Run the summary command with `poetry run python3 manage.py notifications --action summary`
- Mock the "mark as sent" by commenting the line with `if msg_id and email_args.update` and let some reports be marked as sent with whatever msg_id
- Run the summary command again, this time you'll see a log saying that some of the reports were skipped due to being already sent recently

Closes #1216

### DIFF
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -558,7 +558,12 @@ def run_checkout_summary(
 
             if msg_id and email_args.update:
                 mark_checkout_notification_as_sent(
-                    checkout_id=record["checkout_id"], msg_id=msg_id
+                    msg_id=msg_id,
+                    checkout_id=record["checkout_id"],
+                    git_repository_branch=branch,
+                    git_repository_url=giturl,
+                    origin=origin,
+                    path=path,
                 )
 
 
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -25,6 +25,8 @@ from kernelCI_app.management.commands.helpers.common import (
 from kernelCI_app.management.commands.helpers.summary import (
     SUMMARY_SIGNUP_FOLDER,
     PossibleReportOptions,
+    ReportConfigs,
+    TreeKey,
     process_submissions_files,
 )
 from kernelCI_app.queries.notifications import (
@@ -38,6 +40,8 @@ from kernelCI_app.queries.notifications import (
 )
 from kernelCI_app.queries.test import get_test_details_data, get_test_status_history
 from kernelCI_cache.queries.notifications import (
+    RESEND_INTERVAL,
+    check_sent_notifications,
     mark_checkout_notification_as_sent,
     mark_issue_notification_not_sent,
     mark_issue_notification_sent,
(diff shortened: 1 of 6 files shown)


backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -455,12 +458,39 @@ def process_submission_options(
     return recipients
 
 
+def discard_sent_reports(
+    tree_key_set: set[TreeKey],
+    tree_prop_map: dict[TreeKey, list[dict[str, Any]]],
```
giubrocchi, line 463: It seems like the `dict[TreeKey, list[dict[str, Any]]]` is used in more than one place, can it be extracted to a type?

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -455,12 +458,39 @@ def process_submission_options(
...
+    already_sent_report_checkouts = check_sent_notifications(prop_map_copy)
+    discarded = 0
+    for sent_checkout in already_sent_report_checkouts:
+        key_set_copy.discard(sent_checkout)
+        prop_map_copy.pop(sent_checkout, None)
+        discarded += 1
+    print(
+        "Discarded %d reports already sent in the past %s"
+        % (discarded, RESEND_INTERVAL)
+    )
```
giubrocchi, line 481: Can we replace for `len(already_sent_report_checkouts)`?

```suggestion
    for sent_checkout in already_sent_report_checkouts:
        key_set_copy.discard(sent_checkout)
        prop_map_copy.pop(sent_checkout, None)
    print(
        "Discarded %d reports already sent in the past %s"
        % (len(already_sent_report_checkouts), RESEND_INTERVAL)
    )
```

backend/kernelCI_cache/migrations/0012_add_checkout_report_identification_fields.py
```diff
@@ -0,0 +1,120 @@
...
+    notifications_connection = connections["notifications"]
+    notifications_query = """
+                SELECT
+                    id, checkout_id, git_repository_branch, git_repository_url, origin
+                FROM
+                    notifications_checkout
+                WHERE
+                    git_repository_branch = %(temp_default)s
+                    OR git_repository_url = %(temp_default)s
+            """
```
giubrocchi, line 22: nit: fix indentation

backend/kernelCI_cache/migrations/0012_add_checkout_report_identification_fields.py
```diff
@@ -0,0 +1,120 @@
...
+        checkouts_to_update = cursor.fetchall()
+
+    if not checkouts_to_update:
+        return
+
+    # Queries in kcidb for the missing data
+    if settings.USE_DASHBOARD_DB:
+        kcidb_connection = connections["kcidb"]
+    else:
+        kcidb_connection = connections["default"]
```
giubrocchi, line 36: nit:
```suggestion
    kcidb_connection = connections["kcidb"] if settings.USE_DASHBOARD_DB else connections["default"]
```

backend/kernelCI_cache/migrations/0012_add_checkout_report_identification_fields.py
```diff
@@ -0,0 +1,120 @@
...
+
+    checkout_ids = [checkout[1] for checkout in checkouts_to_update]
+    placeholders = ",".join(["%s"] * len(checkout_ids))
+    kcidb_query = f"""
```
giubrocchi, line 40: Ditto about indentation

backend/kernelCI_app/management/commands/notifications.py
MarceloRobert: I extracted the list part to make a meaning out of it, but I don't want to abstract things too much

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -748,6 +790,11 @@ def add_arguments(self, parser):
             help="Limit checkout summary to specific origins (comma-separated list)."
             + " If not provided, any origin will be considered",
         )
+        parser.add_argument(
+            "--skip-sent-reports",
+            action="store_true",
+            help="Skip reports that have already been sent",
+        )
```
padovan, line 797: Add a change to modify the management command invocation to include this option.

PR #1398, opened 2025-08-17
fix(dashboard): Install rsync in Docker image

nuclearcat:
The dashboard container was failing to start because the `rsync` command was not found in the alpine image. This change adds `rsync` to the list of installed packages in the Dockerfile to ensure the container runs correctly.

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -15,6 +15,8 @@ FROM alpine:3.20
 
 VOLUME ["/data/static/"]
 
+RUN apk add --no-cache rsync
+
 COPY --from=builder /dashboard/dist /files
 
 CMD rsync -azvi --checksum --exclude='*.map' /files/ /data/static/


nuclearcat:
> In which cases does this happen? I thought that the errors with the dashboard dockerfile were ironed out already

I was testing self-contained kcidb-ng install for full operation and noticed it during log review.

nuclearcat:
I didnt research deeply how it works, but i guess it builds once frontend and rsync it, this is what i saw on first successful run. 
On second run it didnt show anything, so it just quit (because frontend already in sync). Probably you saw same, because there was successful run before.

MarceloRobert:
Well, it does indeed solve an error so it looks good to me. I still think that there could be an issue describing the goal in which you are working towards with these PRs

MarceloRobert (approved):

MarceloRobert:
Ah right, I didn't see it since it's on another repo :sweat_smile:

PR #1407, opened 2025-08-20
chore: add coverage-v8

LucasSantos27:
Add new dependency to be able to use the coverage script:

`pnpm coverage`

image


MarceloRobert (approved):
Looks good, I didn't know about this test coverage. Even though we have a lot of backend tests, our frontend is lacking

giubrocchi (approved):

LucasSantos27:
> Looks good, I didn't know about this test coverage. Even though we have a lot of backend tests, our frontend is lacking

This test coverage just covers the FE tests. I've just added this dep, thinking a user wants to run this script

PR #1410, opened 2025-08-21
notifications: remove myself from Cc

padovan:
I was receiving all status email for validation. I think we can stop that now and I'll check on the mailing list frequently.


MarceloRobert (approved):
All right

PR #1412, opened 2025-08-21
fix: ensure filtered issue listing does not break

### LINKED ISSUE #1381: Issue listing breaks on 0 results
If an issue listing has 0 results it is returning error 500, meaning that something is breaking. Instead, it should return error 200 with the error message of "No issues found".

For example, trying to get issues with a filter for a nonexistent origin will yield error 500: https://staging.dashboard.kernelci.org:9000/issues?df|o|nonexistent=true

LucasSantos27:
This PR updates the behavior of the issue filtering endpoint.
Previously, when no issues matched the given filters, the API returned a 500 Internal Server Error.
Now, it correctly returns a 200 OK response with a descriptive message.

## Related Issue
- Closes #1381

## How to test it
**Using curl**
```bash
curl -X 'GET' \
  'http://localhost:8000/api/issue/?interval_in_days=120&filter_origin=nonexistent' \
  -H 'accept: application/json'
```

**Or via browser**
 `<local_url>/issues?df|o|nonexistent=true`

### Expected behavior

- **Before (main branch):**

Response: 500 Internal Server Error

- **After (this PR):**
Response: 200 OK

Body:
```json
{"error": "No issues found matching the filters"}
```


MarceloRobert (commented):
I think that it would be best if the endpoint returned the `IssueListingResponse` as usual, because it will have data there (even if 0 issues are returned)

LucasSantos27:
> I think that it would be best if the endpoint returned the IssueListingResponse as usual, because it will have data there (even if 0 issues are returned)

@MarceloRobert Yeah, you're right. It's better if I process the `extra_details` when the filtered issues aren't empty.

I've just updated it.

MarceloRobert (approved):
Looks good to me

PR #1435, opened 2025-08-25
Refactor: abstract cards responsiveness

### LINKED ISSUE #1411: Cards should span a single column on small screens
In some tabs, the cards for configs, architectures, and hardware are still covering two columns or having no size limit at all. This should be fixed so that the website becomes more responsive and easier to use on smaller screens.

Ex.:

Image

Image

MarceloRobert:
Refactors the way cards are used in tree/hardware details. Abstracts the use of `DesktopGrid` and `MobileGrid` with a high level component so that we don't need to keep repeating the definition of every card.

## Changes
- Added `ResponsiveDetailsGrid` component, abstracting the use of Desktop and Mobile Grids
- Refactored all tabs for tree and hardware details to use the Responsive Grid

## How to test
- Go to the tabs in treeDetails and hardwareDetails and check that they are all consistent and responsive
- This is a visual change only, the functionality should all be still the same

Closes #1411

### DIFF
--- a/dashboard/src/components/Tabs/TabGrid.tsx
+++ b/dashboard/src/components/Tabs/TabGrid.tsx
@@ -1,16 +1,50 @@
-import type { ReactNode, JSX } from 'react';
+import { type JSX, memo } from 'react';
 
-type GridProps = {
-  children: ReactNode;
+/**
+ * Top level grid for responsiveness in the tree/hardware details pages.
+ * The topCards will be shown at the top of both columns in large format and have priority in the middle format.
+ * The footerCards will be shown at the bottom and always spanning 2 columns.
+ */
+const ResponsiveDetailsGrid = ({
+  topCards,
+  bodyCards,
+  footerCards,
+}: {
+  topCards: JSX.Element[];
+  bodyCards: JSX.Element[];
+  footerCards: JSX.Element[];
+}): JSX.Element => {
+  const midIndex = Math.ceil(topCards.length / 2);
+  const leftTopCards = topCards.slice(0, midIndex);
+  const rightTopCards = topCards.slice(midIndex);
+
+  const midBodyIndex = Math.ceil(bodyCards.length / 2);
+  const leftBodyCards = bodyCards.slice(0, midBodyIndex);
+  const rightBodyCards = bodyCards.slice(midBodyIndex);
+
+  return (
+    <>
+      <div className="hidden grid-cols-2 gap-4 min-2xl:grid">
+        <div>
+          {leftTopCards}
+          {leftBodyCards}
+        </div>
+        <div>
+          {rightTopCards}
+          {rightBodyCards}
+        </div>
+        <div className="col-span-2">{footerCards}</div>
+      </div>
+      <div className="min-2xl:hidden">
+        {topCards}
+        <div className="grid grid-cols-1 lg:grid-cols-2 lg:gap-x-8">
+          <div>{leftBodyCards}</div>
+          <div>{rightBodyCards}</div>
+        </div>
+        {footerCards}
+      </div>
+    </>
+  );
 };
-export const MobileGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="min-[1652px]:hidden">{children}</div>
-);
 
-export const InnerMobileGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="grid grid-cols-1 lg:grid-cols-2 lg:gap-x-8">{children}</div>
-);
-
-export const DesktopGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="hidden grid-cols-2 gap-4 min-[1652px]:grid">{children}</div>
-);
+export const MemoizedResponsiveDetailsCards = memo(ResponsiveDetailsGrid);
--- a/dashboard/src/pages/hardwareDetails/Tabs/Boots/BootsTab.tsx
+++ b/dashboard/src/pages/hardwareDetails/Tabs/Boots/BootsTab.tsx
@@ -20,11 +20,7 @@ import {
   type PossibleTableFilters,
 } from '@/types/tree/TreeDetails';
 
-import {
-  DesktopGrid,
-  MobileGrid,
-  InnerMobileGrid,
-} from '@/components/Tabs/TabGrid';
+import { MemoizedResponsiveDetailsCards } from '@/components/Tabs/TabGrid';
 
 import { MemoizedStatusCard } from '@/components/Tabs/StatusCard';
 import MemoizedConfigList from '@/components/Tabs/Tests/ConfigsList';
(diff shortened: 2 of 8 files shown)


dashboard/src/components/Tabs/TabGrid.tsx
```diff
@@ -1,16 +1,67 @@
...
+const InnerMobileGrid = ({ children }: GridProps): JSX.Element => (
   <div className="grid grid-cols-1 lg:grid-cols-2 lg:gap-x-8">{children}</div>
 );
 
-export const DesktopGrid = ({ children }: GridProps): JSX.Element => (
+const DesktopGrid = ({ children }: GridProps): JSX.Element => (
   <div className="hidden grid-cols-2 gap-4 min-[1652px]:grid">{children}</div>
 );
+
+const ResponsiveDetailsGrid = ({
```
giubrocchi, line 19: Since we have this `ResponsiveDetailsGrid`, can we remove the logic where each grid gets hidden within a certain width range and instead do a conditional render depending on the width?
Something like this:

```
const width = someFunctionToGetWidthOfComponent();

return (
    width < 1652 ? <MobileGrid /> : <DesktopGrid />
)
```

This way the data isn't loaded two time on the dom

MarceloRobert: It is possible to do that, but I would have to be checking the screenSize with an useEffect and event listeners. It is generally better to use CSS instead of JS for this, specially since it is already dealing with the issue. I did change the display to be `none` instead of `hidden`, so that they won't appear on the inspector tools and similars

dashboard/src/components/Tabs/TabGrid.tsx
```diff
@@ -1,16 +1,50 @@
...
+  const midBodyIndex = Math.ceil(bodyCards.length / 2);
+  const leftBodyCards = bodyCards.slice(0, midBodyIndex);
+  const rightBodyCards = bodyCards.slice(midBodyIndex);
 
-export const InnerMobileGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="grid grid-cols-1 lg:grid-cols-2 lg:gap-x-8">{children}</div>
-);
+  return (
+    <>
+      <div className="[display:none] grid-cols-2 gap-4 min-[1652px]:grid">
```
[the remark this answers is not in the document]
MarceloRobert: Fixed

dashboard/src/components/Tabs/TabGrid.tsx
```diff
@@ -1,16 +1,50 @@
...
+          {leftTopCards}
+          {leftBodyCards}
+        </div>
+        <div>
+          {rightTopCards}
+          {rightBodyCards}
+        </div>
+        <div className="col-span-2">{footerCards}</div>
+      </div>
+      <div className="min-[1652px]:[display:none]">
```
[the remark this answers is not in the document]
MarceloRobert: ditto

dashboard/src/components/Tabs/TabGrid.tsx
```diff
@@ -1,16 +1,67 @@
...
+import { type ReactNode, type JSX, memo } from 'react';
 
 type GridProps = {
   children: ReactNode;
 };
-export const MobileGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="min-[1652px]:hidden">{children}</div>
+
+const MobileGrid = ({ children }: GridProps): JSX.Element => (
+  <div className="min-[1652px]:none">{children}</div>
```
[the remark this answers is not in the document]
MarceloRobert: Ahh now I see that `hidden` is tailwind's `display: none`. But tailwind's `invisible` is `visibility: hidden`... I'll change it

dashboard/src/pages/hardwareDetails/Tabs/Tests/TestsTab.tsx
```diff
@@ -132,75 +128,58 @@ const TestsTab = ({
...
-            configStatusCounts={testsSummary.configs}
-            diffFilter={diffFilter}
-          />
-        </div>
+  const { topCards, bodyCards, footerCards } = useMemo(() => {
+    return {
+      topCards: [
+        <MemoizedStatusCard
+          key="statusGraph"
+          title={<FormattedMessage id="testsTab.testStatus" />}
```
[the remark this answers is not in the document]
MarceloRobert: Thanks for pointing out a fix that I hadn't even noticed. Yes, the original code used bootStatus for the mobile, but it should be using testStatus. The new changes will fix this and make both desktop and mobile use testStatus, as it should be.

giubrocchi (approved):

dashboard/src/pages/hardwareDetails/Tabs/Boots/BootsTab.tsx
```diff
@@ -147,75 +143,58 @@ const BootsTab = ({
...
-            hardwareId={hardwareId}
-          />
-          <MemoizedConfigList
-            title={<FormattedMessage id="bootsTab.configs" />}
-            configStatusCounts={bootsSummary.configs}
-            diffFilter={diffFilter}
-          />
-        </div>
+  const { topCards, bodyCards, footerCards } = useMemo(() => {
+    return {
```
LucasSantos27, line 147: is there any reason to use `useMemo` besides the increase in performance? just question

dashboard/src/components/Tabs/TabGrid.tsx
```diff
@@ -1,16 +1,50 @@
...
+  const midBodyIndex = Math.ceil(bodyCards.length / 2);
+  const leftBodyCards = bodyCards.slice(0, midBodyIndex);
+  const rightBodyCards = bodyCards.slice(midBodyIndex);
 
-export const InnerMobileGrid = ({ children }: GridProps): JSX.Element => (
-  <div className="grid grid-cols-1 lg:grid-cols-2 lg:gap-x-8">{children}</div>
-);
+  return (
+    <>
+      <div className="hidden grid-cols-2 gap-4 min-[1652px]:grid">
```
LucasSantos27, line 27: tailwind has `2xl` as 1536px. Why not use instead of "hardcoding" 1652px?

LucasSantos27 (approved):
Good job!

dashboard/src/components/Tabs/TabGrid.tsx
MarceloRobert: The value was there before my changes so I assume it was a design choice

LucasSantos27: In other sections, we've used `max-2xl`, so we can keep this pattern.

`TestTab / BuildTab / BootsTab`

LucasSantos27 (commented):
It would be better to create one commit because the first is returning a ci error

PR #1439, opened 2025-08-26
Chore: update brace expansion

MarceloRobert:
When updating brace expansion, I also updated the packages that depend on it. The thing is that when updating the packages that depend on it, they got a bit more clever (especially tanstack-router) and pointed out some errors that weren't being shown before, so I had to fix those errors as well.

## Changes
- Removed an unused api function that was also causing a lint error
- Fixed some mismatch type returns in the hardware api
- Fixed one function return and one type import
- Changed some `navigate` issues where the path params were required but they weren't being passed.
  Whenever we are navigating to a route with required path params (such as '/tree/{treeId}') we must pass the parameters. There were cases where the 'navigate' was going from one of those routes, but receiving the route via a variable and so the build step wasn't complaining (but with the package update it started to).
- Updated brace-expansion and its dependants

## How to test
- Install the updated packages
- Check that `pnpm build` won't break
- Check to see if the entire dashboard is still working correctly

Closes Dependabot issues [43](https://github.com/kernelci/dashboard/security/dependabot/43) and [44](https://github.com/kernelci/dashboard/security/dependabot/44)


giubrocchi (approved):

dashboard/package.json
```diff
@@ -19,36 +19,36 @@
     "prettify": "prettier --write ./src"
   },
   "dependencies": {
-    "@date-fns/tz": "^1.2.0",
+    "@date-fns/tz": "^1.4.1",
```
LucasSantos27, line 22: where `@date-fns/tz` is being used? I didn't find

dashboard/package.json
```diff
@@ -19,36 +19,36 @@
...
     "@emotion/react": "^11.14.0",
-    "@emotion/styled": "^11.14.0",
+    "@emotion/styled": "^11.14.1",
     "@hookform/resolvers": "^3.10.0",
-    "@microlink/react-json-view": "^1.26.1",
-    "@mui/material": "^6.4.11",
+    "@microlink/react-json-view": "^1.27.0",
+    "@mui/material": "^6.5.0",
     "@mui/x-charts": "^7.29.1",
```
LucasSantos27, line 28: could we update [@mui/x-charts](https://www.npmjs.com/package/@mui/x-charts?activeTab=versions) to 8.11.0

dashboard/package.json
```diff
@@ -19,36 +19,36 @@
...
-    "@date-fns/tz": "^1.2.0",
+    "@date-fns/tz": "^1.4.1",
     "@emotion/react": "^11.14.0",
-    "@emotion/styled": "^11.14.0",
+    "@emotion/styled": "^11.14.1",
     "@hookform/resolvers": "^3.10.0",
-    "@microlink/react-json-view": "^1.26.1",
-    "@mui/material": "^6.4.11",
+    "@microlink/react-json-view": "^1.27.0",
+    "@mui/material": "^6.5.0",
```
LucasSantos27, line 27: and this to 7.3.2

PR #1450, opened 2025-08-29
feat: ensure tree report includes build issues

### LINKED ISSUE #1449: Update tree-report endpoint to include new build issues data
Extend the tree-report API endpoint to return the new build issues information.

LucasSantos27:
- Update tree report response to use `Annotated`
- Include build issues in the tree report

## Related issues

- Closes #1449 

## Visual Reference

image

### DIFF
--- a/backend/kernelCI_app/constants/localization.py
+++ b/backend/kernelCI_app/constants/localization.py
@@ -152,6 +152,11 @@ class DocStrings:
     TREE_REPORT_UNSTABLE_TESTS_DESCRIPTION = (
         "History of tests that are unstable. " + REGRESSIONS_GROUP
     )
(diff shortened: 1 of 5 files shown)


MarceloRobert (commented):
In order to pave the ground for possible future changes, I think it would be better to use a different structure, instead of
```
{
  ...
  checkout_build_issues: []
}
```
use
```
{
  ...
  issues: {
    builds: []
  }
}
```
just in case we want to add test issues in the future, and this is also closer to the structure in other endpoints.

MarceloRobert:
Also also, it seems ok for me but would @padovan want to split the new issues from the existing issues? This way you could also get rid of the `is_new_issue` field.

PR #1456, opened 2025-09-01
Feat: add fixed date to issue listing query

### LINKED ISSUE #1455: bug(ci): empty issue list breaks ci
Since we are using production data for our integration tests, when there are unexpected scenarios such as empty issue listing we end up with a broken CI

expected: CI runs correctly even if there aren't issues in the last 5 days (default number)

current behavior: CI breaks because the response is not formatted as expected

possible solutions: increase the default interval (would make the requests slower); increase the interval incrementally only if the request fails (would increase the number of requests but would make it work anytime); change the endpoint response to make it work with the CI even with 0 results (would require a small refactor of issue list frontend in order to keep current behavior but would make the response different than other endpoints)

MarceloRobert:
Adding a fixed starting point to the issue listing query allows for the returned data to be consistent for the CI tests, avoiding errors related to missing data.

## Changes
- Adds a new parameter to the issue listing query and endpoint;
- Uses that parameter in the issue listing test;
- Changes the default interval in days to 3 days in the issue listing integration tests.

## How to test
- Check the issue listing and see that it is still working correctly and with the same data as staging
- Make requests to the issue listing endpoint passing different starting points

The issue listing page should be working just like before; The CI should be fixed.

Closes #1455

### DIFF
--- a/backend/kernelCI_app/queries/issues.py
+++ b/backend/kernelCI_app/queries/issues.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from typing import Any, Optional
 from django.db import connection, connections
 from kernelCI_app.cache import get_query_cache, set_query_cache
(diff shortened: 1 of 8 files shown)


backend/kernelCI_app/queries/issues.py
```diff
@@ -95,12 +96,20 @@ def get_issue_tests(*, issue_id: str, version: Optional[int]) -> list[dict]:
 def get_issue_listing_data(
     *,
     interval: str,
+    starting_date_iso_format: Optional[str] = None,
 ) -> list[dict]:
     """Queries the list of all issues from `interval_date` parameter to now.
```
LucasSantos27, line 101: to now or `starting_date`...

LucasSantos27 (approved):

PR #1458, opened 2025-09-02
fix: fix detail page overflow title Fixes (#1434)

### LINKED ISSUE #1434: Long titles on details page don't wrap
Sometimes long titles in build/test/issues details pages don't wrap and clip through the page

Image

khanheric2003:
## Description
- Fix the detail page title overflow by making it inside the gray box

## Related Issue
- Fixes #1434 

## Changes Made
- Updated tailwindcssCSS tag of Section object in src/components/Section/Section.tsx (line 129-132)

## Screenshots
image


dashboard/src/components/Section/Section.tsx
```diff
@@ -126,9 +126,13 @@ const Section = ({
     <div className="text-dim-gray mb-6 flex flex-col gap-4">
       <div className="flex flex-col gap-2">
         {eyebrow && <span className="text-sm">{eyebrow}</span>}
-        <div className="flex flex-row items-center gap-2">
+        <div className="flex flex-row flex-wrap items-center gap-2 overflow-hidden">
```
MarceloRobert, line 129: > `<div className="flex flex-row flex-wrap items-center gap-2 overflow-hidden">`

This is not needed, and the `flex-wrap` actually moves the status icon up independently of screen-width, which isn't good

dashboard/src/components/Section/Section.tsx
```diff
@@ -126,9 +126,13 @@ const Section = ({
...
           {leftIcon}
-          {title && <span className="text-2xl font-bold">{title}</span>}
+          {title && (
+            <span className="max-w-full text-2xl font-bold break-words">
```
MarceloRobert, line 132: > `<span className="max-w-full text-2xl font-bold break-words">`

IMO a `break-all` would be sufficient here. Even though it will break on any character, it makes the layout better than `break-words`

MarceloRobert (commented):
Please squash your commits into one, no need to add something and then revert it in the next commit.

dashboard/src/components/Section/Section.tsx
LucasSantos27: and with `break-all`, `overflow-hidden` isn't also needed

dashboard/src/components/Section/Section.tsx
LucasSantos27: Yep, I would follow with `break-all` to keep the same behavior of `Config` in `Build Info`

khanheric2003: sure thing let me make a change to it

MarceloRobert (approved):

Issue #1123, opened 2025-03-26
Bug: placeholderdata keeps oldest non-null data


MarceloRobert:
Placeholder data from treeDetails is storing the last time there was data, so if you change from commit 1 with data, go to commit 2 without data and then to a commit 3 without data as well, the transition from commit 2 to 3 will show the data from the first commit.

This is a visual bug, doesn't affect performance nor breaks the dashboard.

Issue found on #1117

PR #1462, opened 2025-09-03
feat: ensure all hardware details trees can be deslected

### LINKED ISSUE #614: HardwareDetails: create a way to easily select just one Tree instead of deselecting all but one
Sometimes we want to see the hardware details of just one tree, currently we have to uncheck all but one in this case, which is very difficult to do when we have 27 trees for example.
Image

LucasSantos27:
Users must deselect all hardware details trees at once.

When no trees are selected, a message is displayed to inform the user that no tree has been selected.

## Related Issue
- Closes #614 

## Visual Reference
https://github.com/user-attachments/assets/04871913-37f1-4fa7-a1ab-f93ccb3c28e6

## How to test it

Go to hardware section and choose once to see the details, deselect all trees

### DIFF
--- a/dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
+++ b/dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
@@ -124,7 +124,7 @@ function HardwareDetails(): JSX.Element {
 
   const reqFilter = mapFilterToReq(diffFilter);
 
-  const updateTreeFilters = useCallback((selectedIndexes: number[]) => {
+  const updateTreeFilters = useCallback((selectedIndexes: number[] | null) => {
     navigate({
       search: previousSearch => ({
         ...previousSearch,
@@ -174,7 +174,7 @@ function HardwareDetails(): JSX.Element {
       endTimestampInSeconds: endTimestampInSeconds,
       origin: origin,
       filter: reqFilter,
-      selectedIndexes: treeIndexes ?? [],
+      selectedIndexes: treeIndexes,
       treeCommits: treeCommits,
       treeIndexesLength: treeIndexesLength,
     });
@@ -347,6 +347,8 @@ function HardwareDetails(): JSX.Element {
     ],
   );
 
+  const hasSelectedTrees = treeIndexes === null || treeIndexes.length > 0;
+
   const hardwareTitle = useMemo(() => {
     return formatMessage(
       { id: 'title.hardwareDetails' },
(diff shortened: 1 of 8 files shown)


giubrocchi:
When no trees are selected and I apply a filter, it automatically selects all the trees. Not sure if it's intended, but I believe we can hide the filter since there is no data to show

https://github.com/user-attachments/assets/f9c99ea4-732d-4910-b0b4-5e270171fbce

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -334,29 +329,21 @@ const getColumns = (
...
-const indexesFromRowSelection = (
-  rowSelection: RowSelectionState,
-  maxTreeItems: number,
-): number[] => {
+const indexesFromRowSelection = (rowSelection: RowSelectionState): number[] => {
   const rowSelectionValues = Object.values(rowSelection);
-  if (
-    rowSelectionValues.length === maxTreeItems ||
-    rowSelectionValues.length === 0
-  ) {
```
MarceloRobert, line 357: If following the old behavior of selecting all trees by default, then this should come back. We should also check if the backend has the correct behavior when the indexes are empty

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
```diff
@@ -347,6 +347,10 @@ function HardwareDetails(): JSX.Element {
     ],
   );
 
+  const hasSelectedTrees = useMemo(() => {
+    return treeIndexes?.length < 0;
+  }, [treeIndexes]);
```
jnaconeski, line 352: I think, as this is a boolean value, it doesn't need to be memoized.
Also, I would like to confirm if the logic is correct. Shouldn't it check if it is greater than zero instead (`treeIndexes?.length < 0`)?

LucasSantos27: `treeIndexes?.length > 0`. I am not able to test because the db is not working and I was trying to change it to see the behaviors :smile:

LucasSantos27: Yep, `useMemo` isn't necessary

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -334,29 +329,21 @@ const getColumns = (
 
 const getInitialRowSelection = (
   selectedIndexes: number[],
-  treeItemsLength: number,
 ): Record<string, boolean> => {
   if (selectedIndexes.length === 0) {
-    return Object.fromEntries(
-      Array.from({ length: treeItemsLength }, (_, i) => [i.toString(), true]),
-    );
```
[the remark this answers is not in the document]
LucasSantos27: good catch

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
MarceloRobert: If a hardware has many trees, all of their indexes are shown in the url. If all trees are selected by default (and they are), they might clutter the url and we don't want that. That's why it should be hiding them on all indexes, otherwise the default page will have a long url.

MarceloRobert:
It seems that this change also suffers from #1123.
Example: enter a hardware details page with every tree selected, deselect all, select one tree.
This will show the data of all trees before loading the data for the single tree. It should be addressed, but on another PR.

MarceloRobert (approved):
LGTM, seems to be working well

PR #1469, opened 2025-09-04
fix: ensure search bar does not freeze 

### LINKED ISSUE #1466: Regression: Web app crashes when searching for trees, issues, or hardware
There is a regression when using the search bar in [staging](https://staging.dashboard.kernelci.org:9000/).

If the user searches for trees, issues, or hardware, the web application crashes, and it becomes impossible to navigate between pages from that point forward.

Image

### LINKED ISSUE #1470: Nested <a> tags in status table cause hydration error
When rendering the status table, the component is generating nested <a> tags. This is invalid HTML and causes hydration errors in the frontend.

Error logs:
```
Status.tsx:98 In HTML, <a> cannot be a descendant of <a>.
This will cause a hydration error.

table.tsx:117 <a> cannot contain a nested <a>.
See this log for the ancestor stack trace.
```

The issue occurs in both hardware and tree views.
It occurs when the status cells (e.g., boot, build, test) already contain links and are wrapped with `TableCellWithLink`, resulting in double anchor tags.

LucasSantos27:
### First commit: Search bar freeze

- Updated checkout delay to 0.5s.
- Added a new condition to verify the last searched value.
- This ensures that the search bar no longer freezes when searching for trees, issues, or hardware.

### Second commit: Nested links in table cells
- Removed TableCellWithLink for cases where the content already had a link (e.g., boot, build, and test statuses).
- The issue was happening in both hardware and tree views.
- This avoids nested <a> tags that caused hydration errors:

```
table.tsx:117 <a> cannot contain a nested <a>
Status.tsx:98 In HTML, <a> cannot be a descendant of <a>
```

## Related Issue

- Closes #1466
- Closes #1470

### DIFF
--- a/dashboard/src/components/DebounceInput/DebounceInput.tsx
+++ b/dashboard/src/components/DebounceInput/DebounceInput.tsx
@@ -1,16 +1,16 @@
 import type { ChangeEvent, JSX } from 'react';
-import { useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { useDebounce } from '@/hooks/useDebounce';
 import type { InputProps } from '@/components/ui/input';
 import { Input } from '@/components/ui/input';
 
 interface IDebounceInput extends Omit<InputProps, 'onChange' | 'value'> {
-  debouncedSideEffect: (value: React.ChangeEvent<HTMLInputElement>) => void;
+  debouncedSideEffect: (value: ChangeEvent<HTMLInputElement>) => void;
   startingValue?: string;
 }
 
-const DEBOUNCE_INTERVAL = 666;
+const DEBOUNCE_INTERVAL = 500;
 
 const DebounceInput = ({
   debouncedSideEffect,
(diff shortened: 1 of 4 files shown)


dashboard/src/components/Table/ConditionalTableCell.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    return gitCommitValueSelector({
+      commitHash: rowData.git_commit_hash,
+      commitName: rowData.git_commit_name,
+      commitTags: rowData.git_commit_tags,
+    }).content;
+  }, [columnType, cell.row.original]);
+
+  switch (columnType) {
```
jnaconeski, line 82: Nit: What do you think about using early returns here instead of a switch statement?
Also, since some text is repeated multiple times in this component, you might consider extracting constants. For example:

```
const StatusColumn = 'status';
const CommitTagsColumn = 'git_commit_tags';
```

```
  if (columnType === StatusColumn) {
    return ...;
  }

  if (columnType === CommitTagsColumn) {
    if (gitCommitValue) {
      ...
      return (
        ...
      );
    }
    return ...;
  }

  return (...);
```

LucasSantos27: sounds good, thank you!

dashboard/src/components/DebounceInput/DebounceInput.tsx
```diff
@@ -20,25 +20,38 @@ const DebounceInput = ({
   const [inputValue, setInputValue] = useState<string>(startingValue);
   const [inputEvent, setInputEvent] = useState<ChangeEvent<HTMLInputElement>>();
   const isThisComponentMounted = useRef(true);
+  const lastProcessedValue = useRef<string>(startingValue);
   const debouncedInputEvent = useDebounce(inputEvent, DEBOUNCE_INTERVAL);
 
-  const onInputSearchTextChange = (e: ChangeEvent<HTMLInputElement>): void => {
-    setInputValue(e.target.value);
-    setInputEvent(e);
-  };
```
MarceloRobert, line 28: This not being on a `useCallback` might have been the problem, I assume

dashboard/src/components/Table/ConditionalTableCell.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+interface ConditionalTableCellProps<T = TreeTableBody | HardwareItem> {
+  cell: Cell<T, unknown>;
+  linkProps: LinkProps;
+  linkClassName?: string;
+}
+
+const StatusColumn = 'status';
+const CommitTagsColumn = 'git_commit_tags';
+const RegularColumn = 'regular';
```
MarceloRobert, line 21: I disagree that these should be here. You're mixing the meaning of "column type" with a data field.
For example in `isTreeTableBody` where you compare `CommitTagsColumn in data`, it doesn't make sense for me to ask if a column type is inside the data.

And for `StatusColumn` and `RegularColumn`, their value doesn't seem to matter much as you are not manipulating them.
Instead of variables, I'd prefer if you used a `type` approach directly, with `type ColumnTypes = 'status' | 'commit_tag' | 'regular';` for example, where typescript would provide type safety in the returns and comparisons.

dashboard/src/components/Table/ConditionalTableCell.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }).content;
+  }, [columnType, cell.row.original]);
+
+  switch (columnType) {
+    case StatusColumn:
+      return (
+        <TableCell key={cell.id} className="p-4">
+          {cellContent}
+        </TableCell>
+      );
```
MarceloRobert, line 87: I liked that I could click outside of the status circles and still go directly to the boots or tests tab... but yeah, I don't know how to solve the "link inside of a link" problem right now so I guess this works

LucasSantos27: I consider this new behavior is ok for now

PR #1472, opened 2025-09-05
fix: use regex to avoid issues when branch name contains slashes

### LINKED ISSUE #1468: Bug: trees with `/` on the branch aren't parsed correctly in the url
Trees such as soc / arm/fixes have a `/` in the branch, and when you try to access the tree details page for example on `https://dashboard.kernelci.org/tree/soc/arm%2Ffixes/f2e8f1f5f5c81037fc0b2d78eb75aeb80a0ee507`, it doesn't make the requests correctly

LucasSantos27:
Replace Django path() with re_path() using regex patterns to properly handle git branch names containing forward slashes (e.g., 'arm/fixes').

```
https://dashboard.kernelci.org/tree/mainline/master/c8ed9b5c02a5ceb3d8244f3862a7e64cf0b5648e
```

The regex patterns use **non-greedy matching** and specific commit hash validation to ensure correct URL routing without conflicts.

### How regex works
- `(?P<tree_name>[^/]+)` - Captures tree name (no slashes allowed)
- `(?P<git_branch>.+?)` - Captures git branch with **non-greedy matching** (allows slashes)
- `(?P<commit_hash>[a-f0-9]{40})` - Captures exactly 40-char hex commit hash

The non-greedy `(.+?)` pattern ensures the git_branch capture stops at the first valid commit hash. For this, we ensured the commit hash must have 40 characters.

## Related Issue
- closes #1468


tales-aparecida:
I find your lack of unittests disturbing

image


But on a serious note, I was expecting this to be more about `urllib.parse.unquote` than meddling with the routes.

Have you tested it locally with something like https://staging-dashboard.kernelci.org/tree/soc/arm%2Ffixes/f2e8f1f5f5c81037fc0b2d78eb75aeb80a0ee507?

tales-aparecida:
I guess I found what Marcelo was saying about expeting django to unquote it automatically

https://docs.djangoproject.com/en/5.2/topics/http/urls/?utm_source=chatgpt.com#path-converters

Now I'm more convinced of your solution, but... could we just use `path:` instead? Or that wouldn't work because it would try to match the whole URL?

LucasSantos27:
@tales-aparecida  I tested with this **git_branch** :smile: 

image

LucasSantos27:
> Now I'm more convinced of your solution, but... could we just use path: instead? Or that wouldn't work because it would try to match the whole URL?

@tales-aparecida path() won't work for this use case because `<str:parameter>` converter explicitly excludes forward slashes

/tree/soc/arm%2Ffixes/f2e8f1f5f5c81037fc0b2d78eb75aeb80a0ee507

In this case, for example:

- tree_name = soc
- git_branch = arm
- commit_hash = fixes
- ??? = f2e8f1f5f5c81037fc0b2d78eb75aeb80a0ee507

In other words, path is not compatible with strings that contain `/`

tales-aparecida:
I mean `<path:tree_name>`

```
path(
        "tree/<path:tree_name>/<str:git_branch>",
        view_cache(views.TreeLatestCheckout),
        name="treeLatest",
    ),
```

LucasSantos27:
> I find your lack of unittests disturbing

@tales-aparecida Next week, I’ll start adding some unit tests for this. There’s definitely room for improvement in our test coverage. I’ve already added the badge you mentioned, and as you can see, the coverage is still quite low. 

https://github.com/kernelci/dashboard/blob/main/backend/README.md

PR #1477, opened 2025-09-10
django commands: Fix to use correct database

nuclearcat:


MarceloRobert:
As we discussed in the demo and saw on Discord ([thread link](https://discord.com/channels/1245820301053530313/1414602622043947059)), the use of the database is not the problem.The issue we've seen happened because of a problem on environment variables.

PR #1480, opened 2025-09-10
Feat: add test series field

### LINKED ISSUE #1475: Create trigger or another approach for series in tests
For the tests table, the `series` field depends on information from both the builds and tests tables. One possible solution is to use a trigger, but more research is needed to confirm whether this is the best approach.

Once defined, an index should also be created on this field to enhance query performance.

## Related Issue

- #1454

MarceloRobert:
Adds a new field to the tests model called "series" which will be a junction of path + platform.
For testing, also adds a new endpoint called "TestSeries" that retrieves the last 10 tests with that desired path and platform using the series field.

While I was checking the migrations, I also saw that the cache tables are being created on the postgres database, so I also fixed the database router while I was at it.

## Changes
- Added `series` field to tests model
- Added TestSeries endpoint for testing
- Fixed database router by avoiding cache tables on default database

## How to test
- Run the local database
- Apply the migration
- Check the database for the series field, or check the endpoint as well.

Closes #1475


LucasSantos27 (commented):
Could you please rebase with my [PR](https://github.com/kernelci/dashboard/pull/1479) to get the build info for our TestSeriesView?

PR #1489, opened 2025-09-15
ingester: Port batch ingester to django

nuclearcat:
Also added some verbose logs to monitor ingestion rate, queue, etc.

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -13,9 +13,11 @@ import traceback
 from typing import Any, Literal, Optional
 import yaml
 import kcidb_io
+from django.db import transaction
+from kernelCI_app.models import Issues, Checkouts, Builds, Tests, Incidents
 
 from kernelCI_app.management.commands.helpers.process_submissions import (
-    insert_submission_data,
+    build_instances_from_submission,
 )
 
 VERBOSE = 0
@@ -35,9 +37,41 @@ TREES_FILE = "/app/trees.yaml"
 
 logger = logging.getLogger("ingester")
 
-# Thread-safe queue for database operations
-db_queue = Queue()
+# Batching and backpressure controls
+try:
+    INGEST_BATCH_SIZE = int(os.environ.get("INGEST_BATCH_SIZE", "10000"))
+except (ValueError, TypeError):
+    logger.warning("Invalid INGEST_BATCH_SIZE, using default 10000")
+    INGEST_BATCH_SIZE = 10000
+
+try:
+    INGEST_FLUSH_TIMEOUT_SEC = float(os.environ.get("INGEST_FLUSH_TIMEOUT_SEC", "2.0"))
+except (ValueError, TypeError):
+    logger.warning("Invalid INGEST_FLUSH_TIMEOUT_SEC, using default 2.0")
+    INGEST_FLUSH_TIMEOUT_SEC = 5.0
+
+try:
+    INGEST_QUEUE_MAXSIZE = int(os.environ.get("INGEST_QUEUE_MAXSIZE", "5000"))
+except (ValueError, TypeError):
+    logger.warning("Invalid INGEST_QUEUE_MAXSIZE, using default 5000")
+    INGEST_QUEUE_MAXSIZE = 5000
+
+# Thread-safe queue for database operations (bounded for backpressure)
+db_queue = Queue(maxsize=INGEST_QUEUE_MAXSIZE)
 db_lock = threading.Lock()
+
+
+def _ts() -> str:
+    return time.strftime("%Y-%m-%d %H:%M:%S", time.localtime())
+
+
+def _out(msg: str) -> None:
+    """Write debug/perf output to stdout"""
+    # logger was unreliable in some environments
+    try:
+        print(f"[{_ts()}] {msg}", flush=True)
+    except Exception:
+        pass
 
 
 def move_file_to_failed_dir(filename, failed_dir):
@@ -259,16 +293,7 @@ def prepare_file_data(filename, trees_name, spool_dir):
         }
 
 
-def process_submission_item(data: dict[str, Any], metadata: dict[str, Any]):
-    with db_lock:
-        insert_submission_data(data, metadata)
-
-    if VERBOSE and "processing_time" in metadata:
-        ing_speed = metadata["fsize"] / metadata["processing_time"] / 1024
-        logger.info("Ingested %s in %.2f KB/s", metadata["filename"], ing_speed)
-
-
-def db_worker(stop_event: threading.Event):
+def db_worker(stop_event: threading.Event):  # noqa: C901
     """
     Worker thread that processes the database queue.
     This is the only thread that interacts with the database.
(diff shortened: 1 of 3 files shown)


backend/kernelCI_app/management/commands/monitor_submissions.py
```diff
@@ -70,6 +71,19 @@ def handle(
         try:
             while True:
                 # TODO: retry failed files every x cycles
+                try:
+                    json_files = [
+                        f
+                        for f in os.listdir(spool_dir)
+                        if os.path.isfile(os.path.join(spool_dir, f))
+                        and f.endswith(".json")
+                    ]
```
tales-aparecida, line 80: ChatGPT suggested scandir, which is a bit more cleaner, and https://stackoverflow.com/questions/59268696/why-is-os-scandir-as-slow-as-os-listdir seems to suggest it would be as fast as listdir.

```python
with os.scandir(spool_dir) as it:
    json_files = [
        entry.name
        for entry in it
        if entry.is_file() and entry.name.endswith(".json")
    ]
```

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -35,11 +38,40 @@
 
 logger = logging.getLogger("ingester")
 
-# Thread-safe queue for database operations
-db_queue = Queue()
+# Batching and backpressure controls
+try:
+    INGEST_BATCH_SIZE = int(os.environ.get("INGEST_BATCH_SIZE", "1000"))
+except Exception:
```
tales-aparecida, line 44: nit: prefer being more specific (applies to lines below)

```suggestion
except (ValueError, TypeError):
```

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -35,11 +38,40 @@
...
+    INGEST_BATCH_SIZE = 10000
+
+try:
+    INGEST_FLUSH_TIMEOUT_SEC = float(os.environ.get("INGEST_FLUSH_TIMEOUT_SEC", "2.0"))
```
tales-aparecida, line 48: It might be a nice idea to move these into `settings.py`, to rely on proper `from django.conf import settings` handling, even if we create a `ingester_settings.py` that is included into the main settings file... but I can see this as a future refinement.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -268,7 +301,9 @@ def process_submission_item(data: dict[str, Any], metadata: dict[str, Any]):
         logger.info("Ingested %s in %.2f KB/s", metadata["filename"], ing_speed)
 
 
-def db_worker(stop_event: threading.Event):
+def db_worker(  # noqa: C901
+    stop_event: threading.Event,
+):
```
tales-aparecida, line 306: nit, I'm almost sure you could've put the noqa in front.

```suggestion
def db_worker(stop_event: threading.Event):  # noqa: C901
```

But I'd argue that the c901 alert was accurate, and we should try to refactor this function in the future

tales-aparecida:
Don't worry too much about the nitpicking, it's just how I walk through the patch while I understand the work.

I appreciate the batching and the need to call to `bulk_create`. There are a few "[caveats](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#bulk-create)" to using it, but I believe we can find alternatives whenever they become necessary.

I think the code could be drier... but I also think the functionality was already thoroughly tested manually. So we could merge this critical piece, write unit tests and performance tests over it, and only then think about refining the code.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -277,33 +312,167 @@ def db_worker(stop_event: threading.Event):
...
+                    inst = build_instances_from_submission(data)
+                    issues_buf.extend(inst["issues"])
+                    checkouts_buf.extend(inst["checkouts"])
+                    builds_buf.extend(inst["builds"])
+                    tests_buf.extend(inst["tests"])
+                    incidents_buf.extend(inst["incidents"])
+
+                # Flush if batch size reached
+                if buffered_total() >= INGEST_BATCH_SIZE:
+                    flush_buffers()
```
MarceloRobert, line 435: nit: no need for comments like this when the code is clear
```suggestion
                if buffered_total() >= INGEST_BATCH_SIZE:
                    flush_buffers()
```

backend/kernelCI_app/management/commands/monitor_submissions.py
```diff
@@ -70,6 +71,19 @@ def handle(
...
+                    ts = time.strftime("%Y-%m-%d %H:%M:%S", time.localtime())
+                    self.stdout.write(
+                        f"[{ts}] Spool has {len(json_files)} .json files pending"
+                    )
+                except Exception:
+                    pass
```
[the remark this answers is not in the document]
nuclearcat: In normal situation this exception should not appear.
During disk space issues it might create corrupted(incomplete) or 0 bytes file, it will result in queue growth on monitoring (it just counts json files), so by this parameter i will get monitoring alert.
Sure such should not crash ingestion, but also logging/printing anything will be worse, it will flood the logs. Moving file in failed will hide issue.
I might add in future some metrics here and send them somewhere.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
nuclearcat: Agree

backend/kernelCI_app/management/commands/helpers/process_submissions.py
```diff
@@ -130,6 +130,106 @@ def save_incident(incident) -> None:
     return
 
 
+# New helpers: build unsaved model instances for batch inserts
```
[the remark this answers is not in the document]
nuclearcat: Yes, i am leaving code, as sometimes i am enabling legacy path to test, and sometimes looking as reference.
On next iteration likely legacy will be removed.

MarceloRobert (changes requested):
I don't see why add clutter to the code when it can simply be refactored without much effort. I understand the reasoning, but this ingester is not even being currently used, we don't need to keep "legacy" code.

nuclearcat:
@MarceloRobert i think i removed legacy code

MarceloRobert (approved):
There could still be some improvements, but looks good enough to me.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -388,25 +568,82 @@ def ingest_submissions_parallel(
...
+        )
         # Wait for all database operations to complete
         db_queue.join()
 
+    except KeyboardInterrupt:
+        _out("[%s] KeyboardInterrupt: stopping ingestion and flushing..." % _ts())
+        try:
+            # Attempt to cancel remaining futures and exit early
+            # Note: this only cancels tasks not yet started
+            pass
```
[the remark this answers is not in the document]
nuclearcat: More graceful shutdown if process is interrupted. I decided it is not critical for now, as in general this ingester better to not interrupt, especially when it is in this part of code.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -277,33 +302,166 @@ def db_worker(stop_event: threading.Event):
...
+            + len(tests_buf)
+            + len(incidents_buf)
+        )
+
+    def flush_buffers():
+        nonlocal last_flush_ts
+        total = buffered_total()
+        if total == 0:
+            return
+        batch_size = INGEST_BATCH_SIZE
```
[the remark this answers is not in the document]
nuclearcat: Thanks for spotting, fixed. Leftover from previous approach (i used batch_size as counter that was decrementing during each batch add).

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -35,11 +37,43 @@
...
+    return time.strftime("%Y-%m-%d %H:%M:%S", time.localtime())
+
+
+def _out(msg: str) -> None:
+    """Write debug/perf output to stdout"""
+    # logger was unreliable in some environments
+    try:
+        print(msg, flush=True)
+    except Exception:
+        pass
```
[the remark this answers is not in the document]
nuclearcat: done

PR #1508, opened 2025-09-17
Refactor: unify tree latest

### LINKED ISSUE #1476: Unify treeLatest and treeLatestHash endpoints
Currently, these two endpoints are separate, but they could actually be unified: instead of using `commit_hash` as a path parameter, we could treat it as an optional query parameter.

By doing so, both would share the same path, and we would avoid problems in these cases.

## Related Issue
- #1468

MarceloRobert:
Since the commit hash was a parameter that could be or not be sent in order to get the endpoint data, it was turned into an optional query parameter instead of being a separate endpoint.

In the frontend, the routes for `/checkout/$treeName/$branch/$hash` and `/c/$treeName/$branch/$hash` still exists in order to keep backwards compatibility. _IF_ someone tries to access the url for a tree such as `soc arm/fixes`, they can avoid parameter conflict by encoding the `/`, for example `http://localhost:5173/checkout/soc/arm%2Ffixes/0c952efa0d7cf5258879406077df33d23cb06d5e`.

In the backend, the branch was altered to a `path` type, and the parameters are now returned by the endpoint if it doesn't encounter a tree in order to be clearer on how the variables were parsed.

## Changes
- Renamed treeLatest (view) to treeLatestView
- Unified treeLatest and treeLatestHash
- Updated types, documentation, integration tests, schema, example request and frontend api call

## How to test
Access the /tree/tree_name/branch endpoint with different arguments for commit_hash. Also check the frontend routes for `/tree/$treeName/$branch`, `/checkout/$treeName/$branch`, `/checkout/$treeName/$branch/$hash`, `/c/treeName/$branch`, and `/c/$treeName/$branch/$hash`

Closes #1476


backend/kernelCI_app/tests/integrationTests/treeLatest_test.py
```diff
@@ -33,52 +33,26 @@
...
             {
                 "origin": "maestro",
-                "git_commit_hash": "1064be81d694bc2b3c308de6b2951c21a3f89677",
+                "commit_hash": "1064be81d694bc2b3c308de6b2951c21a3f89677",
             },
             False,
         ),
-    ],
-)
-def test_tree_latest(tree_name, git_branch, query, has_error_body):
```
LucasSantos27, line 42: omg I'll need to refactor my PR :cry:

backend/kernelCI_app/views/treeLatestView.py
```diff
@@ -25,46 +22,57 @@
...
 
         if tree_data is None:
-            return create_api_error_response(
-                error_message=tree_not_found_error_message,
-                status_code=HTTPStatus.OK,
+            tree_not_found_error_message = ClientStrings.TREE_NOT_FOUND
+            if request.GET.get("origin") is None:
+                tree_not_found_error_message += (
+                    f" {ClientStrings.TREE_LATEST_DEFAULT_ORIGIN} {DEFAULT_ORIGIN}"
+                )
```
LucasSantos27, line 63: nice catch

PR #1510, opened 2025-09-17
Feat: show independent issue tests

### LINKED ISSUE #1290: Tests related to unexistent builds don't appear in the table
This problem was found in an issue details page, which said that it had incidents but there was no table to be seen.

This is because those issues had incidents with tests that were related to builds that didn't exist. Example of issue details: [redhat issue 2710](https://dashboard.kernelci.org/issue/redhat%3Aissue_2701?iv=1749443907&o=redhat)

If a test is related to a build that exists but the build is related to a checkout that doesn't exist, then it appears normally, such as in this issue details: [redhat issue 3684](https://dashboard.kernelci.org/issue/redhat%3Aissue_3684?iv=1743992875&o=redhat)

MarceloRobert:
Some incidents got confused, because the first incident of an issue does not consider its version, meaning that it is possible that an issue (id, version) has 0 incidents, but has a first_incident: when its _first_ version has it.

Along with that, there were a bunch of issues where there were incidents relating them to nonexistent tests, making it confusing in the debug to match the incident count. Not only that, but showing the nonexistent tests might reveal tests which weren't properly inserted into the database or incidents that were wrongfully generated.

## Changes
- Shows issue version in the first incident section of the issue details page
- Shows nonexistent tests on an issue incidents table

## How to test
- Go to an issue details page and check the first incident section and the tests table

This [localhost redhat issue](http://localhost:5173/issue/redhat%3Aissue_2701?iv=1749443907&o=redhat) is the one linked in the bug description, it has 0 incidents, but an earlier version does, and the first incident section now shows that the first incident doesn't belong to the issue on the page

This other [localhost redhat issue](http://localhost:5173/issue/redhat%3Aissue_3432?iv=1740066292&o=redhat) has 35 incidents on nonexistent tests

Both examples use kcidb on Google Cloud

Closes #1290


backend/kernelCI_app/queries/issues.py
```diff
@@ -66,7 +66,7 @@ def get_issue_tests(*, issue_id: str, version: Optional[int]) -> list[dict]:
 
     query = f"""
         SELECT
-            T.ID,
+            INC.TEST_ID AS ID,
```
LucasSantos27, line 69: should not be a difference, right?

backend/kernelCI_app/queries/issues.py
```diff
@@ -78,14 +78,15 @@ def get_issue_tests(*, issue_id: str, version: Optional[int]) -> list[dict]:
             C.GIT_REPOSITORY_URL
         FROM
             INCIDENTS INC
-            INNER JOIN TESTS T ON (INC.TEST_ID = T.ID)
+            LEFT JOIN TESTS T ON (INC.TEST_ID = T.ID)
```
LucasSantos27, line 81: with this change, we could have cases with nullable tests, right?

backend/kernelCI_app/queries/issues.py
LucasSantos27: I got it! So there are some cases where an incident can exist, but the test doesn't

dashboard/src/components/BootsTable/BootsTable.tsx
```diff
@@ -141,6 +142,10 @@ export function BootsTable({
     (): TTestByCommitHashResponse => ({
       tests: testHistory
         ? testHistory.map((e): TestByCommitHash => {
+            if (!e.path) {
+              e.path = EMPTY_VALUE;
+            }
+
```
LucasSantos27, line 148: Can status be NULL as well?

backend/kernelCI_app/queries/issues.py
MarceloRobert: > So there are some cases where an incident can exist, but the test doesn't

Exactly

backend/kernelCI_app/queries/issues.py
MarceloRobert: There will be cases when the link goes to a nonexistent test, if that's what you mean

LucasSantos27 (approved):

Issue #1496, opened 2025-09-15
Add tests for helpers


LucasSantos27:
- [x] **commonDetails.py** - current: 43%
- [x] **database.py** - current: 33%
- [x] **detailsIssues.py** - current: 25%
- [x] **discordWebhook.py** - current: 35%
- [x] **email.py** - current: 18%
- [x] **errorHandling.py** - current: 75%
- [x] **filters.py** - current: 31%
- [x] **hardwareDetails.py** - current: 19%
- [x] **issueExtras.py** - current: 23%
- [x] **issueListing.py** - current: 59%
- [x] **logger.py** - current: 31%
- [x] **misc.py** - current: 30%
- [x] **system.py** - current: 38%
- [x] **treeDetails.py** - current: 15%
- [x] **trees.py** - current: 31%

victor-accarini:
For the next tests split the commits into smaller PRs, it's easier and faster to review/merge/test.

MarceloRobert:
Tests for all files have been added and merged

PR #1543, opened 2025-10-01
test: add unit tests for filters helpers

LucasSantos27:
**Increase to 49%**

- part of #1496

### DIFF
--- a/backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
+++ b/backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
@@ -20,67 +20,20 @@ from kernelCI_app.helpers.filters import (
 )
 from kernelCI_app.constants.general import UNCATEGORIZED_STRING
 from kernelCI_app.tests.unitTests.helpers.fixtures.filter_fixtures import (
-    boot_filter_data_with_origin,
-    boot_filter_data_with_platform,
-    boot_path_filter,
-    build_filter_data_with_origin,
-    empty_filters,
-    filter_body_data,
-    empty_filter_body,
-    filter_data_with_tuple_match,
-    filter_data_with_tuple_no_match_id,
-    filter_data_with_tuple_no_match_version,
-    filter_data_with_uncategorized,
-    filter_data_with_uncategorized_issue,
+    BOOT_FILTER_DATA_SCENARIOS,
+    FILTER_DATA_SCENARIOS,
+    FILTER_OBJECTS,
+    FILTER_BODY_DATA,
     filter_params_with_exact_filter,
     filter_params_with_filters,
     filter_params_with_invalid_op,
-    invalid_issue_tuple_filter,
-    issue_filters,
+    MATCHING_ISSUE_FILTERS,
     mock_request,
-    no_filter_body,
     mock_request_with_filters,
     mock_request_with_multiple_filters,
     mock_request_with_string_filter,
     mock_request_with_regex_filter,
     mock_request_with_invalid_param,
-    non_matching_issue_filters,
-    string_filter_body,
-    exact_filter_body,
-    issue_tuple_filter,
-    issue_tuple_null_version_filter,
-    platform_filter,
-    origin_filter,
-    issue_categories_filter,
-    build_origin_filter,
-    boot_origin_filter,
-    test_origin_filter,
-    boot_duration_lte_filter,
-    boot_duration_gte_filter,
-    test_duration_lte_filter,
-    test_duration_gte_filter,
-    build_duration_lte_filter,
-    build_duration_gte_filter,
-    boot_status_filter,
-    test_path_filter,
-    test_status_filter,
-    config_name_filter,
-    compiler_filter,
-    architecture_filter,
-    hardware_filter,
-    build_status_filter,
-    uncategorized_filters,
-    valid_culprit_filter,
-    valid_option_filter,
-    uncategorized_issue_filter,
-    invalid_culprit_filter,
-    invalid_option_filter,
-    build_filter_data,
-    record_filter_data,
-    boot_filter_data,
-    test_filter_data,
-    test_filter_data_with_platform,
-    test_filter_data_with_origin,
 )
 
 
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/tests/unitTests/helpers/fixtures/filter_fixtures.py
```diff
@@ -0,0 +1,600 @@
...
+        "status": "PASS",
+        "duration": 100,
+        "issue_id": "issue123",
+        "issue_version": 1,
+        "incident_test_id": "test123",
+        "origin": "origin2",
+    }
+
+
+def test_filter_data():
```
gustavobtflores: maybe it’s a nit, but wouldn’t it be better to avoid naming the fixture functions with the `test_` prefix so pytest doesn’t collect them as tests?

backend/kernelCI_app/tests/unitTests/helpers/fixtures/filter_fixtures.py
```diff
@@ -0,0 +1,600 @@
...
+        "path": "test.other",
+        "status": "PASS",
+        "duration": 100,
+        "issue_id": "issue123",
+        "issue_version": 1,
+        "incident_test_id": "test123",
+    }
+
+
+def test_filter_data_with_platform():
```
gustavobtflores: ditto

backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
```diff
@@ -0,0 +1,1355 @@
...
+    def test_is_exclusively_build_issue_with_known_issue_no_test(self):
+        """Test is_exclusively_build_issue with known issue no test."""
+        result = is_exclusively_build_issue("issue123", 1, None)
+        assert result is True
+
+    def test_is_exclusively_build_issue_with_known_issue_with_test(self):
+        """Test is_exclusively_build_issue with known issue with test."""
+        result = is_exclusively_build_issue("issue123", 1, "test123")
+        assert result is False
+
```
gustavobtflores: here isn't worth to add also a test for a unknown issue with valid test id? or its an impossible scenario?

```suggestion

    def test_is_exclusively_build_issue_with_unknown_issue_with_test(self):
        """Test is_exclusively_build_issue with unknown issue but with test."""
        result = is_exclusively_build_issue(None, None, "test123")
        assert result is False
        
```

backend/kernelCI_app/tests/unitTests/helpers/fixtures/filter_fixtures.py
```diff
@@ -0,0 +1,600 @@
...
+        "incident_test_id": "test123",
+        "origin": "origin2",
+    }
+
+
+def test_filter_data():
+    """Test filter data for testing."""
+    return {
+        "path": "test.other",
+        "status": "PASS",
```
gustavobtflores: maybe it’s a nit, but wouldn’t it be better to avoid naming the fixture functions with the `test_` prefix so pytest doesn’t collect them as tests?

backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
```diff
@@ -0,0 +1,1326 @@
...
+        result = to_int_or_default("", 0)
+        assert result == 0
+
+    def test_to_int_or_default_with_negative_number(self):
+        """Test to_int_or_default with negative number."""
+        result = to_int_or_default("-123", 0)
+        assert result == -123
+
+
+class TestFilterParamsInitialization:
```
MarceloRobert: @gustavobtflores did you get why we have that flag of `process_body`?

backend/kernelCI_app/tests/unitTests/helpers/fixtures/filter_fixtures.py
```diff
@@ -0,0 +1,330 @@
...
+    return [
+        {"field": "test.status", "value": "PASS", "comparison_op": "exact"},
+        {"field": "test.status", "value": "FAIL", "comparison_op": "exact"},
+    ]
+
+
+def filter_params_with_filters():
+    """FilterParams with pre-configured filters."""
+    from kernelCI_app.helpers.filters import FilterParams
+    from django.http import QueryDict
```
MarceloRobert: These are imported inside 3 functions, why not import it in the file?

backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
```diff
@@ -0,0 +1,1326 @@
...
+    def test_validate_comparison_op_valid(self):
+        """Test validate_comparison_op with valid operator."""
+        filter_params = FilterParams(mock_request(), process_body=False)
+        filter_params.validate_comparison_op("exact")
+
+    def test_validate_comparison_op_invalid(self):
+        """Test validate_comparison_op with invalid operator."""
+        filter_params = FilterParams(mock_request(), process_body=False)
+        with pytest.raises(InvalidComparisonOPError):
+            filter_params.validate_comparison_op("invalid_op")
```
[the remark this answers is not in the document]
gustavobtflores: since we are testing the functionality of the FilterParams here, i don't think it should be mocked. if we mock the class, we would be testing the mock functionality, not the class functionality itself

backend/kernelCI_app/tests/unitTests/helpers/fixtures/filter_fixtures.py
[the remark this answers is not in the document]
MarceloRobert: Sorry if my comment was confusing, but I meant that the dicts could be defined where they are used in the tests. No need for either functions or constants because even with constants they would be single use

backend/kernelCI_app/tests/unitTests/helpers/filters_helpers_test.py
MarceloRobert: That's basically it. That happens because we request some pages such as tree details with `GET` requests, while the hardware details is with `POST` requests

PR #1545, opened 2025-10-01
test: add unit tests for remaining helpers

LucasSantos27:
**Increase to 47**

100% coverage for
- commonDetails
- database
- discordWebhook
- email
- errorHandling
- logger
- misc
- system

part of #1496

### DIFF
--- a/backend/kernelCI_app/tests/unitTests/helpers/discordWebhook_test.py
+++ b/backend/kernelCI_app/tests/unitTests/helpers/discordWebhook_test.py
@@ -1,0 +1,235 @@
+from unittest.mock import patch
+import requests
+from kernelCI_app.helpers.discordWebhook import (
+    validate_notification,
+    send_discord_notification,
+    AVATAR_URL,
+    WEBHOOK_NAME,
+)
+
+
+class TestValidateNotification:
+    @patch("kernelCI_app.helpers.discordWebhook.log_message")
+    def test_validate_notification_no_url(self, mock_log_message):
+        """Test validate_notification with no URL or with empty URL."""
+        result = validate_notification(url=None, content="Test content", embeds=None)
+
+        assert result is False
+        mock_log_message.assert_called_once_with(
+            "DISCORD_WEBHOOK_URL environment variable is not set."
+        )
+
+    @patch("kernelCI_app.helpers.discordWebhook.log_message")
+    def test_validate_notification_no_content_or_embeds(self, mock_log_message):
+        """Test validate_notification with no or empty content or embeds."""
+        result = validate_notification(
+            url="https://discord.com/api/webhooks/test", content=None, embeds=None
+        )
+
+        assert result is False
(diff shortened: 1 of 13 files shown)


backend/kernelCI_app/tests/unitTests/helpers/discordWebhook_test.py
```diff
@@ -0,0 +1,235 @@
...
+import requests
+from kernelCI_app.helpers.discordWebhook import (
+    validate_notification,
+    send_discord_notification,
+    AVATAR_URL,
+    WEBHOOK_NAME,
+)
+
+
+class TestValidateNotification:
```
MarceloRobert, line 11: Same thing with another batch of unit tests, it seems a bit too manual to compare the messages by hand... maybe we should be returning an error code or something. For now I think it's ok since, as I said in the other PR, if the message changes it will be caught by the tests.

backend/kernelCI_app/tests/unitTests/helpers/misc_test.py
MarceloRobert, line 1: This one feels like there too many functions on /helpers/misc. I don't see any future distinction between "build" misc and environment misc, so the build variants could be removed and the `environment_misc` functions are 100% equivalent to them already. Could you check if it is possible to do that? There would be 2 less functions to add unit tests to

backend/kernelCI_app/tests/unitTests/helpers/system_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @patch("kernelCI_app.helpers.system.settings")
+    def test_get_running_instance_none_when_no_debug_attr(self, mock_settings):
+        """Test get_running_instance returns None when DEBUG attribute doesn't exist."""
+        mock_settings.CORS_ALLOWED_ORIGINS = ["https://staging.com"]
+        delattr(mock_settings, "DEBUG")
+
+        result = get_running_instance()
+
+        assert result is None
```
MarceloRobert, line 55: This is really close to `test_get_running_instance_staging`, would there be a difference between `DEBUG=False` and no `DEBUG`?

backend/kernelCI_app/tests/unitTests/helpers/system_test.py
MarceloRobert, line 1: Could you make a case for the production CORS but with `DEBUG=True`, please?

MarceloRobert (changes requested):

backend/kernelCI_app/tests/unitTests/helpers/misc_test.py
MarceloRobert: Thanks!

backend/kernelCI_app/tests/unitTests/helpers/system_test.py
MarceloRobert: Thanks

MarceloRobert (approved):
Looks great, thanks for the misc refactor

PR #1548, opened 2025-10-03
Update composer setup

nuclearcat:
We need to update composer setup:
1)Add dashboard ingester
2)Clean up entries that are not used anymore


MarceloRobert:
This PR was continued and finished in #1723

PR #1554, opened 2025-10-07
Fix metrics

nuclearcat:
This fixes are just minimal to run, they might be not right, so feel free to drop and reimplement proper ones if necessary.

1)We cannot use port 8001, as its used by staging instance, but env variable didnt work, small fix.
2)It didnt run with 5 gunicorn workers, at least i noticed errors, and this is way to lock/run metrics in first thread.

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from utils.validation import is_boolean_or_string_true
 import os
 import json
+import socket
 import threading
 from django_prometheus import exports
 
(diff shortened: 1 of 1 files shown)


backend/kernelCI/settings.py
```diff
@@ -365,18 +366,29 @@ def get_json_env_var(name, default):
...
         try:
+            sock.bind(('0.0.0.0', PROMETHEUS_METRICS_PORT))
+            sock.close()
+            # Port is free, we can start the server
+            print(f"Starting Prometheus metrics server on port {PROMETHEUS_METRICS_PORT}")
             exports.SetupPrometheusEndpointOnPort(
                 PROMETHEUS_METRICS_PORT, addr="0.0.0.0"
             )
+        except OSError:
+            # Port already in use, another worker has started the server
```
LucasSantos27, line 383: Please add a comment here just to track

PR #1562, opened 2025-10-08
Feat: tree listing denormalization

### LINKED ISSUE #1558: Create table, view, and logic for treeView
With the denormalization, we must create the table and logic, and then we can create a new endpoint that fetches the tree listing data from this new table. Later on we can replace the current tree listing endpoints with the new one.

MarceloRobert:
## Changes

- Adds a new table with the direct TreeListing data, without the need for processing
- Adds a new view to consume that data
- Updates ingester to insert the data into PendingBuilds
- Updates the processPending command to also process the data for treeListing
  - While at it, also refactors a couple of functions in order to be reused between hardware and treeListing denormalization
  - Also renames the ProcessedHardwareStatus to be more generic and also used by the TreeListing

## Benefits

Testing with a direct call to the old and new tree listing endpoints, the average of 10 calls was:

<table><thead>
  <tr>
    <th></th>
    <th>average time</th>
  </tr></thead>
<tbody>
  <tr>
    <td>old</td>
    <td>313.2ms</td>
  </tr>
  <tr>
    <td>new</td>
    <td>23.1ms</td>
  </tr>
</tbody>
</table>

This represents a **13.56x** improvement in performance with a small amount of data. Since the query for the new listing is a single `SELECT` statement, the response time won't grow as fast as the old query when more data is present, meaning that there will be an even better ratio. The tests were made with 50000 real-world submissions locally.

## How it works

When a new checkout is received, it adds the tree to treeListing with fresh numbers. If that tree already exists and the checkout is newer, the tree will be updated and all counts will reset to 0; if the checkout is older then it does nothing. It will show the count as 0 until the pending tests and builds are processed. Showing 0 counts is useful in order to see if a checkout worked but the builds/tests weren't send for some reason, and such problem should be seen and addressed as soon as possible.

The processing of new builds follows what has been done for the hardwareListing. When a new build is received it updates the `PendingBuilds` table, and when the `process_pending_aggregations` command is executed it will count the pending build's status on a single treeListing item, then this item will update the treeListing table only for the checkouts that are already existing (so that builds of older checkouts don't update the counts of newer checkouts for a tree).

## How to test

- Run the migrations for the new table
- Start the ingester
  - Create a temporary folder for the submissions as `backend/kernelCI_app/management/commands/tmp_submissions` (you can change the path, just update the command below)
  - Make sure you have your trees-name.yml file somewhere. It usually is on `/backend/volume_data` as a result of the `treeproof` command. If you are feeling lazy and you don't want to run the command, there's an example file [here](https://drive.google.com/file/d/1JpWJ2zoo5wDgDBCtLKC_EEOn-EEzkp2V/view?usp=sharing) (Google Drive link)
  - Make sure you have a couple of example submission files for testing. You can grab some in this zip file [here](https://drive.google.com/file/d/13oIkTYWleOWLeBp1o3aFtb3ydlVS8Xd-/view?usp=sharing) (Google Drive link)
  - Export the environment variables in your terminal, pointing one of the databases to the local one. Make sure the local one is set as the default one with `USE_DASHBOARD_DB=True`
  - _Now_ you'll be able to run the command with `poetry run python3 ./manage.py monitor_submissions --spool-dir kernelCI_app/management/commands/tmp_submissions --trees-file volume_data/trees-name.yaml` (being in the `/backend` directory, update the paths if necessary)
- Insert submissions in the folder and check that they are being inserted in the database
You can also test the ingester on docker, make sure the environment variables are on `.env.backend` and use `docker compose run --rm backend` before running the command shown above.

Closes #1558


backend/kernelCI_app/management/commands/helpers/denormal.py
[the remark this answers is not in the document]
MarceloRobert: that file won't be needed anymore

backend/kernelCI_app/management/commands/helpers/denormal.py
```diff
@@ -0,0 +1,55 @@
+from kernelCI_app.models import TreeListing
+
+
+def _update_tree_listing(*, tree_listing_objects: list[TreeListing]) -> None:
```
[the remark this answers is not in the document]
MarceloRobert: ditto not needed anymore

backend/kernelCI_app/models.py
```diff
@@ -24,6 +26,22 @@ class SimplifiedStatusChoices(models.TextChoices):
     INCONCLUSIVE = "I"
 
 
+SUFFIX_MAP = {
+    "_pass": SimplifiedStatusChoices.PASS,
+    "_failed": SimplifiedStatusChoices.FAIL,
+    "_inc": SimplifiedStatusChoices.INCONCLUSIVE,
+}
```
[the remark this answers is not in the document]
MarceloRobert: I had made this so it could eventually be used for HardwareListing as well, but since I went back to using dicts then I won't need this anymore, the usage will be like the one for hardwareListing (mapping a status to a tuple and then incrementing the specific tab's status with that tuple)

backend/kernelCI_app/management/commands/helpers/aggregation_helpers.py
```diff
@@ -36,6 +38,73 @@ def convert_test(t: Tests) -> PendingTest:
...
+            checkout.start_time,
+            checkout.origin_builds_finish_time,
+            checkout.origin_tests_finish_time,
+        )
+        for checkout in checkouts_instances
+    ]
+
+    with connections["default"].cursor() as cursor:
+        # Set values as 0 when inserting a new tree and update as 0 when tree already exists
+        cursor.executemany(
```
MarceloRobert: query is outdated, will update

backend/kernelCI_app/models.py
```diff
@@ -322,15 +322,68 @@ class Meta:
...
+    origin = models.CharField(max_length=100)
+    checkout_id = models.TextField()
+    status = models.CharField(max_length=1, choices=SimplifiedStatusChoices.choices)
+
+    class Meta:
+        db_table = "pending_builds"
+
+
+class TreeListing(models.Model):
+    id = models.AutoField(primary_key=True)
```
barbieri: worth documenting why we're not using the composite primary key

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -182,22 +204,176 @@ def _process_build_status(
...
+            build_listing_key=build_keys_by_id[build.build_id],
+            status_record=status_record,
+            existing_processed=existing_processed,
+            new_processed_entries=new_processed_entries,
+        )
+
+    return tree_listing_data, new_processed_entries
+
+
+def _collect_test_contexts(
```
barbieri: add `platform` or `hardware` to the name to avoid confusion with more general tests...

since it's called by `aggregate_hardware_status` I suggest `_collect_hardware_status_contexts()`

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -182,22 +204,176 @@ def _process_build_status(
...
+        keys_to_check.add(test_key)
+        keys_to_check.add(build_key)
+
+    return contexts, keys_to_check
+
+
 def aggregate_hardware_status(
     tests_instances: Sequence[PendingTest],
-    builds_by_id: dict[str, Builds],
+    test_builds_by_id: dict[str, Builds],
```
barbieri: why are you doing this rename?

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -408,21 +618,220 @@ def _process_ready_tests(
...
+                )
+
+                (
+                    ready_builds,
+                    build_checkouts_by_id,
+                    last_processed_build_id,
+                    skipped_no_checkout,
+                ) = self._get_ready_builds(
+                    last_processed_build_id=last_processed_build_id,
+                    batch_size=batch_size,
```
barbieri: since this is only used by `_aggregate_tree_listing`, whenever you reorganize into smaller functions, move this block + `_delete_ready_builds` inside it...

that is:

1. here in this function you call `_get_ready_tests()` since it's shared.
2. call `self._process_hardware_status(ready_tests, test_builds_by_id)`, internally it will call `aggregate_hardware_status()` and `self._process_new_processed_entries(new_processed_entries_hardware)`
3. call `self._process_tree_listing(ready_tests, test_builds_by_id)`, internally it will call  `_get_ready_builds()`, `aggregate_tree_listing()`, `self._process_new_processed_entries(new_processed_entries_tree)` and then `self._delete_ready_builds(ready_builds)`.
4. this function cleans up `self._delete_ready_tests(ready_tests)`

backend/kernelCI_app/helpers/trees.py
```diff
@@ -87,13 +88,23 @@ def sanitize_tree(
         "skip": checkout["skip_boots"],
     }
 
-    if isinstance(checkout.get("git_commit_tags"), str):
+    # Has to check if it's a string because sqlite doesn't support ArrayFields.
+    # So if the query came from sqlite, it will be a string.
+    git_commit_tags = checkout.get("git_commit_tags")
+    if isinstance(git_commit_tags, str):
```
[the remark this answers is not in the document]
MarceloRobert: I altered this change, but it isn't needed for this PR, so I moved it to #1686

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -278,87 +454,121 @@ def handle(self, *args, **options):
...
+                    checkout_id, origin, tree_name,
+                    git_repository_url, git_repository_branch,
+                    git_commit_hash, git_commit_name,
+                    git_commit_tags, start_time,
+                    build_pass, build_failed, build_inc,
+                    boot_pass, boot_failed, boot_inc,
+                    test_pass, test_failed, test_inc
                 )
+                VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+                ON CONFLICT (origin, tree_name, git_repository_url, git_repository_branch) DO UPDATE SET
```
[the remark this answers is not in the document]
gustavobtflores: @MarceloRobert i think we also need to update the other git_* values, no?

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -408,21 +616,215 @@ def _process_ready_tests(
...
+
+                tree_listing_data, new_processed_entries_tree = aggregate_tree_listing(
+                    ready_tests,
+                    test_builds_by_id,
+                    ready_builds,
+                    build_checkouts_by_id,
+                )
+                self._process_tree_listing(
+                    tree_listing_data, new_processed_entries_tree
                 )
```
gustavobtflores: the function is receiving an additional argument `new_processed_entries_tree`, that seems not used

backend/kernelCI_app/management/commands/process_pending_aggregations.py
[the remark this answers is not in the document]
MarceloRobert: true, we just don't need to update the fields used on `on conflict`, but I missed those other ones

backend/kernelCI_app/management/commands/process_pending_aggregations.py
MarceloRobert: Since `_get_ready_builds` uses `last_processed_build_id` and `skipped_no_checkout`, which are variables declared/used within this loop, I think that they could stay inside the loop as well. The similarity between processing the tests for hardware and builds for trees also makes it easier to group the functions and understand them IMO; I made some small changes as you proposed, not much.

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -108,26 +159,34 @@ def _collect_test_contexts(
...
+    Checks if test is already processed, if not,
+    updates status record and marks it as processed.
+    """
+    # TODO: we should be checking if it is already processed before entering this function
+    to_process = ProcessedListingItems(
+        listing_item_key=test_listing_key, checkout_id=checkout_id, status=test.status
     )
 
     if to_process in existing_processed:
         return False
```
gustavobtflores: i think here we have the culprit on why the countings in the tree listing are incorrect, we could check if the existing_processed has a different status and then count a -1 or +1 to the tree that was affected by the status change

gustavobtflores (approved):
LGTM, would be nice to have an issue to track the linaro/redhat differences

PR #1574, opened 2025-10-15
Refactor: ingester changes

### LINKED ISSUE #1571: Refactor dashboard ingester
Some functions can be made cleaner, and some can be extracted to separate files in order to better organize the different parts of the ingester. A cleaner code will help move further with additions and modifications to it.

MarceloRobert:
## Changes
Refactors ingester code:
- Moves constants to a specific file (also makes constants able to be changed from env vars)
- Moves some functions to more specific files
- Adds more typing indicators
- Lowers the complexity of the most complex functions

## How to test
The behavior should be the same as before. So:
- Start the ingester
  - Create a temporary folder for the submissions as `backend/kernelCI_app/management/commands/tmp_submissions` (you can change the path, just update the command below)
  - Make sure you have your trees-name.yml file somewhere. It usually is on `/backend/volume_data` as a result of the `treeproof` command. If you are feeling lazy and you don't want to run the command, there's an example file [here](https://drive.google.com/file/d/1JpWJ2zoo5wDgDBCtLKC_EEOn-EEzkp2V/view?usp=sharing) (Google Drive link)
  - Make sure you have a couple of example submission files for testing. You can grab some in this zip file [here](https://drive.google.com/file/d/13oIkTYWleOWLeBp1o3aFtb3ydlVS8Xd-/view?usp=sharing) (Google Drive link)
  - Export the environment variables in your terminal, pointing one of the databases to the local one. Make sure the local one is set as the default one with `USE_DASHBOARD_DB=True`
  - _Now_ you'll be able to run the command with `poetry run python3 ./manage.py monitor_submissions --spool-dir kernelCI_app/management/commands/tmp_submissions --trees-file volume_data/trees-name.yaml` (being in the `/backend` directory, update the paths if necessary)
- Insert submissions in the folder and check that they are being inserted in the database
You can also test the ingester on docker, make sure the environment variables are on `.env.backend` and use `docker compose run --rm backend` before running the command shown above.

Closes #1571

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -130,6 +130,64 @@ def consume_buffer(buffer: list[TableModels], item_type: TableNames) -> None:
     out("bulk_create %s: n=%d in %.3fs" % (item_type, len(buffer), time.time() - t0))
 
 
+def flush_buffers(
+    *,
+    issues_buf: list[Issues],
+    checkouts_buf: list[Checkouts],
+    builds_buf: list[Builds],
+    tests_buf: list[Tests],
+    incidents_buf: list[Incidents],
+) -> None:
+    """
+    Consumes the list of objects and tries to insert them into the database.
+    """
+    total = (
+        len(issues_buf)
+        + len(checkouts_buf)
+        + len(builds_buf)
+        + len(tests_buf)
+        + len(incidents_buf)
+    )
+
+    if total == 0:
+        return
+
+    # Insert in dependency-safe order
+    flush_start = time.time()
+    try:
+        # Single transaction for all tables in the flush
+        with transaction.atomic():
+            consume_buffer(issues_buf, "issues")
+            consume_buffer(checkouts_buf, "checkouts")
+            consume_buffer(builds_buf, "builds")
+            consume_buffer(tests_buf, "tests")
+            consume_buffer(incidents_buf, "incidents")
+    except Exception as e:
+        logger.error("Error during bulk_create flush: %s", e)
+    finally:
+        flush_dur = time.time() - flush_start
+        rate = total / flush_dur if flush_dur > 0 else 0.0
+        msg = (
+            "Flushed batch in %.3fs (%.1f items/s): "
+            "issues=%d checkouts=%d builds=%d tests=%d incidents=%d"
+            % (
+                flush_dur,
+                rate,
+                len(issues_buf),
+                len(checkouts_buf),
+                len(builds_buf),
+                len(tests_buf),
+                len(incidents_buf),
+            )
+        )
+        out(msg)
+        issues_buf.clear()
+        checkouts_buf.clear()
+        builds_buf.clear()
+        tests_buf.clear()
+        incidents_buf.clear()
+
+
 # TODO: lower the complexity of this function
 def db_worker(stop_event: threading.Event) -> None:  # noqa: C901
     """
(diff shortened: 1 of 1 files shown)


backend/kernelCI_app/management/commands/helpers/log_excerpt_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        logexcerpt: the unchanged logexcerpt
+        id: the hash of the logexcerpt
+
+    Returns:
+        str: On success upload: the reference url. On failed upload: the original logexcerpt
+    """
+
+    upload_url = f"{STORAGE_BASE_URL}/upload"
+    if VERBOSE:
+        logger.info("Uploading logexcerpt for %s to %s", id, upload_url)
```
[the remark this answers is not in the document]
MarceloRobert: You mean the description for the `VERBOSE` variable or the string in this specific `logger.info`?

backend/kernelCI_app/management/commands/helpers/process_submissions.py
```diff
@@ -128,12 +137,12 @@ def make_incident_instance(incident) -> Incidents:
     return obj
 
 
-def build_instances_from_submission(data: dict[str, Any]) -> dict[TableNames, list]:
+def build_instances_from_submission(data: dict[str, Any]) -> ProcessedSubmission:
```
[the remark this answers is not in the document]
MarceloRobert: this data is the submission json, so the `Any` means any value that an item can have: str, int, list, dict, none... IMO in this case it is ok to leave it as `Any` (and other similar cases around these functions too). This use will at least define that they dict keys are `str`, otherwise we would be using just `dict` which just means `dict[Any, Any]`

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -498,37 +338,16 @@ def process_file(filename, trees_name, spool_dir):
...
-            pass
-
-    _out(
-        "Spool status: %d .json files queued (%.2f MB)"
-        % (
-            len(json_files),
-            total_bytes / (1024 * 1024) if total_bytes else 0.0,
-        )
-    )
+    json_files, total_bytes = get_spool_files(spool_dir)
```
nuclearcat: if not json_files: maybe? i there is no spool files, we might need to put small delay and check again

backend/kernelCI_app/typeModels/modelTypes.py
```diff
@@ -0,0 +1,13 @@
+from typing import Literal
+from kernelCI_app.models import Builds, Checkouts, Incidents, Issues, Tests
+
+type TableNames = Literal["issues", "checkouts", "builds", "tests", "incidents"]
+type TableModels = Issues | Checkouts | Builds | Tests | Incidents
```
nuclearcat: I am not sure in this part...
`TableModels = Issues | Checkouts | Builds | Tests | Incidents` denotes instances, but MODEL_MAP stores classes and consume_buffer expects a class

nuclearcat (changes requested):

backend/kernelCI_app/typeModels/modelTypes.py
MarceloRobert: does it denote instances? IIUC it denotes the classes themselves, hovering over the `TableModels` variable I see `(type) TableModels = type[Issues] | type[Checkouts]...` directly

backend/kernelCI_app/management/commands/helpers/file_utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def get_spool_files(spool_dir: str) -> tuple[list[str], int]:
+    json_files = [
+        f
+        for f in os.listdir(spool_dir)
+        if os.path.isfile(os.path.join(spool_dir, f)) and f.endswith(".json")
+    ]
+    if not json_files:
+        return
```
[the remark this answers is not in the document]
MarceloRobert: Changed with the rebase, not needed anymore

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -130,6 +130,64 @@ def consume_buffer(buffer: list[TableModels], item_type: TableNames) -> None:
...
 
 
+def flush_buffers(
+    *,
+    issues_buf: list[Issues],
+    checkouts_buf: list[Checkouts],
+    builds_buf: list[Builds],
+    tests_buf: list[Tests],
+    incidents_buf: list[Incidents],
+) -> None:
```
AmadeusK525: We could create a helper object that holds these lists, so we can just pass the object instead of passing 5 different lists

PR #1581, opened 2025-10-20
Feat: add tests_origin_time_platform index

### LINKED ISSUE #1578: Formalize new hardware listing index
@nuclearcat tried out the following index:
```
CREATE INDEX CONCURRENTLY tests_origin_start_time_platform_idx
ON tests (origin, start_time)
WHERE (environment_misc ->> 'platform') IS NOT NULL;
```

this index seems to have worked well, and should be formalized with django

MarceloRobert:
This index helps in the hardware listing query, being very specific but very helpful.
At the time of coding, this index was already added manually to the database, so some workarounds have to be made. The renaming is required because Django has a limit of 30 characters for index names.

Check Django's documentation on the index condition:
- See the PostgreSQL comment on https://docs.djangoproject.com/en/5.2/topics/db/queries/#module-django.db.models.fields.json
- Also check out the next section, https://docs.djangoproject.com/en/5.2/topics/db/queries/#complex-lookups-with-q-objects

## Changes
- Added new index in tests model
- Added new migration manually
- Updated docs

## How to test
Up the database and run the migrations. You should see the new index in the tests table.
You can use the following query to see all tests indexes:
```
SELECT
    *
FROM
    pg_indexes
WHERE
    tablename = 'tests';
```
Also test when the old index name is present, it should rename it first.

Closes #1578

### DIFF
--- a/backend/docs/database-logic.md
+++ b/backend/docs/database-logic.md
@@ -134,4 +134,6 @@ Django also creates some indexes automatically:
 - Index for foreign key as b-tree;
 - Index for foreign key as b-tree with `text_pattern_ops`, specialized for `LIKE` operations.
 
-One of the operations that we are doing benefits from this double indexes, which is the filtering of dummy builds (builds where id LIKE `maestro:_dummy%`)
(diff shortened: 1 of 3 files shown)


gustavobtflores (approved):
LGTM, this seems to be the optimal way to add the index in django

PR #1582, opened 2025-10-21
Hardware listing tuple test

### LINKED ISSUE #1583: Use tuple instead of dict in hardwareListing query
The `dict_fetchall` function explicitly says to not use it on large queries, and yet we are using it in the hardware listing query, one of the longest ones. We should return a simple tuple specially because it is a simple change and will improve the performance even if just a bit.

MarceloRobert:
From the k6 performance tests, the response time lowered from an average from 131ms to 127ms. The reason why the times here are much lower than in production is due to database size, but the same data and same interval were used between tests. The changes don't look big, but there's no tradeoff for these changes.

## Changes
- Fixes the script for k6 performance tests
- Reorganizes hardware listing query to follow the usual pattern of build/boot/test and to have the same status disposition in the count clause
- Removes unnecessary test_origin field from the select of a hardware listing subquery
- Changes hardware listing query to return a tuple instead of using dict zip

## How to test
- Aquire dump data (can be just for maestro, you can use the `copy_db_data` script)
- Set that dump data for k6 tests, and adjust the test file to consider the right interval (check the current interval by going to the production/staging instance and looking into the network tab)
- Run the performance tests before and after the changes

Closes #1583

### DIFF
--- a/backend/kernelCI_app/queries/hardware.py
+++ b/backend/kernelCI_app/queries/hardware.py
@@ -47,6 +47,7 @@ def _get_hardware_tree_heads_clause(*, id_only: bool) -> str:
                     C.start_time DESC"""
 
 
+# TODO: unify with get_tree_listing_count_clause
 def _get_hardware_listing_count_clauses() -> str:
     build_count_clause = """
     COUNT(DISTINCT CASE WHEN "build_status" = 'PASS' AND build_id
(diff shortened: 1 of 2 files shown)


k6/tests/hardwareListing.js
```diff
@@ -1,8 +1,13 @@
...
 import { check } from "k6";
 
+export const options = {
+    iterations: 100,
+};
+
+// TODO: have a dev-friendly time and convert it before sending the request
 export default function () {
-    const defaultRes = http.get('http://test-backend:8000/api/hardware');
+    const defaultRes = http.get('http://test-backend:8000/api/hardware?startTimestampInSeconds=1760643000&endTimestampInSeconds=1761075000&origin=maestro');
```
gustavobtflores: i think we can remove the todo, if it should get the last 5 days or something like that, you can use:

```js
const now = Math.floor(Date.now() / 1000);
const oneDay = 24 * 60 * 60;

const startTime = now - 5 * oneDay;
const endTime = now;

(...)

const defaultRes = http.get(`http://test-backend:8000/api/hardware?startTimestampInSeconds=${startTime}&endTimestampInSeconds=${endTime}&origin=maestro`);
```

docs/performanceTests.md
```diff
@@ -15,7 +15,7 @@ The Docker setup handles all K6 dependencies automatically.
 
 - **Test files**: Place your K6 test scripts in `/k6/tests/`
 - **Results**: Test outputs are saved to `/k6/results/`
-- **Database setup**: For a quick start, add your testing data as a .sql dump in `/k6/data`. See `/k6/data/README.md` for initialization instructions.
+- **Dataset**: Test data should be placed in `/k6/data`. See `/k6/data/README.md` for initialization instructions.
```
AmadeusK525: These are all relative paths, right? Use `./` instead of `/`. `/` is ambiguous with system paths.

MarceloRobert: Done

PR #1588, opened 2025-10-25
feat(ingester): some improvements

AmadeusK525:


backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -47,7 +47,7 @@
 try:
     INGEST_FLUSH_TIMEOUT_SEC = float(os.environ.get("INGEST_FLUSH_TIMEOUT_SEC", "2.0"))
 except (ValueError, TypeError):
-    logger.warning("Invalid INGEST_FLUSH_TIMEOUT_SEC, using default 2.0")
+    logger.warning("Invalid INGEST_FLUSH_TIMEOUT_SEC, using default 5.0")
     INGEST_FLUSH_TIMEOUT_SEC = 5.0
```
victor-accarini: It seems the value should be `2.0`, if we are going with 5 then line 48 also needs to be updated

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -221,13 +217,10 @@ def process_log_excerpt_from_item(
             set_log_excerpt_ofile(item, cached_url)
 
 
-def extract_log_excerpt(input_data: dict[str, Any]) -> dict[str, Any]:
+def extract_log_excerpt(input_data: dict[str, Any]) -> None:
```
victor-accarini: Don't forget to change all returns, missing a change in line 227

MarceloRobert:
Please add a description to the PR, even if small. Also, `Merging is blocked: Commits must have verified signatures.`

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -10,7 +11,7 @@
 import threading
 import time
 import traceback
-from typing import Any, Literal, Optional
+from typing import Any, Literal, Optional, Tuple
```
MarceloRobert: Prefer using `tuple` instead of `Tuple`, the latter one is deprecated

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
AmadeusK525: True, changed all of them to 2.0

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -484,41 +479,33 @@ def process_file(filename, trees_name, spool_dir):
         # Empty file, already deleted
         return True
 
-    db_queue.put((data, metadata))
+    db_queue.put((file.name, build_instances_from_submission(data)))
```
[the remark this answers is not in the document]
AmadeusK525: It's in the commit description, I'm leveraging the fact that reading files takes a couple of milliseconds and executing python code while other files may still be loading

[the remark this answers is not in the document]
MarceloRobert: Ahh right, and previously the models were being created in the single thread of `db_worker`, right?

Issue #1569, opened 2025-10-13
Create unit tests for ingester


MarceloRobert:
The ingester has become a crucial part of our repository. The performance is definitely part of our concern, but we should also guarantee that the performance changes won't break the ingester's behavior. A set of unit tests should be created for that purpose.

PR #1592, opened 2025-10-30
Test: ingester file functions unit tests

MarceloRobert:
## Changes
- Adds unit tests for the file functions used in the ingester, being:
    load_tree_names,
    move_file_to_failed_dir,
    verify_dir,
    verify_spool_dirs,

## How to test
Run the unit tests of the file and check that they are all passing.
Also check if the test cases cover the function well

Part of #1569


gustavobtflores (commented):
Do we need to mock functions like os.path.basename and os.path.join? They are kind of pure functions and do not interact with filesystem

gustavobtflores (approved):
LGTM

PR #1610, opened 2025-11-12
feat(notifications): add hardware_summary action

### LINKED ISSUE #1081: notifications: add recipients list based on hardware where the test executed
Today we have a [recipient list](https://github.com/kernelci/dashboard/blob/main/backend/kernelCI_app/management/commands/data/recipients.yaml) based on trees. We want o add an `hardware:` section, so people can subscribe to specific hw names.

ximiali:
### Description
Add a new hardware_summary action to the notifications management command. 
This action generates hardware reports based on the hardware subscription files.
### Changes
- Add a new `hardware_summary` action to the notifications command with CLI argument.
- Add functions to parse hardware subscription files, query hardware data, and normalize hardware results.
- Add Jinja template for email reports.
- Add example and Qualcomm hardware subscription files.
### How to Test
Run the following command to generate the hardware summary report:
```bash
poetry run python manage.py notifications --action hardware_summary 
```
### Related Issue
Closes #1081

### DIFF
--- a/backend/kernelCI_app/management/commands/notifications.py
+++ b/backend/kernelCI_app/management/commands/notifications.py
@@ -5,7 +5,7 @@ import os
 import sys
 
 from collections import defaultdict
-from datetime import datetime, timezone
+from datetime import datetime, timezone, timedelta
 from types import SimpleNamespace
 from urllib.parse import quote_plus
 
@@ -29,6 +29,7 @@ from kernelCI_app.management.commands.helpers.summary import (
     TreeKey,
     get_build_issues_from_checkout,
     process_submissions_files,
+    process_hardware_submissions_files,
 )
 from kernelCI_app.queries.notifications import (
     get_checkout_summary_data,
@@ -54,6 +55,12 @@ from kernelCI_cache.queries.issues import (
 )
 from kernelCI_cache.typeModels.databases import PossibleIssueType
 from kernelCI_app.utils import group_status
+
+from kernelCI_app.queries.hardware import (
+    get_hardware_summary_data,
+    get_hardware_listing_data_bulk,
+)
+from kernelCI_app.helpers.hardwares import sanitize_hardware
 
 KERNELCI_RESULTS = "kernelci-results@groups.io"
 KERNELCI_REPLYTO = "kernelci@lists.linux.dev"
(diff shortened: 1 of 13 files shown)


padovan:
It is looking great. Thanks for sharing. I am fine with the output being generated and with the yaml config file for hardware.

I'll let the Dashboard team review it in detail and comment on the implementation.

backend/kernelCI_app/queries/hardware.py
```diff
@@ -294,6 +360,80 @@ def query_records(
...
+                tests.comment,
+                tests.log_url,
+                tests.status,
+                tests.start_time,
+                tests.duration,
+                tests.misc,
+                tests.build_id,
+                tests.environment_compatible,
+                tests.misc ->> 'runtime' AS runtime,
+                tests.environment_misc ->> 'job_id' AS job_id,
```
MarceloRobert: thought you can select these values at query level, I would prefer if you got them from the already-fetched fields at the Python level, which might have less impact in the database

backend/kernelCI_app/queries/hardware.py
```diff
@@ -294,6 +360,80 @@ def query_records(
         return dict_fetchall(cursor)
 
 
+# select the detailed data depends on hardware_id
```
MarceloRobert: nit: If you want to add a comment explaining the function, prefer using a docstring at the beginning of the function

MarceloRobert (changes requested):
Could you update the notifications.md document with this new command too?

backend/kernelCI_app/queries/hardware.py
ximiali: Thanks, I’ve removed the selection at query level and now handle it in Python instead.

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -723,6 +727,93 @@ def generate_test_report(*, service, test_id, email_args, signup_folder):
...
+        signup_folder=signup_folder,
+        hardware_origins=hardware_origins,
+    )
+
+    # get detailed data for all hardware
+    hardwares_data_raw = get_hardware_summary_data(
+        keys=list(hardware_key_set),
+        start_date=start_date,
+        end_date=end_date,
+    )
```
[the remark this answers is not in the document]
ximiali: Thanks, added the check for empty hardwares_data_raw.

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -723,6 +727,93 @@ def generate_test_report(*, service, test_id, email_args, signup_folder):
...
+    service,
+    hardware_origins: Optional[list[str]],
+    email_args,
+    signup_folder: Optional[str] = None,
+):
+    """Generate monthly hardware reports for hardware submission file."""
+
+    now = datetime.now(timezone.utc)
+    start_date = now - timedelta(days=30)
+    end_date = now
```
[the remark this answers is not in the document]
ximiali: Adjusted to 7 days instead of 30, as user cannot modify this value.

MarceloRobert (commented):
If you want to have automatic reports, you have to add a cron job in `settings.py` like the other ones set up there. Also please fix the CI issues, otherwise it looks good now

MarceloRobert (approved):
Looking great

PR #1614, opened 2025-11-13
feat: hardware denormalization for hardware listing

### LINKED ISSUE #1605: Implement hardware listing async denormalization
Follow through with the denormalization idea on #1585 

The implementation should be efficient, having little impact in the ingester performance. The suggestions for better performance can be applied, such as:

- bulk update only after x amount of time instead of updating every insert
- use sqlite instead of postres for helper tables

or other suggestions

gustavobtflores:
## Objective
Implement hardware status denormalization to improve query performance for hardware listing data

## Problem Solved
The current hardware listing endpoint suffers from performance issues due to complex joins across multiple tables (checkouts, tests, builds). This PR introduces a denormalized approach to aggregate hardware status data.

## Key Changes
- Added 4 new models (HardwareStatus, LatestCheckout, PendingTest, ProcessedHardwareStatus) with appropriate indexes
- Integrated aggregation logic into the kcidbng_ingester to checkouts and pending tables during data ingestion
- Created `process_pending_aggregations` management command for batch processing of pending tests with configurable batch size and loop mode
- Added `get_hardware_listing_data_from_status_table` function to query denormalized data instead of complex joins

### 1. Run Database Migration
```bash
cd backend
poetry python manage.py migrate
```

### 2. Process Test Submissions (Optional - populate test data)
Extract submissions.zip to a folder and run the ingester:

```bash
poetry python manage.py monitor_submissions --spool-dir ./{folder-name}
```

### 3. Test Backlog Processing Command

**Single batch processing:**
```bash
poetry python manage.py process_pending_aggregations --batch-size 1000
```

**Continuous loop mode:**
```bash
python manage.py process_pending_aggregations --loop --interval 60 --batch-size 1000
```

### 4. Verify Hardware Status Data
Check that hardware_status table is populated in PostgreSQL using psql or other software that can communicate with the DB

### 5. Test Hardware Listing Endpoint
Query the hardware listing v2 endpoint and verify improved performance and correct data aggregation.

## Ingestion timings

Each benchmark consists of **10 rounds** processing **1000 submission files** per round.

| Test ID | Description                                       | Mean Time (s) | Files/s | Total Time (s) | Overhead |
|---------|---------------------------------------------------|----------------|----------|------------------|-------------------|
| **0001** | With inserts into `PendingTest` + `LatestCheckout` | **18.19**      | 54.96   | 181.96          | **+1.64s (+9.9%)** |
| **0002** | Without denormalization inserts                   | **16.55**      | 60.42   | 165.52          | baseline          |

## Query Performance Comparison

Both plans were produced using `EXPLAIN (ANALYZE, BUFFERS)` against a dataset containing **50,000 submission files** and a total of **1.7 million tests**.  

Because the dashboard only cares about **latest checkouts**, the denormalized approach reduces the effective dataset dramatically: instead of scanning all 1.7M tests, only **~170k tests/builds** meaningfully contribute to the counters, avoiding work over the majority of irrelevant historical data.

| Query | Execution Time | Buffers (read/hit/temp) | Rows Processed | Overhead Reduction |
|-------|----------------|---------------------------|----------------|---------------------|
| **Old Query** | **~253.9 ms** | **hit=6,731**, **read=24,088**, **temp=~4.9MB** | ~84,000 scanned, 109 returned | baseline |
| **New Query** | **~2.53 ms** | **hit=78**, **read=0**, **temp=0** | ~1,420 scanned, 109 returned | **−251.37 ms (≈ −99.0%)** in time<br>**−82,580 rows (≈ −98.3%)** rows scanned |

### Notes
- The **old query** performs multi-table joins, runtime aggregation, external sorting, nested loops, and touches over **1.7M test rows**, producing large temp files.  
- The **new query** operates only over pre-aggregated data (~170k relevant tests/builds), requiring just a simple hash join + hash aggregate.  
- The new query consistently returns results in **2–3 ms**, achieving ~**100× speedup**.

## Dashboard Performance

You can see the endpoint response timings in panel on the right

[dashboard-hardware-load.webm](https://github.com/user-attachments/assets/e45a5c3a-c242-417b-aab1-9b4aae687b45)

Closes #1605

### DIFF
--- a/backend/kernelCI_app/management/commands/backfill_hardware_aggregations.py
+++ b/backend/kernelCI_app/management/commands/backfill_hardware_aggregations.py
@@ -1,0 +1,133 @@
+import time
+from datetime import timedelta
+from typing import Any, Callable, Sequence
+
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from django.db.models.query import QuerySet
+from django.utils import timezone
+
+from kernelCI_app.management.commands.helpers.aggregation_helpers import (
+    aggregate_checkouts,
+    aggregate_tests,
+)
+from kernelCI_app.models import (
+    Checkouts,
+    LatestCheckout,
+    PendingTest,
+    Tests,
+)
+from kernelCI_app.helpers.logger import out
+
+
+class Command(BaseCommand):
+    help = """
+        Backfill hardware aggregations (LatestCheckout and PendingTest)
+        which can later be collected and processed by the process_pending_aggregations command.
+    """
+
+    def add_arguments(self, parser: Any) -> None:
+        parser.add_argument(
+            "--days",
+            type=int,
+            default=30,
+            help="Number of days to look back (default: 30)",
+        )
+        parser.add_argument(
+            "--truncate",
+            action="store_true",
+            help="""Truncate destination tables before backfilling.
+                This is a destructive operation and should be used with caution.""",
+        )
+        parser.add_argument(
+            "--batch-size",
+            type=int,
+            default=2000,
+            help="Batch size for processing (default: 2000)",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        days = options["days"]
+        truncate = options["truncate"]
+        batch_size = options["batch_size"]
+
+        cutoff_date = timezone.now() - timedelta(days=days)
+        out(f"Backfilling hardware aggregations since {cutoff_date}...")
+
+        if truncate:
+            out("Truncating LatestCheckout and PendingTest tables...")
+            LatestCheckout.objects.all().delete()
+            PendingTest.objects.all().delete()
+            out("Truncation complete.")
+
+        out("Backfilling LatestCheckout...")
+        checkouts_qs = Checkouts.objects.filter(start_time__gte=cutoff_date).order_by(
+            "-start_time"
+        )
+
+        # Process checkouts to insert into latest_checkout table
+        self.process_queryset(
+            checkouts_qs, aggregate_checkouts, batch_size, "Checkouts"
+        )
+
+        out("Backfilling PendingTest...")
+        tests_qs = (
+            Tests.objects.filter(
+                start_time__gte=cutoff_date,
+                environment_misc__platform__isnull=False,
+                build__checkout__id__in=LatestCheckout.objects.values_list(
+                    "checkout_id", flat=True
+                ),
+            )
+            .select_related("build")
+            .order_by("start_time")
+        )
+
+        # Process tests to insert into pending_tests table
+        self.process_queryset(tests_qs, aggregate_tests, batch_size, "Tests")
+
+        out("Backfill complete.")
+
+    def process_queryset(
+        self,
+        queryset: QuerySet[Any],
+        aggregator_func: Callable[[Sequence[Any]], None],
+        batch_size: int,
+        label: str,
+    ) -> None:
+        """
+        Process a queryset in batches using the provided aggregator function
+
+        Iterates through the queryset in chunks, collects items into batches,
+        and processes each batch using the aggregator function within a transaction
+        """
+        count = 0
+        total_processed = 0
+        batch = []
+
+        t0 = time.time()
+        for item in queryset.iterator(chunk_size=batch_size):
+            try:
+                batch.append(item)
+                count += 1
+                if count >= batch_size:
+                    with transaction.atomic():
+                        aggregator_func(batch)
+                    total_processed += count
+                    out(
+                        f"Processed {total_processed} {label} (elapsed time: {time.time() - t0:.2f}s)"
+                    )
+                    batch = []
+                    count = 0
+            except Exception as e:
+                out(f"Error processing {label}: {e}")
+                batch = []
+                count = 0
+                continue
+
+        if batch:
+            aggregator_func(batch)
+            total_processed += count
+            out(
+                f"Processed {total_processed} {label} (elapsed time: {time.time() - t0:.2f}s)"
+            )


backend/kernelCI_app/models.py
```diff
@@ -228,3 +228,73 @@ class Meta:
...
+    boot_inc = models.IntegerField(default=0)
+
+    # test status
+    test_pass = models.IntegerField(default=0)
+    test_failed = models.IntegerField(default=0)
+    test_inc = models.IntegerField(default=0)
+
+    class Meta:
+        db_table = "hardware_status"
+        unique_together = ("checkout_id", "origin", "platform")
```
barbieri: I'd have origin, platform first so indexes reaching only them will work.

also need another index on origin, platform, start_time... since we also query based on those.

backend/kernelCI_app/models.py
```diff
@@ -228,3 +228,73 @@ class Meta:
...
+
+
+class LatestCheckout(models.Model):
+    checkout_id = models.TextField(primary_key=True)
```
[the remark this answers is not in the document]
barbieri: also not likely, but checkout_id is not a PK, the same commit can be in multiple trees...

backend/kernelCI_app/management/commands/helpers/aggregation_helpers.py
```diff
@@ -0,0 +1,93 @@
...
+            checkout.tree_name,
+            checkout.git_repository_url,
+            checkout.git_repository_branch,
+            checkout.start_time,
+        )
+        for checkout in checkouts_instances
+    ]
+
+    with connection.cursor() as cursor:
+        cursor.executemany(
```
MarceloRobert: What is the difference between `executemany` and `bulk_create`? Alternatively, would it be possible to run a single `execute` but sending all the values at once?

backend/kernelCI_app/management/commands/helpers/aggregation_helpers.py
MarceloRobert: nit: it would be useful to have a high-level docstring for the `aggregate_` functions

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+    StatusChoices,
+)
+
+
+def get_hardware_key(origin: str, platform: str, checkout_id: str) -> bytes:
+    return hashlib.sha256(f"{origin}|{platform}|{checkout_id}".encode("utf-8")).digest()
+
+
+def calculate_status_count(status: str) -> tuple[int, int, int]:
+    if status in [StatusChoices.PASS, SimplifiedStatusChoices.PASS]:
```
barbieri: this is creating a list every time... either replace with 2 explicit comparison or use a tuple and declare it as a toplevel constant. Same for the case below below.

Also I'd stick with explicit `SimplifiedStatusChoices` values, if needed then you convert from `StatusChoices` to `SimplifiedStatusChoices` and then you call this function

barbieri: btw the name is also misleading, it's not calculating anything. And if you stick with a single choice you can do a mapping:

```py
simplified_status_choice_to_count = {
    SimplifiedStatusChoices.PASS: (1, 0, 0),
    SimplifiedStatusChoices.FAIL: (0, 1, 0),
    SimplifiedStatusChoices.INCOMPLETE: (0, 0, 1),
}
```

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+            if last_processed_test_id:
+                qs = qs.filter(test_id__gt=last_processed_test_id)
+
+            pending_tests_batch = list(qs[:batch_size])
+
+            if not pending_tests_batch:
+                self.stdout.write("No pending tests found, exiting batch")
+                return 0
+
+            last_processed_test_id = pending_tests_batch[-1].test_id
```
barbieri: it's better to set this after the processing loop ends, or during the loop... so you don't need to -1 here

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+                        data["build_failed"],
+                        data["build_inc"],
+                        data["boot_pass"],
+                        data["boot_failed"],
+                        data["boot_inc"],
+                        data["test_pass"],
+                        data["test_failed"],
+                        data["test_inc"],
+                    )
+                    for data in hardware_status_data.values()
```
barbieri: `values_list("checkout_id", "origin"...)` will return all items as an array in the correct order, much cheaper and no need to loop again

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+    return hashlib.sha256(f"{origin}|{platform}|{checkout_id}".encode("utf-8")).digest()
+
+
+def calculate_status_count(status: str) -> tuple[int, int, int]:
+    if status in [StatusChoices.PASS, SimplifiedStatusChoices.PASS]:
+        return (1, 0, 0)
+    elif status in [StatusChoices.FAIL, SimplifiedStatusChoices.FAIL]:
+        return (0, 1, 0)
+    else:
+        return (0, 0, 1)
```
barbieri: stick with SimplifiedStatusChoices, with that you can create a map which you can use directly as `simplified_status_choices_count[status]`

```py
simplified_status_choices_count = {
    SimplifiedStatusChoices.PASS: (1, 0, 0),
    SimplifiedStatusChoices.FAIL: (0, 1, 0),
    SimplifiedStatusChoices.INCOMPLETE: (0, 0, 1),
}
```

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+        "test_failed": 0,
+        "test_inc": 0,
+    }
+
+
+def _collect_test_contexts(
+    tests_instances: Sequence[PendingTest],
+    builds_by_id: dict[str, Builds],
+    checkouts_by_id: dict[str, Checkouts],
+):
```
barbieri: fully type the return, the returned tuple and its elements

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+def _process_build_status(
+    build,
+    h_key,
+    status_record,
+    existing_processed,
+    processed_builds_in_batch,
+    new_processed_entries,
+):
```
barbieri: fully type parameters and return

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+
+    for test, build, checkout, platform, h_key in contexts:
+        record_key = (checkout.origin, platform, checkout.id)
+
+        if record_key not in hardware_status_data:
+            hardware_status_data[record_key] = _init_status_record(
+                checkout, platform, test.compatible
+            )
+
+        status_record = hardware_status_data[record_key]
```
barbieri:
```py
try:
    status_record = hardware_status_data[record_key]
except KeyError:
    status_record = _init_status_record(checkout, platform, test.compatible)
    hardware_status_data[record_key] = status_record
```

this avoids the double fetch and is more efficient, another shorter but less efficient:

```py
status_record = hardware_status_data.setdefault(record_key, _init_status_record(checkout, platform, test.compatible))
```

it's more inefficient since it would be calling `_init_status_record(checkout, platform, test.compatible)` every time, which creates a dict that will be then discarded...

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+        )
+
+        _process_build_status(
+            build,
+            h_key,
+            status_record,
+            existing_processed,
+            processed_builds_in_batch,
+            new_processed_entries,
+        )
```
barbieri: given that we'll never have to process a build if the test itself wasn't processed (implies the build was already done), it's better to return a boolean from `_process_test_status()`, accumulate `build` and only process it afterwards:

```py
if _process_test_status(...):
   builds_to_process.add(build)
```

where `builds_to_process` is a `set[Build]`

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+    existing_processed,
+    new_processed_entries,
+):
+    """Process test status and update status record if not already processed."""
+    if (
+        h_key,
+        test.test_id,
+        HardwareStatusEntityType.TEST.value,
+    ) in existing_processed:
+        return
```
barbieri: if we move to `set[ProcessedHardwareStatus]`, here would be:

```py
processed = ProcessedHardwareStatus(h_key, test.test_id, HardwareStatusEntityType.TEST.value)
if processed in existing_processed:
    return False
...

new_processed_entries.add(processed)
return True
```

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+
+    if (
+        h_key,
+        build.id,
+        HardwareStatusEntityType.BUILD.value,
+    ) in existing_processed:
+        return
+
+    if (h_key, build.id) in processed_builds_in_batch:
+        return
```
barbieri: ditto, if we go with a `set[ProcessedHardwareStatus]` for both `existing_processed` and `new_processed_entries` we can:

```py
processed = ProcessedHardwareStatus(h_key, build.id, HardwareStatusEntityType.BUILD.value)
if processed in existing_processed:
    return False
if processed in new_processed_entries:
    return False
...

new_processed_entries.add(processed)
return True
```

note it removes the need for `processed_builds_in_batch`

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,394 @@
...
+    build,
+    h_key,
+    status_record,
+    existing_processed,
+    processed_builds_in_batch,
+    new_processed_entries,
+):
+    """Process build status and update status record if not already processed."""
+    if build.id.startswith(MAESTRO_DUMMY_BUILD_PREFIX):
+        return
```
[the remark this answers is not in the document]
gustavobtflores: for this case, we can’t filter earlier, because although we don’t count dummy builds for build status, we do have tests linked to those dummy builds, and those test results are been counted

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
```diff
@@ -0,0 +1,92 @@
...
+        batch_size = options["batch_size"]
+
+        valid_checkout_ids = set(
+            LatestCheckout.objects.values_list("checkout_id", flat=True)
+        )
+
+        orphaned_entries = HardwareStatus.objects.exclude(
+            checkout_id__in=valid_checkout_ids
+        )
+        orphaned_count = orphaned_entries.count()
```
barbieri: this must be inside the transaction

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
```diff
@@ -0,0 +1,92 @@
...
+            )
+            return
+
+        self.stdout.write(
+            f"Found {orphaned_count} HardwareStatus entries "
+            "with no corresponding LatestCheckout."
+        )
+
+        total_deleted = 0
+        while True:
```
barbieri: why?

backend/kernelCI_app/management/commands/helpers/aggregation_helpers.py
MarceloRobert: Please create an issue for adding unit tests for this and the other functions added in this PR just so that we don't forget

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
barbieri: this must be inside the transaction

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
```diff
@@ -0,0 +1,92 @@
...
+            with transaction.atomic():
+                batch_ids = list(
+                    orphaned_entries.values_list("checkout_id", flat=True)[:batch_size]
```
barbieri: the `values_list("checkout_id", flat=True)` you can specify when you create the `orphaned_entries`, here you just slice

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+            status_record = _init_status_record(
+                checkout, test.origin, test.platform, test.compatible
+            )
+            hardware_status_data[record_key] = status_record
+
+        _process_test_status(
+            test, h_key, status_record, existing_processed, new_processed_entries
+        )
+
+        _process_build_status(
```
barbieri: only process build if the test was processed (return boolean in `_process_test_status` and check before this line)

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+        last_processed_test_id = None
+
+        while True:
+            out(
+                f"Starting batch processing "
+                f"(last_processed_test_id={str(last_processed_test_id)[:20]}, "
+                f"batch_size={batch_size})..."
+            )
+
+            qs = PendingTest.objects.order_by("test_id")
```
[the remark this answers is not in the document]
MarceloRobert: at least that's what I see since `test_id` is a TextField, and in `aggregate_tests` you assign `test_id` as `t.id` from a `Tests` object

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+            if not found_builds:
+                out("No builds found for pending tests, skipping batch")
+                continue
+
+            ready_tests = []
+            ready_builds = {}
+            skipped_no_build = 0
+            skipped_no_checkout = 0
+
+            for pt in pending_tests_batch:
```
MarceloRobert: nit:
```suggestion
            for pending_test in pending_tests_batch:
```

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+    """Aggregate hardware status from pending tests, builds, and checkouts."""
+    hardware_status_data = {}
+    new_processed_entries = set()
+
+    contexts, keys_to_check = _collect_test_contexts(tests_instances, builds_by_id)
+
+    if not contexts:
+        return hardware_status_data, new_processed_entries
+
+    existing_processed = _get_existing_processed(keys_to_check)
```
MarceloRobert: just as a comment, I see that if you did `_get_existing_processed(set(c[4] for c in contexts))` you wouldn't need the `keys_to_check` at all, but I don't think it would make a difference tbh

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+                f"Batch processed: {tests_count} tests aggregated, "
+                f"skipped (no build): {skipped_no_build}, "
+                f"skipped (no checkout): {skipped_no_checkout} "
+                f"in {time.time() - t0:.3f}s"
+            )
+
+    def _process_ready_tests(
+        self,
+        ready_tests,
+        ready_builds,
```
MarceloRobert: nit: untyped parameters

backend/kernelCI_app/management/commands/process_pending_aggregations.py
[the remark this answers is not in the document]
gustavobtflores: just to guarantee that we are processing tests always in the same order, without that we could receive data in an arbitrary order

backend/kernelCI_app/views/hardwareViewV2.py
```diff
@@ -0,0 +1,89 @@
...
+
+        try:
+            sanitized_records = self._sanitize_records(hardwares_raw=hardwares_raw)
+            result = HardwareListingResponseV2(hardware=sanitized_records)
+
+            if len(result.hardware) < 1:
+                return create_api_error_response(
+                    error_message=ClientStrings.NO_HARDWARE_FOUND,
+                    status_code=HTTPStatus.OK,
+                )
```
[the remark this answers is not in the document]
gustavobtflores: just followed the current approach on `hardwareView`, but you're right, we can check it earlier

backend/kernelCI_app/management/commands/process_pending_aggregations.py
gustavobtflores: it is a possibility, but i think it makes the code less readable, the first thought that i had was "what is c[4]?"

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,413 @@
...
+
+            pending_build_ids = {pt.build_id for pt in pending_tests_batch}
+            t0 = time.time()
+            found_builds = (
+                Builds.objects.select_related("checkout")
+                .only(
+                    "id",
+                    "checkout_id",
+                    "status",
+                    "checkout__id",
```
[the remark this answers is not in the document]
gustavobtflores: we don't need `checkout_id`, but we do need `checkout__id`. AFAIK `only()` is just to tell Django that we only want these columns, not the other ones, so as we are using `checkout__id` and `checkout__start_time`, we need to specify it here

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,425 @@
...
+        return
+
+    to_process = ProcessedHardwareStatus(
+        hardware_key=h_key,
+        entity_id=build.id,
+        entity_type=HardwareStatusEntityType.BUILD,
+    )
+
+    if to_process in existing_processed:
+        return
```
barbieri: when we move to `hardware_key` to include entity_id/type, then `build_h_key` is already what we'll check in `existing_processed`, no need to create the model instance here.

backend/kernelCI_app/management/commands/process_pending_aggregations.py
barbieri: agreed

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,425 @@
...
+    new_processed_entries: set[ProcessedHardwareStatus] = set()
+
+    contexts, keys_to_check = _collect_test_contexts(tests_instances, builds_by_id)
+
+    if not contexts:
+        return hardware_status_data, new_processed_entries
+
+    existing_processed = _get_existing_processed(keys_to_check)
+
+    for test, build, checkout, h_key in contexts:
```
barbieri: here you will get `test_h_key` and `build_h_key`, then be careful when calling `_process_*_status()`

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,425 @@
...
+                f"skipped (no checkout): {skipped_no_checkout} "
+                f"in {time.time() - t0:.3f}s"
+            )
+
+    def _process_ready_tests(
+        self,
+        ready_tests: Sequence[PendingTest],
+        ready_builds: dict[str, Builds],
+    ) -> int:
+        with transaction.atomic():
```
barbieri: then this transaction is in process_pending_batch and not here

backend/kernelCI_app/models.py
```diff
@@ -228,3 +234,102 @@ class Meta:
...
+        "entity_id",
+        "entity_type",
+    )
+    hardware_key = models.BinaryField(
+        max_length=32
+    )  # this holds a sha256, thus digest_size = 32 bytes
+    entity_id = models.TextField()
+    entity_type = models.CharField(
+        max_length=1, choices=HardwareStatusEntityType.choices
+    )
```
barbieri: we should include entity_id/type inside the `hardware_key`, currently the lookup for `hardware_key` will return all entities ever processed for that hardware, as the system grows and we never delete `ProcessedHardwareStatus` entries, the set will be large and this will hurt performance (and memory usage).

maybe we should keep a `checkout_id` field from a cron job we should remove those that are not related to the latest checkouts, same as we did with `delete_unused_hardware_...` (or even do it from inside that job)

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -0,0 +1,425 @@
...
+                        time.sleep(interval)
+            except KeyboardInterrupt:
+                out("Stopping pending aggregation processor...")
+        else:
+            self.process_pending_batch(batch_size)
+
+    def process_pending_batch(self, batch_size: int) -> int:
+        last_processed_test_id = None
+
+        while True:
```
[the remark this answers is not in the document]
gustavobtflores: i think it is safer to keep the transaction inside the loop, with that we don't lose time reprocessing the previous batches if some of them fails to commit

backend/kernelCI_app/management/commands/process_pending_aggregations.py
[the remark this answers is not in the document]
MarceloRobert: Ah right, I was thinking that it was not necessary to fetch both at the same time but it wasn't too clear on my message

MarceloRobert (approved):
Looks good to me, I made some tests and it is working as intended. We'll check on how to integrate this with the existing dashboard. There's just the comment on the schema and endpoint example left, but I'll pre-approve

PR #1618, opened 2025-11-17
Feat: treeDetails lab filter

### LINKED ISSUE #1616: Add lab filtering to treeDetails
Labs are present in builds and tests. It would be really expensive to add lab filtering to the listing pages, and the specific issueDetails and buildDetails pages currently don't have filters. So the lab filter should be added to treeDetails and hardwareDetails.

- Add lab summary per tab
- Add lab filter logic
- Add lab card (only showing when there is more than one lab in the tab)
- Add filter interaction in the frontend

MarceloRobert:
Adds lab filter feature for treeDetails

## Changes
- Returns build, boot and test labs in summary and filter fields for treeDetails and treeDetailsSummary views
- Adds filter logic for labs
- Refactors origin card to be reusable as any filter card
- Adds labs cards to treeDetails tabs

## How to test
Go to treeDetails, check the lab cards, click on their value and see the filter working.
Check if the filter is reflecting in the commit graph as well.
The hardware details pages don't have this filter yet, but should be working just like before.

Closes #1616

### DIFF
--- a/docs/IntegrationTests.md
+++ b/docs/IntegrationTests.md
@@ -252,7 +252,7 @@ docker compose -f docker-compose.test.yml run --rm test_backend python manage.py
 
 **Step 4: Start test backend**
 ```bash
-docker compose -f docker-compose.test.yml up test_db redis -d
+docker compose -f docker-compose.test.yml up --build -d
 ```
 
 **Step 5: Run integration tests**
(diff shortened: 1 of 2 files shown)


dashboard/src/components/Cards/FilterCard.tsx
```diff
@@ -13,35 +13,38 @@ import type {
   TFilter,
   TFilterObjectsKeys,
 } from '@/types/general';
+import type { MessagesKey } from '@/locales/messages';
 
 interface IOriginsCard {
```
gustavobtflores: we could rename it

```suggestion
interface IFilterCard {
```

PR #1628, opened 2025-11-21
Fix: correct cache interval

### LINKED ISSUE #1555: Cip trees show for 29 days but not 30 days
When filtering by 29 days:
Image

When filtering by 30 days:
Image

MarceloRobert:
One possibility for cache problems is if the cache only exists for a very old period, and the same checkouts in the cache also appear in the query without cache (which is always made for the latest 3 days). This means that checkouts older than the 3-day limit and outside of the cache will not appear.

So the solution is not to query without cache from now only up to 3 days, but from now up to the earliest cached checkout, so that no gap is made.

This problem happens if the cache is not working though, since it would mean that it only exists way before the 3-day limit, so a health-check is needed for our cache file.

## Changes

- Updated the interval for treeListing query without cache

## How to test

Since it involves the sqlite cache, there are some steps: copy some "old" data to the local database (between 7 and 6 days ago for example), cache it by triggering the `update_checkout_cache` function (you can place it in some other endpoint for easy triggering), then copy more data (between 4 and 3 days for example).
In the old behavior, you wouldn't see the 4-3 data when using an interval of 7 days, but now you'll be able to see it.

Closes #1555

### DIFF
--- a/backend/kernelCI_app/views/treeView.py
+++ b/backend/kernelCI_app/views/treeView.py
@@ -18,6 +18,7 @@ from pydantic import ValidationError
 from kernelCI_cache.constants import UNSTABLE_CHECKOUT_THRESHOLD
 from kernelCI_cache.queries.tree import get_cached_tree_listing_data
 from kernelCI_app.constants.localization import ClientStrings
+from django.utils.timezone import now, make_aware
 
 
 class TreeView(APIView):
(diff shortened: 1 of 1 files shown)


backend/kernelCI_app/views/treeView.py
```diff
@@ -45,11 +46,19 @@ def get(self, request: HttpRequest) -> Response:
...
+        if has_cache:
+            # Query kcidb from now up to the earliest cache entry so that we can
+            # avoid a cache gap between the last cache entry and UNSTABLE_CHECKOUT_THRESHOLD
+            earliest_cache_time = min(
+                checkout["start_time"] for checkout in cached_checkouts
+            )
+            if earliest_cache_time.tzinfo is None:
+                earliest_cache_time = make_aware(earliest_cache_time)
+            days_to_earliest = (now() - earliest_cache_time).days
+            kcidb_interval = max(days_to_earliest, UNSTABLE_CHECKOUT_THRESHOLD)
```
gustavobtflores: i think that with this approach in some cases we could get more data than actually requested at start, like if cache has data from 10 days ago and we request 5, we would get:

```
days_to_earliest=10
kcidb_interval=10
```

and then fetch data for more than the 5 days that we initially queried, maybe we could filter after?

Issue #1461, opened 2025-09-03
User cannot change date range if issues query fails


LucasSantos27:
Currently, when the issues query throws an error, the user loses access to the date range selector and is forced to manually update the URL parameter `i` to change the number of days.

For example:

> https://dashboard.kernelci.org/issues?i=9

Where `i` represents the number of days.

### Current Behavior
<img width="1801" height="714" alt="Image" src="https://github.com/user-attachments/assets/eab68d8d-c3aa-40d3-8894-5c1ee9750a63" />

### Expected Behavior
<img width="1801" height="714" alt="Image" src="https://github.com/user-attachments/assets/fac5eb38-65ef-4706-b7e3-61ae5e40509a" />

- The table should remain visible even if the query fails.
- Instead of displaying "No Results Were Found", the user should see an appropriate error message.
- The user should still be able to change the date range from the UI without needing to manually edit the URL.

gustavobtflores:
Closed by #1637

PR #1637, opened 2025-11-26
fix: show hardware/tree table even when no data is available

### LINKED ISSUE #57: Add better error handling for Tree Listing
A very basic error handling was added to Tree Listing
a new better one would be nice

More details in:
https://github.com/kernelci/dashboard/pull/56#discussion_r1691845346

### LINKED ISSUE #1636: Dashboard hides filters when there is no data for the specified period
If we access trees/hardware view and it has no data for the last X days, it hides the table filters (including the date range one) making it impossible for the user to filter by other date ranges

Image

gustavobtflores:
Display tables even when the API returns a "not found" error (HTTP 200 with an error message), showing an empty table instead of an error component

## Problem Solved

A "Not Found" response from the backend caused the UI to hide the whole table, blocking users from changing the date range and other filters

## Impact

Improves UX by showing empty tables for "Not Found" cases while preserving error handling for real failures. The table remains visible and filterable even when no results are found

## Visual Reference

Before:

image

After:

image

Closes #1636, closes #57

### DIFF
--- a/dashboard/src/components/IssueTable/IssueTable.tsx
+++ b/dashboard/src/components/IssueTable/IssueTable.tsx
@@ -20,6 +20,7 @@ import type { LinkProps } from '@tanstack/react-router';
 import { useNavigate, useSearch } from '@tanstack/react-router';
 
 import { useCallback, useMemo, useState } from 'react';
+import type { UseQueryResult } from '@tanstack/react-query';
 
 import { FormattedMessage } from 'react-intl';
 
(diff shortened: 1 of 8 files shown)


MarceloRobert (commented):
Why does this change not touch IssueListing page if it can also be empty?

MarceloRobert (commented):
I think you can put the "empty" labels inside the tables as well, such as `treeListing.notFound` instead of the generic `global.noResults`

MarceloRobert (commented):
The error for the treeListing (not fast) is resulting in an infinite "loading" in the columns. I think that showing a `-` in the columns and adding a "Couldn't load status counts" message at the top of the table would solve the issue that we were discussing

MarceloRobert (approved):
Great, I didn't think we would have to kind of ignore the non-fast error but it makes sense. Maybe in the future we might show the error message but for now it looks good already

dede999:
It also closes #1461

PR #1638, opened 2025-11-27
Feat: change ingester data insertion policy

### LINKED ISSUE #1552: Verify if django-ingester is respecting kcidb limitations on fields update
In old kcidb ingestion we had certain limitations imposed, such as to not change checkout/build/test fields after they are set (we cannot update/delete non-null values).

Location(approx): kcidb/db/sql/schema.py:264-286
The INSERT statement uses ON CONFLICT ... DO UPDATE SET with COALESCE functions to preserve existing values:

ON CONFLICT (primary_key_columns) DO UPDATE SET
      column_name = COALESCE(table.column_name, excluded.column_name)

The logic works as follows:

  1. Lines 264-269: Identifies primary key columns for the conflict resolution
  2. Lines 270-286: For non-primary-key columns, it uses COALESCE to decide which value to keep:
    - If prio_db is True: COALESCE(table.column, excluded.column) - keeps existing DB values if not NULL
    - If prio_db is False: COALESCE(excluded.column, table.column) - prefers new values if not NULL
    - For columns with conflict_func (like timestamps): Uses a specific function (e.g., GREATEST) to resolve conflicts
 (still need to check where and how prio_db used)

MarceloRobert:
## Description
Changes the insertion policy of the ingester from "never update" to "only update null fields". This follows the policy adopted by kcidb, it always prioritizes the incoming data in the following logic (considering existing / incoming):
- null + new data = new data
- existing data + new data = existing data
- existing data + null = existing data

## Changes
- Added a new command to generate the SQL insertion query for any model
  - This command generates a python file with a dictionary relating a table with its insertion query and which fields should be updated
- Used SQL queries on `consume_buffer` instead of Django's `bulk_create`
- Updated unit tests

## How to test
Start the ingester and insert some data to it referencing the same object, check if the policy is respected.
Example realistic data for testing can be found [here](https://drive.google.com/file/d/1enFNW3Z3qH1QC7J0aXAdtlzDgj9lZY-R/view?usp=sharing), but other tests are encouraged.

## Future changes
2. Adding automated integration tests so that we can guarantee that this policy is kept regardless of changes in the ingester is also useful.

Closes #1552


backend/kernelCI_app/constants/ingester.py
```diff
@@ -48,3 +48,9 @@
 except (ValueError, TypeError):
     logger.warning("Invalid INGEST_QUEUE_MAXSIZE, using default 5000")
     INGEST_QUEUE_MAXSIZE = 5000
+
+
+PRIO_DB = is_boolean_or_string_true(os.environ.get("PRIO_DB", "True"))
```
tales-aparecida: I think adding PRIO_DB is bringing a lot of complexity to this pull request. If we decide to change the policy again in the future, we can send another PR.
I believe we can assume you aren't going to need it, though.

backend/kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py
```diff
@@ -206,6 +206,86 @@ def test_prepare_file_data_file_error(self, mock_file_open, mock_logger):
         mock_file_open.assert_called_once()
 
 
+class TestGenerateInsertQuery:
```
tales-aparecida: wow, I never read this file before, it's quite high-level and scary. 

Do you think you could add a test case achieving something close to:

```
input = json.loads(Path("fixtures/inputs/kcidb.json").read_text())
expect_data = json.loads(Path("fixtures/outputs/tables.json").read_text())

with self.assertNumQueries(5):
    ingest(input)

result_data = {
  "checkouts": Checkout.objects.all().values(),
  "builds": Build.objects.all().values(),
  "tests": Test.objects.all().values(),
  "issues": Issue.objects.all().values(),
  "incidents": Incident.objects.all().values(),
}
self.assertEqual(result_data, expected_data)
```

MarceloRobert: I changed it so that the queries are now generated with a command, so there's not a `_generate_model_insert_query()` anymore

tales-aparecida: It's much better now!

Issue #930, opened 2025-02-14
update deprecated type aliases


murilx:
We've been using `typing.Dict`, `typing.List`, `typing.Set`, etc. but they are deprecated since python 3.9 according to the documentation

https://docs.python.org/3/library/typing.html#aliases-to-built-in-types

PR #1671, opened 2025-12-11
Feat: hardware details lab filter

### LINKED ISSUE #1164: Bug: HardwareDetails is not returning test summary
When checking the summary of a hardware, the data is returned correctly, but when checking the hardware details, the summary is returned but without the tests and boots data.
We have the "hardware summary" endpoint, which only returns the basic information for each tab, and the "hardware full" endpoint, which should return more information as well as the same data as the summary.

**Problem**: when accessing the "hardware full" endpoint, the test summary data is empty, every dict value is 0.

**Expected behavior**: the test summary data should be the same as the data in the "hardware summary" endpoint

### LINKED ISSUE #1617: Add lab filtering in hardwareDetails
Labs are present in builds and tests. It would be really expensive to add lab filtering to the listing pages, and the specific issueDetails and buildDetails pages currently don't have filters. So the lab filter should be added to treeDetails and hardwareDetails.

- Add lab summary per tab
- Add lab filter logic
- Add lab card (only showing when there is more than one lab in the tab)
- Add filter interaction in the frontend

MarceloRobert:
Adds filter functionality for the lab data in hardware details, also shows the cards for labs in the frontend page.

## Changes
- Connects the lab data from hardware endpoints to the existing lab returns and filters
- Adds the cards to the hardware tabs
- Adds the lab filter sections to the hardware details filter modal
- Also refactors a function that had an unused argument

## How to test
Check some harware pages and filter by the lab through the card and through the modal
[This localhost hardware](http://localhost:5173/hardware/bcm2711-rpi-4-b?et=1765470600&p=t&st=1765038600) has a couple of labs for boots and tests (and [this is the staging equivalent](https://staging.dashboard.kernelci.org:9000/hardware/bcm2711-rpi-4-b?et=1765470600&p=t&st=1765038600))

Closes #1164 
Closes #1617 

Example video:
https://github.com/user-attachments/assets/740009be-c7bb-4034-8190-6ea90963cdc1

### DIFF
--- a/backend/kernelCI_app/helpers/hardwareDetails.py
+++ b/backend/kernelCI_app/helpers/hardwareDetails.py
@@ -215,6 +215,7 @@ def generate_test_dict() -> dict[str, Any]:
         "configs": defaultdict(lambda: defaultdict(int)),
         "issues": {},
         "failedWithUnknownIssues": 0,
+        "labSummary": defaultdict(int),
     }
 
 
@@ -312,10 +313,10 @@ def handle_test_history(
 
 def handle_test_summary(
     *,
-    record: Dict,
+    record: dict,
     task: TestSummary,
-    issue_dict: Dict,
-    processed_archs: Dict[str, TestArchSummaryItem],
+    issue_dict: dict,
+    processed_archs: dict[str, TestArchSummaryItem],
 ) -> None:
     status = record["status"]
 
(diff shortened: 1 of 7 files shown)


backend/kernelCI_app/helpers/hardwareDetails.py
```diff
@@ -312,10 +313,10 @@ def handle_test_history(
 
 def handle_test_summary(
     *,
-    record: Dict,
+    record: dict,
     task: TestSummary,
-    issue_dict: Dict,
-    processed_archs: Dict[str, TestArchSummaryItem],
```
gustavobtflores: why exactly the change from Dict -> dict?

MarceloRobert: So that we slowly address https://github.com/kernelci/dashboard/issues/930

PR #1672, opened 2025-12-11
refactor: modify ingester to use processes in a 'fan-out' pattern

### LINKED ISSUE #1673: Parallelize the ingester by turning each worker into a full ingestion pipeline
Today multiple workers parse/validate files, but all inserts are funneled through a single process, which becomes a bottleneck. This limits throughput and leaves workers idle during busy ingestion cycles.

We should evaluate shifting to a model where each worker performs the full pipeline, parse, validate, and insert, using its own database connection. This removes the single-process bottleneck and allows inserts to happen in parallel.

gustavobtflores:
This PR reviews the ingester architecture to use processes both for file processing and database insertion, allowing faster processing using multiple database connections doing batch inserts at the same time.

## How to test

1. Get some submissions files and a trees.yaml file
2. Run the monitor_submissions command `poetry run python manage.py monitor_submissions [...options]`
3. Check if the submissions are being moved to the correct directory and data inserted correctly in the database

## Performance

Before: 
```
[2026-01-07 22:16:28] Ingest cycle: 48000 files (ok=48000, fail=0) in 664.84s | 72.20 files/s | 9002.68 MB processed (13.54 MB/s)
```

After:
```
[2026-01-08 13:33:54] Ingest cycle: 48000 files (ok=48000, fail=0) in 411.70s | 116.59 files/s | 9002.68 MB processed (21.87 MB/s)
```

Closes #1673

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -1,16 +1,16 @@
 import multiprocessing
-from multiprocessing.synchronize import Event as EventClass
+from multiprocessing.sharedctypes import Synchronized
+from multiprocessing.synchronize import Lock as ProcessLock
 from os import DirEntry
-from concurrent.futures import ProcessPoolExecutor, as_completed
 import json
 import logging
 import os
-from queue import Queue, Empty
+from queue import Queue
 from typing_extensions import Literal
 from kernelCI_app.constants.ingester import (
     CONVERT_LOG_EXCERPT,
     INGEST_BATCH_SIZE,
-    INGEST_FLUSH_TIMEOUT_SEC,
+    INGEST_FILES_BATCH_SIZE,
     INGEST_QUEUE_MAXSIZE,
     INGESTER_GRAFANA_LABEL,
     VERBOSE,
@@ -91,7 +91,7 @@ def standardize_tree_names(
 
 def prepare_file_data(
     file: SubmissionFileMetadata, tree_names: dict[str, str]
-) -> tuple[Optional[dict[str, Any]], dict[str, Any]]:
+) -> tuple[Optional[dict[str, Any]], Optional[dict[str, Any]]]:
     """
     Prepare file data: read, extract log excerpts, standardize tree names, validate.
     This function does everything except the actual database load.
@@ -173,8 +173,10 @@ def flush_buffers(
     tests_buf: list[Tests],
     incidents_buf: list[Incidents],
     buffer_files: set[tuple[str, str]],
-    archive_dir: str,
-    pending_retry_dir: str,
+    dirs: dict[INGESTER_DIRS, str],
+    stat_ok: Synchronized,
+    stat_fail: Synchronized,
+    counter_lock: ProcessLock,
 ) -> None:
     """
     Consumes the list of objects and tries to insert them into the database.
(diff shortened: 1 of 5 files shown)


backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -371,60 +253,218 @@ def buffered_total() -> int:
...
-    failed_dir: str,
-    db_queue: Queue,
-) -> bool:
-    """
-    Process a single file in a process, then queue it for database insertion.
+    dirs: dict[INGESTER_DIRS, str],
+    processed,
+    stat_ok,
+    stat_fail,
+    counter_lock,
```
MarceloRobert: Can you type those parameters? I see that you can import the types from `multiprocessing.synchronize` since the multiprocessing classes are defined there.

https://stackoverflow.com/questions/71988089/python-typing-for-multiprocessing-value

https://stackoverflow.com/questions/70424427/using-python-multiprocessing-lock-as-an-argument-type-in-mypy

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -371,60 +253,218 @@ def buffered_total() -> int:
...
+    Ingest submissions in parallel using child processes for I/O and database operations.
     """
+    cycle_start = time.time()
     total_bytes = 0
-    for f in json_files:
+    total_files_count = len(json_files)
+
+    process_queue: multiprocessing.Queue[Optional[list[SubmissionFileMetadata]]] = (
+        multiprocessing.Queue(maxsize=INGEST_QUEUE_MAXSIZE)
+    )
```
MarceloRobert: the `cycle_start`, `total_files_count` and `process_queue` don't need to be moved up here, they could be defined after the first `out()` is called

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -433,130 +473,56 @@ def ingest_submissions_parallel(  # noqa: C901 - orchestrator with IO + threadin
...
-        files_per_sec = total_files / elapsed if elapsed > 0 else 0.0
-        mb = total_bytes / (1024 * 1024)
-        mb_per_sec = mb / elapsed if elapsed > 0 else 0.0
-        msg = (
-            "Ingest cycle: %d files (ok=%d, fail=%d) in %.2fs | "
-            "%.2f files/s | %.2f MB processed (%.2f MB/s)"
-            % (
-                total_files,
+    writers = []
+    for _ in range(max_workers):
```
MarceloRobert: why remove the `try-except` for keyboard interrupt?

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
gustavobtflores: i'm using these variables in the loop right below

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -371,60 +253,218 @@ def buffered_total() -> int:
...
+            dirs=dirs,
+            stat_ok=stat_ok,
+            stat_fail=stat_fail,
+            counter_lock=counter_lock,
         )
-    )
-    FILES_INGESTER_COUNTER.labels(INGESTER_GRAFANA_LABEL).inc()
 
-    return True
+    if any(len(instances_dict[table]) for table in instances_dict):  # type: ignore
```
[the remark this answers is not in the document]
MarceloRobert: doesn't complain for me, is Mypy an extension that you have enabled?

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
MarceloRobert: I mean that in the old code they were being declared right after the `out` for `Spool status...`, but you moved them up a bit

PR #1696, opened 2026-01-13
fix: modify width of main content to avoid overflowing

### LINKED ISSUE #1700: Fix horizontal scrolling on smaller screens
Right now when seeing the dashboard on smaller screens the viewport gets a horizontal scroll, which causes some tables to not be fully visible without scrolling

gustavobtflores:
## Description

The main content of the pages in the dashboard weren't considering the size of the sidenav menu, what caused the main content to overflow on smaller screens, this PR modifies the size of the main content to consider the sidenav menu width avoiding overflow

## Visual reference

References below in ~800px screensize

### Before

Screenshot from 2026-01-13 17-13-11

### After

Screenshot from 2026-01-13 17-11-51

Closes #1700

### DIFF
--- a/dashboard/src/routes/_main/route.tsx
+++ b/dashboard/src/routes/_main/route.tsx
@@ -20,7 +20,7 @@ const RouteComponent = (): JSX.Element => {
       <div className="h-full w-full">
         <div className="flex w-full flex-row">
           <SideMenu />
-          <main className="flex w-full flex-col">
+          <main className="flex min-w-0 flex-1 flex-col">
             <TopBar />
             <div className="bg-light-gray h-full w-full px-16 pt-24">
               <Outlet />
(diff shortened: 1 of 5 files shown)


dashboard/src/routes/_main/route.tsx
```diff
@@ -20,7 +20,7 @@ const RouteComponent = (): JSX.Element => {
       <div className="h-full w-full">
         <div className="flex w-full flex-row">
           <SideMenu />
-          <main className="flex w-full flex-col">
+          <main className="flex w-[calc(100%-224px)] flex-col">
```
MarceloRobert: I don't believe that using a fixed size is the go-to here. I made some experiments and asked copilot for suggestions and we arrived at this, which seems to work just like your change:

```suggestion
          <main className="flex min-w-0 flex-1 flex-col">
```

Thanks for finding out which file was the culprit though, that helped a lot

MarceloRobert:
Also, does this PR close any issue?

MarceloRobert (approved):
LGTM, the overflow was fixed and the buttons didn't change much, that's good

PR #1723, opened 2026-01-30
Chore: update composer setup

### LINKED ISSUE #1721: Update composer setup
The docker compose is outdated with the google cloud stuff and could also be used to start the ingester separately.

Issue from #1548

MarceloRobert:
## Changes
- Fixes default for trees_file when it was not passed to the ingester command
- Adds an ingester container
- Adds a container for `process_pending_aggregation`
- Fixes signal handling for both commands when running on docker
- Changes backend entrypoint so that these new command containers don't start cronjobs
- Removes google cloud-sql references

Since there are profiles in the new containers, by running `docker compose up` they won't be started by default. But if you run `docker compose --profile with-commands up` you'll run *all containers* including the new services.

I wasn't able to connect to the production database through docker, so the instructions reflect that.

Closes #1721


nuclearcat (approved):
looks good

PR #1741, opened 2026-02-11
Feat: add responsivity to searchbar

### LINKED ISSUE #1416: Fix search bar placement on small screens
For the listing pages, the search bar might be too large. We could replace it with just an icon or make it much smaller and move it down

Image

MarceloRobert:
## Changes
- Removes the searchbar from all listing pages and moves the control to a specific component used in TopBar
- Improves usability of customDialog for use in the searchBox
- Adds media query and replaces box with just a search icon, but with the same behavior

## How to test
Go to any listing page, use the search box and also check the responsivity of the component. The search behavior should stil be the same.

## Visual examples
Large screen:
image

Medium-large screens:
image

Medium screens:
image

Medium-small screens:
image

Small screens:
image

When clicking on the search icon:
image

Closes #1416


gustavobtflores (approved):
LGTM

PR #1743, opened 2026-02-11
feat: introduce hardware listing versioning and update routing

gustavobtflores:
## Description

Add a feature flag to switch the Hardware listing between v1 and v2, and refactor route/search handling to be type-safe and avoid hardcoded route strings

## Changes

- Add `VITE_FEATURE_FLAG_HARDWARE_LISTING_VERSION` and document it in .env.example and README.md
- Introduce a typed HardwareListingRoutesMap and pass route/search sources through Hardware pages and API hooks
- Update hardware routes so /_main/hardware renders v1 or v2 based on the feature flag (while keeping /_main/hardware/v1 working)

## How to test

1. Set VITE_FEATURE_FLAG_HARDWARE_LISTING_VERSION=v1, start the dashboard, open /hardware and confirm you see the v1 listing (and no “old page” banner).
2. Set VITE_FEATURE_FLAG_HARDWARE_LISTING_VERSION=v2, restart, open /hardware and confirm you see the v2 listing (and no “new page” banner).
3. Open /hardware/v1 and confirm it still renders the v1 listing and URL/search params (origin, interval, search) behave as before.
4. Open /hardware/v2 and confirm it renders the v2 listing and URL/search params (origin, interval, search) behave as expected.

### DIFF
--- a/dashboard/src/pages/Hardware/HardwareTable.tsx
+++ b/dashboard/src/pages/Hardware/HardwareTable.tsx
@@ -69,7 +69,7 @@ interface IHardwareTable {
   navigateFrom: HardwareListingRoutes;
 }
 
-type HardwareListingRoutes = '/hardware' | '/hardware/v1';
+type HardwareListingRoutes = '/hardware' | '/hardware/v1' | '/hardware/v2';
 
 const getLinkProps = (
   row: Row<HardwareItem>,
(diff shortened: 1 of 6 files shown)


dashboard/src/pages/Hardware/HardwareV2.tsx
```diff
@@ -50,12 +56,17 @@ export const HardwareV2 = (): JSX.Element => {
...
-      <NewPageBanner
-        pageNameId="hardwareListing.bannerTitle"
-        pageRoute="/hardware/v1"
-      />
+      {hardwareListingVersion !== 'v2' && (
+        <NewPageBanner
+          pageNameId="hardwareListing.bannerTitle"
+          pageRoute="/hardware/v1"
+        />
+      )}
```
MarceloRobert: This doesn't really show the banner because if you are at v1 then you can't currently see this page, and if you are at v2 then you disable the banner. I think that for now we should always show this banner. On treeListing there's the same mistake actually...

Issue #1738, opened 2026-02-09
New 3‑level hardware summary layout not applied


ximiali:
As the title indicates, we noticed the new 3-level layout format for the hardware summary is not taking effect; the original format is still being used. Since the PR has already been merged, is there an error or something missing?

Related: #1726

ximiali:
Thanks for the clarification, that makes sense.

ximiali:
Sure, I have raised a PR to CC the public mailing list on hardware status emails. It will send hardware status report on Mondays.

PR #1745, opened 2026-02-13
feat(notifications): CC public mailing list on hardware status emails

ximiali:
As suggested by @MarceloRobert, this PR makes the hardware status command send an email copy to the public mailing list for confirmation.
Related: #1738


MarceloRobert (approved):
Thanks!

PR #1751, opened 2026-02-23
Feature: new deploy setup

MarceloRobert:
Major changes for the deploy setup


docker-compose.yml
```diff
@@ -13,48 +13,34 @@ secrets:
...
             - backend-data:${BACKEND_VOLUME_DIR:-/volume_data}
-        restart: always
-        container_name: dashboard_backend_service
+        env_file:
+            - .env.backend
         networks:
             - private
             - public
+        restart: always
+        image: ghcr.io/${IMAGE_NAME:-local}/backend:latest
```
gustavobtflores: it should really be "local" the default here? and IMAGE_NAME seems incorrect too, it wouldn't be the user/organization?

PR #1770, opened 2026-03-02
feat: split the "full" tree details data fetch into three separate queries

gustavobtflores:
## Description

Split the "full" tree details data fetch into three separate, tab-specific queries (builds, boots, tests). Previously, a single heavy query fetched all builds, boots, and tests data at once after the summary loaded. Now each tab triggers only its own query when the user navigates to it, reducing unnecessary data transfer and backend load.

## Changes

- Added get_tree_data() query function in queries/tree.py that accepts a data_type literal ("builds", "boots", "tests") and builds the appropriate SQL join (filtering test paths for boots/tests, or omitting test columns entirely for builds)
- Added get_current_row_data_builds() helper in treeDetails.py that maps build-only query rows while stubbing out test fields so existing filter/aggregation logic remains unchanged
- Updated treeDetailsBuildsView.py, treeDetailsBootsView.py, and treeDetailsTestsView.py to use the new get_tree_data() and get_current_row_data_builds() instead of the generic full-data equivalents
- Replaced the single full query in useTreeDetailsLazyLoadQuery with three separate builds, boots, and tests queries, each gated behind both summary availability and the currently active tab
- Updated BuildTab, BootsTab, TestsTab, and TreeDetails components to consume the new per-tab query results

## How to test

1. Start the full stack (docker compose up --build -d)
2. Navigate to a tree details page and open the Builds tab, verify builds load correctly and no boots/tests requests fire
3. Switch to the Boots tab, verify only the boots query fires
4. Switch to the Tests tab, verify only the tests query fires
5. Confirm filters and issue categorization still work as expected on each tab

### DIFF
--- a/dashboard/src/hooks/useTreeDetailsLazyLoadQuery.ts
+++ b/dashboard/src/hooks/useTreeDetailsLazyLoadQuery.ts
@@ -3,12 +3,22 @@ import type { UseQueryResult } from '@tanstack/react-query';
 import type { UseTreeDetailsWithoutVariant } from '@/api/treeDetails';
 import { useTreeDetails } from '@/api/treeDetails';
 import type {
-  TreeDetailsFullData,
+  TreeDetailsBuilds,
+  TreeDetailsBoots,
+  TreeDetailsTests,
   TreeDetailsSummary,
+  PossibleTabs,
 } from '@/types/tree/TreeDetails';
 import type { QuerySelectorStatus } from '@/components/QuerySwitcher/QuerySwitcher';
 import { useIssueExtraDetails } from '@/api/issueExtras';
 import type { IssueExtraDetailsResponse } from '@/types/issueExtras';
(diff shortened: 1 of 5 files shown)


backend/kernelCI_app/views/treeDetailsTestsView.py
```diff
@@ -97,14 +98,6 @@ def get(
                 status_code=HTTPStatus.OK,
             )
 
-        if len(rows) == 1:
-            row_data = get_current_row_data(current_row=rows[0])
-            if row_data["test_id"] is None:
-                return create_api_error_response(
-                    error_message=ClientStrings.TREE_TESTS_NO_RESULTS,
```
MarceloRobert: just a nit though: this string won't be used anywhere else if you remove this condition, so you can remove it from the clientstrings too

PR #1788, opened 2026-03-10
Fix: cronjob run and log

MarceloRobert:
## Changes
- Makes cronjob logs go to the docker output, making it easier to debug.
- Also changes the command from 'crond start' to just 'crond' so that it works best with busybox.

## How to test
You must have data in your local database so that you can run the cronjobs within the backend container. Having the data, edit the cronjobs so that they run in a smaller interval, then check if their logs are going to the container logs.
The notification and other commands behavior stay the same.

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -128,6 +128,10 @@ SKIP_CRONJOBS = is_boolean_or_string_true(os.environ.get("SKIP_CRONJOBS", False)
 if SKIP_CRONJOBS:
     CRONJOBS = []
 else:
+    # Make cron task output visible in `docker compose logs backend`.
+    CRONTAB_COMMAND_SUFFIX = os.environ.get(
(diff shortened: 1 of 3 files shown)


backend/kernelCI/settings.py
```diff
@@ -128,6 +128,10 @@ def get_json_env_var(name, default):
 if SKIP_CRONJOBS:
     CRONJOBS = []
 else:
+    # Make cron task output visible in `docker compose logs backend`.
+    CRONTAB_COMMAND_SUFFIX = os.environ.get(
+        "CRONTAB_COMMAND_SUFFIX", ">> /proc/1/fd/1 2>&1"
+    )
```
AmadeusK525: Do you need this if you already set it in the `backend_entrypoint.sh`?

AmadeusK525 (approved):

gustavobtflores (approved):
Looks good

PR #1796, opened 2026-03-12
fix: Fix/treedetails smallscreen

### LINKED ISSUE #1422: Fix TreeDetails tabs visualization on small screens
They don't have enough space to be side-by-side. We should think of a better way to display them

Image

alanpeixinho:
## Description
* Includes horizontal scroll for tabs that do not fit the screen.

## Changes
* Includes a CSS property to horizontal auto scroll only if needed.

## How to test
* Run application locally.
* Open the browser using a small screen (360x780), and assess that the tabs do not overflow screen horizontal size.

| Before| After |
| --- | --- |
| before_scroll | tabs_scroll |

Closes #1422

### DIFF
--- a/dashboard/eslint.config.mjs
+++ b/dashboard/eslint.config.mjs
@@ -20,7 +20,7 @@ const compat = new FlatCompat({
 });
 
 export default [{
-    ignores: ["**/.eslintrc.cjs", "**/*.config.*js", "src/components/ui/*", ".storybook/**", "dist/*"],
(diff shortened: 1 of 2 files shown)


WilsonNet (approved):

MarceloRobert (changes requested):
It's not really working for large numbers, and in fact it stopped having a reactive width even for the desktop use

Image

Image

MarceloRobert (approved):

PR #1799, opened 2026-03-12
Fix: checkout cache cronjob

MarceloRobert:
Fixes cronjob that updates checkout cache with type/empty checking and adds debug logs

## Changes
- Type checks/fixes on the checkout cache task and query used by it
- Added logs for that task since it is used in a cronjob

## How to test
You can add this task somewhere to trigger it manually (like inside a view for example) or you can change the interval that it runs on in the cron job (in settings.py, you can use the interval `*/2 * * * *` for example to run every 2 minutes) and run it locally or within docker.

This might fix the problem that we are running on production, or at least it will give some more insight into where the sqlite file is being searched

### DIFF
--- a/backend/kernelCI_app/queries/tree.py
+++ b/backend/kernelCI_app/queries/tree.py
@@ -239,7 +239,10 @@ def get_tree_listing_fast(
     return list(checkouts)
 
 
-def get_tree_listing_data_by_checkout_id(*, checkout_ids: list[str]):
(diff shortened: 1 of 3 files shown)


dede999 (approved):
This PR does not seem to include a new feature, but improves codebase quality by adding return types, readability improvements, and log messages for already existing methods, seems LGTM to me

alanpeixinho (commented):
LGTM

alanpeixinho (approved):

gustavobtflores (approved):
LGTM

PR #1802, opened 2026-03-13
fix: use single columns on smaller screens

### LINKED ISSUE #1432: Change details pages infor to be single column on small screens
For big screens it is obvious that we should use two columns, but it becomes too tight for small screens. We should change it so that for small screens the info is shown only on a single column

dede999:
Fixes layout responsiveness for smaller viewpoints by setting the grid size for default, medium and/or large screens

## Changes

- On `BuildDetails` component, I have wrapped the `DetailsInfoCard` with grid classes changing on large views, as I thought it was okay to have only one column on small and medium screens
- On `Subsection` component, I have made it change the grid size on medium screens as sub sets generally represent more granular data

## How to test
- Visit `build/$buildId` page, inspect it and check different viewpoints to look for points of improval
- This page can be reached by checking a hardware details page and then you check on builds details

## Related issue

- Closes #1432

### DIFF
--- a/dashboard/src/components/BuildDetails/BuildDetails.tsx
+++ b/dashboard/src/components/BuildDetails/BuildDetails.tsx
@@ -159,7 +159,7 @@ const BuildDetails = ({
             infos: [
               {
                 children: (
-                  <div className="grid grid-cols-2 gap-4">
+                  <div className="grid grid-cols-1 gap-4 lg:grid-cols-2">
                     <DetailsInfoCard
                       cardTitle="buildDetails.buildInfo"
                       data={[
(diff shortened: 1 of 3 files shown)


MarceloRobert (commented):
your changes made it so that the treeDetails page now has a horizontal scroll. It shows the tables entirely, but makes the cards have a very useless empty space
Image

also, cards in details pages can still clip to the right; not necessary, but it's a nice-to-have
Image

alanpeixinho (approved):
Seems to be working on all screens.

gustavobtflores (approved):
LGTM

PR #1804, opened 2026-03-16
fix: use definitive sqlite path for cronjobs

MarceloRobert:
Fixes the case when an empty BACKEND_VOLUME_DIR would cause cronjobs to not find the right file


gustavobtflores (approved):
LGTM

PR #1807, opened 2026-03-17
fix: too many logs on cronjob

### LINKED ISSUE #1806: Checkout cache cronjob has too many logs
Every time that cronjob runs it logs that a bunch of start_times are already aware; this is not a breaking-error so the command still runs, but it bloats the logs a lot

MarceloRobert:
Postgres start_time is already timezone-aware, meaning that we don't need to try to make it aware and log the error all the time

## How to test

You can add this task somewhere to trigger it manually (like inside a view for example) or you can change the interval that it runs on in the cron job (in settings.py, you can use the interval */2 * * * * for example to run every 2 minutes) and run the cronjob locally or within docker.

Closes #1806

### DIFF
--- a/backend/kernelCI_app/tasks.py
+++ b/backend/kernelCI_app/tasks.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from kernelCI_app.helpers.logger import out, log_message
+from kernelCI_app.helpers.logger import out
 from kernelCI_app.models import Checkouts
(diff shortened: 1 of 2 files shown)


gustavobtflores (approved):

alanpeixinho (approved):
LGTM

dede999 (approved):
LGTM

PR #1813, opened 2026-03-18
feat: add env var for db migrations

MarceloRobert:
Removes the automatic migrations in favor of a more granular environment variable to allow migrations; that variable defaults to false so auto migrations are essentially disabled until set

## How to test
Change the models with something, set the new variable to `false` (or just don't set it) and start the backend through docker (`docker compose up backend --build`); you'll see that the migrations will not be applied. Set the variable to `true` and restart the backend, then the migrations will be applied.

Closes #1818


alanpeixinho (approved):
LGTM

gustavobtflores (approved):
Looks good

PR #1824, opened 2026-03-25
fix: Fix unecessary scroll on some SideMenu

alanpeixinho:
## Description

Avoid showing scrollbar when SideMenu fits the screen.

## How to test

Run locally on a chromium based browser.
Verify that the left side menu does not show a scrollbar.
Reduce the window until the side menu no longer fits, and verify that a scrollbar is shown.

left-menu-no-scroll

Closes #1822


gustavobtflores (approved):
Looks good

PR #1825, opened 2026-03-25
fix: improve IssuesList layout and text truncation

### LINKED ISSUE #1426: Move issue first seen data under its title for small screens
For bigger screens it is correct to show then side by side, but it clips onto the card side after a certain size. That info shold be shown under the issue comment/title for a better visibility

Image

dede999:
## Description

- Extract per-issue rendering from `IssuesList` map callback into a
  dedicated `IssueItem` component, reducing cyclomatic complexity
- Replace flex layout with `grid-cols-[auto_minmax(0,1fr)_auto]` in
  `IssueItem` so the text column always truncates correctly and tag
  pills / first_seen date are placed in a second grid row (col-start-2)
  without competing with the text column on small screens
- Fix `ListingItem` tooltip trigger: replace hardcoded per-breakpoint
  `max-w-[Xpx]` values with `max-w-full` so truncation respects the
  container rather than an arbitrary pixel cap

## How to test

- Verify issue text truncates at the correct boundary (fills
      available space, no premature cut) at 425px, 768px, 1024px and
      full width
- Verify `first_seen` date and tag pills appear below the text on
      small screens without overlapping or hiding the issue comment
- Verify action icons (report link, details) remain pinned to the
      right at all breakpoints
- Check other pages that use `ListingItem` with `tooltip` for
      unexpected layout changes

### Related task

- Closes #1426

### DIFF
--- a/dashboard/src/components/Cards/IssuesList.tsx
+++ b/dashboard/src/components/Cards/IssuesList.tsx
@@ -40,6 +40,119 @@ import { MemoizedSectionError } from '@/components/DetailsPages/SectionError';
 
 import { GroupedTestStatus } from '@/components/Status/Status';
 
+type IssueItemProps = {
+  issue: TIssue;
+  extraDetails?: IssueExtraDetailsDict[string];
+  extraDetailsLoading?: boolean;
+  detailsId?: string;
+  getIssueLink: (id: string, version: number) => LinkProps;
+  issueFilterSection: TFilterObjectsKeys;
+  diffFilter: TFilter;
+};
+
+const IssueItem = ({
+  issue,
+  extraDetails,
+  extraDetailsLoading,
+  detailsId,
+  getIssueLink,
+  issueFilterSection,
+  diffFilter,
+}: IssueItemProps): JSX.Element => {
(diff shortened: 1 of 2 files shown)


MarceloRobert (commented):
This is changing the layout in large screens as well, which is not supposed to happen. I like the refactoring, but since the only broken thing was on small screens, we shouldn't (and don't even need to) change the layout on larger screens

alanpeixinho (approved):
Pretty good

MarceloRobert (commented):
There's no point in limiting the text when there is space to show more of it. Either go back to showing the issue comment on the whole line or move the "first seen" text right after the comment, so that they are inline.

We can merge this with the same layout from mobile and desktop, but the current change is a step back to what we had, unnecessary change.

Also, please squash the commits that you have, they are way too small and end up 1st not being useful and 2nd being slower to maintain as a developer

gustavobtflores (approved):
Looks good

MarceloRobert (approved):
Looks ok, this fixes the mobile screen. If we decide to roll back/change the layout we can create another issue, but nothing's broken

PR #1827, opened 2026-03-26
fix: Fix Hardware Details Link

alanpeixinho:
## Description

Uses the `from` property, instead of directly capturing the hardwareId from url.
This way, any page that uses a proper `id` and indicates that we are coming from a hardware page is going to be interpreted as a hardwareId.

## How to test

1. Open the hardware listing page
2. Click on a hardware to go to Hardware Details page
3. Go to TestDetails or BuildDetails tab
4. On the configs table access the BuildDetails/TestDetails page through the information icon (i).
5. Click on the `View Log Excerpt`.
6. Click on `View full log for` link.
7. The second breadcrumb should show (and link to) HardwareDetails page, and not TreeDetails page.

 Closes #1823 hardwareDetails -> boots tab -> testDetails/bootDetails -> log viewer

### DIFF
--- a/dashboard/src/components/Log/LogViewerCard.tsx
+++ b/dashboard/src/components/Log/LogViewerCard.tsx
@@ -2,7 +2,7 @@ import { FormattedMessage, useIntl } from 'react-intl';
 
 import { useMemo, type JSX } from 'react';
 
-import { Link, useParams, useRouterState } from '@tanstack/react-router';
+import { Link, useRouterState } from '@tanstack/react-router';
 
 import { SearchIcon } from '@/components/Icons/SearchIcon';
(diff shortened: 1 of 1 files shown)


MarceloRobert (commented):
One route is still incorrect:

hardwareListing -> hardwareDetails -> build/testDetails -> view full logs -> build/testDetails -> hardware Details. The text of the last link is correct but it redirects to `/hardware/<tree_commit_hash>`

MarceloRobert (approved):
The main problem was fixed, what I pointed out could be considered "a very specific case", so sure let's open an issue for improving the breadcrumb system and we can merge this small fix

gustavobtflores (approved):

Issue #1718, opened 2026-01-28
Unify lab field


MarceloRobert:
Currently the lab value comes from `build['misc']['lab']` and `test['misc']['runtime']`. We could rearrange the data through the ingester to make it so that both of them exist in the `lab` field.

Currently this would require some refactoring since there are some functions in place for runtime already

Mentioned in https://github.com/kernelci/dashboard/pull/1717#pullrequestreview-3717905590

PR #1832, opened 2026-03-30
Semi aggregated query for hardware details summary endpoint

alanpeixinho:
## Description

This implements performance improvement on the Hardware details summary endpoint.

* Semi aggregating data on query to reduce number of returned rows, while  allowing for filtering and detailed aggregation.
* Ignoring fields not used on the frontend (failed_reasons).

## How to test
* Open the dashboard.
* Go to Hardware page.
* Select any available hardware.
* The presented information should match previous versions (and staging/production).
* The page should load fast even for cases with many instances of builds/boots and tests.


MarceloRobert (changes requested):
Seems like there's some discrepancy between the listing and the details, specially with larger hardware such as kubernetes

backend/kernelCI_app/helpers/filters.py
```diff
@@ -361,7 +372,7 @@ def __init__(self, data: Dict, process_body=False) -> None:
             "test.status": self._handle_test_status,
             "test.duration": self._handle_test_duration,
             "build.status": self._handle_build_status,
```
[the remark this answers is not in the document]
MarceloRobert: > DO you think is a good approach to keep both, for safety?

I think it is. Honestly i think it would make more sense for it to be `build.duration` anyway but let's not change that for now

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,190 +55,396 @@
...
+            and is_filtered_out(status, filters.filterTestStatus)
+        ):
+            return True
+        if is_filtered_out(config, filters.filterConfigs):
+            return True
+        if is_filtered_out(lab, filters.filter_labs):
+            return True
+        if is_filtered_out(architecure, filters.filterArchitecture):
+            return True
+
```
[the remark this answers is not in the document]
alanpeixinho: fixed

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,190 +55,396 @@
...
+        return False
```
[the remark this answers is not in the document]
alanpeixinho: included, even hardware (which might be not needed, since we are in a detailing of the hardware)

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,190 +55,396 @@
...
+                t.tree_name or "",
+                t.git_repository_branch or "",
+                t.head_git_commit_name or "",
+            ),
+        )
+        for i, tree in enumerate(sorted_trees):
+            tree.index = str(i)
+
+        return sorted_trees, sorted(all_compatibles)
+
```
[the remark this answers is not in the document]
alanpeixinho: Nice catch, I included an option to discard the hardware_id from the all compatibles

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,190 +55,396 @@
...
+        builds_labs = {*builds_summary.labs}
+        boots_labs = {*boots_summary.labs}
+        tests_labs = {*tests_summary.labs}
+
+        builds_origins = {*builds_summary.origins}
+        boots_origins = {*boots_summary.origins}
+        tests_origins = {*tests_summary.origins}
+
+        boots_platforms = {*(boots_summary.platforms or [])}
+        tests_platforms = {*(boots_summary.platforms or [])}
```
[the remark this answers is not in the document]
alanpeixinho: Nice catch

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -262,118 +456,104 @@ def _format_processing_for_response(self, hardware_id: str) -> None:
...
-                ),
+                summary=summary, filters=filters, common=commons
             )
-        except ValidationError as e:
-            return Response(data=e.json(), status=HTTPStatus.INTERNAL_SERVER_ERROR)
 
-        return Response(valid_response.model_dump())
+            return Response(valid_response.model_dump())
+        except ValidationError as e:
+            return Response(data=e.json(), status=HTTPStatus.BAD_REQUEST)
```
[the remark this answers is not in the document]
alanpeixinho: Thanks, I fixed this

backend/kernelCI_app/helpers/filters.py
```diff
@@ -22,6 +23,10 @@
 NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
 
 
+def is_valid_status(status: str) -> bool:
+    return status in StatusChoices or status == "NULL"
```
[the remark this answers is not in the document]
MarceloRobert: I meant `status` as a whole, not just the `NULL` option. So that would be more like `status.upper() in StatusChoices or status.upper() == "NULL"` (not sure if .upper() exists though)

backend/kernelCI_app/helpers/filters.py
```diff
@@ -22,6 +23,10 @@
 NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
 
 
+def is_valid_status(status: str) -> bool:
+    return status in StatusChoices or status == "NULL" or status == "null"
```
gustavobtflores: very nit: `status.lower() == "null"` or `status.upper() == "NULL"` wouldn't work?

backend/kernelCI_app/helpers/filters.py
```diff
@@ -129,6 +134,12 @@ def is_issue_filtered_out(
     return not in_filter
 
 
+def is_filtered_out(value: str, filter_values: set[set]):
+    if filter_values and value not in filter_values:
+        return True
+    return False
```
gustavobtflores: I think its more concise:

```suggestion
def is_filtered_out(value: str, filter_values: set[set]):
    return filter_values and value not in filter_values
```

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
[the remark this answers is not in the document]
MarceloRobert: @alanpeixinho there are still some cases of `architecure`s in other variables, like `tests_architecures`

backend/kernelCI_app/tests/integrationTests/hardwareDetailsSummary_test.py
```diff
@@ -411,6 +413,7 @@ def test_filter_issues(issues_input):
 
 
 def test_invalid_filters(invalid_filters_input):
+
```
gustavobtflores: patch noise

backend/kernelCI_app/typeModels/common.py
```diff
@@ -14,12 +14,12 @@ class StatusCount(BaseModel):
     DONE: Optional[int] = 0
     NULL: Optional[int] = 0
 
-    def increment(self, status: Optional[str]) -> None:
+    def increment(self, status: Optional[str], value=1) -> None:
```
gustavobtflores: maybe type value here?

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,190 +55,396 @@
...
-            boots_summary=self.boots_summary,
-            tests_summary=self.tests_summary,
-            issue_dicts=self.issue_dicts,
+    def sanitize_filters(self, available_filters: dict[str, list]):
+        filters: FilterParams = self.filters
+        filters.filterCompiler.intersection_update(available_filters["compiler"])
+        filters.filterArchitecture.intersection_update(
+            available_filters["architecture"]
         )
+        filters.filterConfigs.intersection_update(available_filters["config_name"])
```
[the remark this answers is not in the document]
MarceloRobert: @alanpeixinho Can you comment on the behavior change?

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,414 @@
...
+        origin: str,
+        lab: str,
+        compiler: str,
+        architecture: str,
+        status: str,
+        known_issues: set[str],
+        is_build: bool,
+        is_boot: bool,
+        is_test: bool,
+    ) -> bool:
```
gustavobtflores, line 106: origin seems unused here

MarceloRobert (commented):
Looks good, the behavior seems correct. Only some small changes left about the comments that were already made

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,414 @@
...
+        if is_build and is_filtered_out(status, filters.filterBuildStatus):
+            return True
+        if is_boot and is_filtered_out(status, filters.filterBootStatus):
+            return True
+        if (
+            is_test
+            and not is_boot
+            and is_filtered_out(status, filters.filterTestStatus)
+        ):
+            return True
```
gustavobtflores: here I think we could do:

```suggestion
      filter_type = self.get_filter_type(is_build, is_boot, is_test)
      status_filter_map = {
          "build": filters.filterBuildStatus,
          "boot":  filters.filterBootStatus,
          "test":  filters.filterTestStatus,
      }
      if is_filtered_out(status, status_filter_map[filter_type]):
          return True
```

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,414 @@
...
+        if is_filtered_out(config, filters.filterConfigs):
+            return True
+        if is_filtered_out(lab, filters.filter_labs):
+            return True
+        if is_filtered_out(architecture, filters.filterArchitecture):
+            return True
+        if is_filtered_out(hardware_id, filters.filterHardware):
+            return True
```
gustavobtflores, line 127: these ones couldn't be a list together with compiler? or order is important here?

```suggestion
        field_checks = [
            (compiler,     filters.filterCompiler),
            (config,       filters.filterConfigs),
            (lab,          filters.filter_labs),
            (architecture, filters.filterArchitecture),
            (hardware_id,  filters.filterHardware),
        ]
        if any(is_filtered_out(val, filt) for val, filt in field_checks):
            return True
```

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,414 @@
...
+                    tests_summary=tests_summary,
+                    status_count=status_count,
+                    config=config,
+                    lab=lab,
+                    origin=origin,
+                    known_issues=len(known_issues) - 1,
+                    architecture=architecture,
+                    compiler=compiler,
+                    platform=platform,
+                )
```
gustavobtflores, line 218: I think we could use `get_filter_type` here too to avoid multiple branches, not sure though, because of the params differences

backend/kernelCI_app/helpers/filters.py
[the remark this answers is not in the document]
alanpeixinho: got it, I was under the impression only the null status had case changes.

backend/kernelCI_app/helpers/filters.py
alanpeixinho: got it, the only problem here, is that we might accept combinations such as "Null", "nUll". But I dont think this is a big problem.

backend/kernelCI_app/helpers/filters.py
alanpeixinho: good point

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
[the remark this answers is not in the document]
alanpeixinho: This is no longer truth, we reverted back to the original behavior of filtering every row on invalid filters.

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
alanpeixinho: good catch

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
alanpeixinho: fixed

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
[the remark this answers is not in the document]
alanpeixinho: I have learned today that I have some selective dyslexia towards the word *architecture*.

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
alanpeixinho: good call

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
alanpeixinho: we could go for something like:

```python

            summary_type = self.get_summary_type(is_build=is_build, is_boot=is_boot, is_test=is_test)
            increment_strategy = {
                'builds': partial(increment_build, self, builds_summary=builds_summary),
                'boots': partial(increment_test, self, tests_summary=boots_summary),
                'tests': partial(increment_test, self, tests_summary=tests_summary),
            }

            increment_strategy[summary_type](
                status_count=status_count,
                architecture=architecture,
                config=config,
                lab=lab,
                origin=origin,
                known_issues=len(known_issues) - 1,
                compiler=compiler,
                platform=platform,
            )
```

But I am afraid is a little bit of overengineer for 3 conditionals

alanpeixinho: What you think of it?

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +340,260 @@ def get_hardware_details_data(
...
+    boots_tests_duration_clause = _get_boot_test_duration_clause(
+        boots_duration, tests_duration
+    )
+
+    query = """
+           (SELECT
+                 COUNT(distinct builds.id) AS count,
+                 checkouts.origin,
+                 builds.status AS status,
+                 count(incidents) as incidents_count,
```
MarceloRobert: do you know if `count(distinct incidents.id)` would make a difference?

MarceloRobert (commented):
From testing, seems like the hardware compatible filter is always returning empty, and the tree filter is not working

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +340,260 @@ def get_hardware_details_data(
     return records
 
 
+def _get_build_duration_clause(builds_duration) -> str:
+    clause = ""
+
+    # builds
+    duration_min, duration_max = builds_duration
+    if duration_min:
+        clause += f"AND builds.duration >= {duration_min}\n"
```
dede999: 🔴 **bad** — SQL injection via f-string interpolation.

`duration_min` and `duration_max` are interpolated directly into SQL via f-strings here and on the lines below (`_get_boot_test_duration_clause` has the same pattern). The rest of the query correctly uses `%s` placeholders — these should too.

```python
def _get_build_duration_clause(builds_duration, params: list) -> str:
    clause = ""
    duration_min, duration_max = builds_duration
    if duration_min:
        clause += "AND builds.duration >= %s\n"
        params.append(duration_min)
    if duration_max:
        clause += "AND builds.duration <= %s\n"
        params.append(duration_max)
    return clause
```

Same fix needed for `_get_boot_test_duration_clause`.

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +340,260 @@ def get_hardware_details_data(
...
+                 checkouts.git_commit_tags,
+                 checkouts.git_commit_name,
+                 checkouts.git_repository_branch,
+                 checkouts.git_commit_hash,
+                 true AS is_build,
+                 false AS is_test,
+                 false AS is_boot
+             FROM
+                builds
+            INNER JOIN tests ON
```
dede999, line 459: 🟡 **medium** — The build subquery does `INNER JOIN tests ON tests.build_id = builds.id`, which means builds without any test records will be silently dropped from the summary. Is this intentional? If a build exists but no tests have run for it yet, it won't appear.

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +340,260 @@ def get_hardware_details_data(
...
+                 checkouts.git_commit_tags,
+                 checkouts.git_commit_name,
+                 checkouts.git_repository_branch,
+                 checkouts.git_commit_hash,
+                 false AS is_build,
+                 true AS is_test,
+                 (tests.path like 'boot.%%' or tests.path = 'boot') AS is_boot
+             FROM
+                builds
+            inner JOIN tests ON
```
dede999: 🔵 **nit** — Inconsistent SQL casing: `inner JOIN` here vs `INNER JOIN` on the build subquery (line ~505). Minor but easy to standardize.

dede999 (commented):
## Review Summary

The performance strategy is solid — moving aggregation into SQL (`GROUP BY` + `UNION ALL`) to reduce rows returned to Python is the right approach. A few items to address before merge, the most critical being the SQL injection in duration clauses.

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +340,260 @@ def get_hardware_details_data(
...
+        hardware_id,
+        hardware_id,
+        origin,
+        start_datetime,
+        end_datetime,
+        *commit_hashes,
+    ]
+
+    # TODO: check if we can reuse parameters to avoid double passing
+    params = [*params, *params]
```
dede999: 🟡 **medium** — `params = [*params, *params]` duplicates all params including commit hashes.

The TODO acknowledges this, but if the two `SELECT` branches ever diverge in parameter needs (one already has `builds_duration_clause` vs `boots_tests_duration_clause`), this duplication will silently break. Consider building params per-branch to make it explicit.

backend/kernelCI_app/helpers/filters.py
```diff
@@ -22,6 +23,10 @@
 NULL_STRINGS = set(["null", UNKNOWN_STRING, "NULL"])
 
 
+def is_valid_status(status: str) -> bool:
```
dede999, line 26: 🔴 **bad** — `{*StatusChoices, "NULL"}` reconstructs the set on every call, and if `status` is `None`, `status.upper()` will raise `AttributeError`.

Suggestion:
```python
VALID_STATUSES = {choice.value for choice in StatusChoices} | {"NULL"}

def is_valid_status(status: str) -> bool:
    return status is not None and status.upper() in VALID_STATUSES
```

backend/kernelCI_app/helpers/filters.py
```diff
@@ -129,6 +134,10 @@ def is_issue_filtered_out(
     return not in_filter
 
 
+def is_filtered_out(value: str, filter_values: set[set]):
```
dede999: 🟡 **medium** — Type hint `set[set]` should be `set[str]` (or just `set`). Won't break at runtime but misleads type checkers and readers.

backend/kernelCI_app/helpers/filters.py
```diff
@@ -361,6 +370,7 @@ def __init__(self, data: Dict, process_body=False) -> None:
             "test.status": self._handle_test_status,
             "test.duration": self._handle_test_duration,
             "build.status": self._handle_build_status,
+            "duration": self._handle_build_duration,  # TODO: same as build.duration (should be standardized)
```
dede999, line 373: 🔵 **nit** — If this `"duration"` alias is known tech debt, consider creating a ticket instead of a TODO — it may confuse future contributors about which filter key to use.

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,410 @@
...
+
+        tests_summary.platforms = {}
+        boots_summary.platforms = {}
+
+        # aggregation
+        for instance in summary:
+            status = instance["status"]
+            count = instance["count"]
+            incidents = instance["incidents_count"]
+            known_issues = set(parse_issue(issue) for issue in instance["known_issues"])
```
dede999, line 150: 🟡 **medium** — When `instance["known_issues"]` contains `None` entries (from `array_agg` when there are no incidents), `parse_issue(None)` returns `(UNCATEGORIZED_STRING, None)`. These phantom tuples are then checked against `filters.filterIssues`, which could cause false-positive filtering.

Suggestion:
```python
known_issues = set(parse_issue(issue) for issue in instance["known_issues"] if issue is not None)
```

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,189 +59,410 @@
...
+                    lab=lab,
+                    origin=origin,
+                    incidents=incidents,
+                    architecture=architecture,
+                    compiler=compiler,
+                    platform=platform,
+                )
+
+        # ensure uniqueness on architecture and compilers (maybe we could change data structures???)
+        for summary in builds_summary.architectures.values():
```
dede999, line 221: 🔵 **nit** — The loop variable `summary` shadows the method parameter `summary: list[dict]`. Consider renaming to `arch_summary` or similar to avoid confusion.

backend/kernelCI_app/typeModels/commonDetails.py
```diff
@@ -39,6 +39,28 @@ class TestArchSummaryItem(BaseModel):
...
+    def __add__(self, other: StatusCount) -> "BuildArchitectures":
+        return BuildArchitectures(
+            PASS=self.PASS + other.PASS,
+            ERROR=self.ERROR + other.ERROR,
+            FAIL=self.FAIL + other.FAIL,
+            SKIP=self.SKIP + other.SKIP,
+            MISS=self.MISS + other.MISS,
+            DONE=self.DONE + other.DONE,
+            NULL=self.NULL + other.NULL,
+            compilers=self.compilers,
```
dede999: 🟡 **medium** — `__add__` only keeps `self.compilers`, silently discarding `other.compilers`. `__iadd__` also doesn't merge compilers. This might be intentional (compilers are merged separately in `aggregate_summaries`), but it's a silent data-loss trap if these operators are used elsewhere.

backend/kernelCI_app/helpers/filters.py
alanpeixinho: This shouldn't be a problem, as it is not in any hot path of the application, also the set is quite small.
The status is a str, not Optional[str], it should not be None.

backend/kernelCI_app/helpers/filters.py
alanpeixinho: Good point, let this one slip

backend/kernelCI_app/queries/hardware.py
alanpeixinho: Good call, I will include named parametrization to avoid this

backend/kernelCI_app/queries/hardware.py
alanpeixinho: @MarceloRobert , builds without associated should also be considered in such scenario?

backend/kernelCI_app/typeModels/commonDetails.py
alanpeixinho: StatusCount does not have a compilers, if we do include an option to add two BuildArchitectures (which could be the case some point in future), than we would need to deal with this

alanpeixinho: I also dont know if Build Architectures should inherit from StatusCount, composition might make more sense in this case, instead of inheritance

backend/kernelCI_app/queries/hardware.py
alanpeixinho: I am using distinct now

backend/kernelCI_app/queries/hardware.py
MarceloRobert: I don't think so, the hardwareDetails page only shows the builds related to the tests that were performed in that hardware, so if a build isn't related to any tests then it won't be shown in that hardware

dashboard/src/pages/hardwareDetails/HardwareDetailsHeaderTable.tsx
```diff
@@ -362,10 +361,6 @@ const indexesFromRowSelection = (
     parseInt(rowId),
   );
 
-  if (selectedIndexes.length === maxTreeItems) {
-    return null;
-  }
-
   return selectedIndexes;
 };
```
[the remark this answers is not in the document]
alanpeixinho: No, this behavior could trigger a re-render infinite loop with no clear benefit.

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
```diff
@@ -144,20 +144,21 @@ function HardwareDetails(): JSX.Element {
...
+      if (selectedIndexes !== null) {
+        // avoid infinite loop on re-render
+        navigate({
+          search: previousSearch => ({
+            ...previousSearch,
+            treeIndexes: selectedIndexes,
+          }),
+          state: s => s,
+        });
+      }
```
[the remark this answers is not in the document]
alanpeixinho: No, this behavior could create a re-render loop.

backend/kernelCI_app/queries/hardware.py
alanpeixinho: Solved when changed to named parameters

backend/kernelCI_app/typeModels/commonDetails.py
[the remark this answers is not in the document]
alanpeixinho: Fixed. Should not be triggered in our scenario, but it is safer this way.

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,199 +58,436 @@
...
+        if is_filtered_out(compiler, filters.filterCompiler):
+            return True
+        if is_filtered_out(config, filters.filterConfigs):
+            return True
+        if is_filtered_out(lab, filters.filter_labs):
+            return True
+        if is_filtered_out(architecture, filters.filterArchitecture):
+            return True
+        if is_filtered_out(hardware_id, filters.filterHardware):
+            return True
```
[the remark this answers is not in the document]
alanpeixinho: Fixed

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +339,220 @@ def get_hardware_details_data(
...
+        "start_date": start_datetime.timestamp(),
+        "end_date": end_datetime.timestamp(),
+        "builds_duration": builds_duration,
+        "boots_duration": boots_duration,
+        "tests_duration": tests_duration,
+    }
+
+    query_rows = get_query_cache(cache_key, tests_cache_params)
+
+    if query_rows:
```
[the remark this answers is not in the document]
alanpeixinho: Fixed

backend/kernelCI_app/queries/hardware.py
```diff
@@ -339,15 +339,220 @@ def get_hardware_details_data(
...
+           (SELECT
+                 COUNT(distinct builds.id) AS count,
+                 checkouts.origin,
+                 builds.status AS status,
+                 count(distinct incidents.id) as incidents_count,
+                 array_agg(distinct incidents.issue_id || ',' || incidents.issue_version::text)
+                     as known_issues,
+                 array[builds.compiler, builds.architecture] AS compiler_arch,
+                 builds.config_name,
+                 builds.misc->>'runtime' AS lab,
```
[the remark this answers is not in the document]
alanpeixinho: @MarceloRobert  any toughts on this one?

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,199 +58,436 @@
...
+
+        builds_summary.status += status_count
+        builds_summary.configs[config] += status_count
+        builds_summary.labs[lab] += status_count
+        builds_summary.origins[origin] += status_count
+        builds_summary.architectures[architecture] += status_count
+        if compiler not in (builds_summary.architectures[architecture].compilers or []):
+            builds_summary.architectures[architecture].compilers.append(compiler)
+        if status_count.FAIL > 0:
+            builds_summary.unknown_issues += status_count.FAIL - incidents
```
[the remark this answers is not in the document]
alanpeixinho: It could hide a potential bug, it is better to let it surface and fix if it happens.

backend/kernelCI_app/queries/hardware.py
```diff
@@ -512,6 +720,90 @@ def get_hardware_summary_data(
         return dict_fetchall(cursor)
 
 
+def get_hardware_trees_head_commits(
+    *,
+    hardware_id: str,
+    origin: str,
+    start_datetime: datetime,
+    end_datetime: datetime,
+) -> list[(str, str)]:
```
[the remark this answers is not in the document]
alanpeixinho: Fixed

backend/kernelCI_app/queries/hardware.py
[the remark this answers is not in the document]
MarceloRobert: builds use `lab` while tests use `runtime`, the correction is right. Also see https://github.com/kernelci/dashboard/issues/1718

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
```diff
@@ -144,20 +144,21 @@ function HardwareDetails(): JSX.Element {
...
-    });
-    // Since UpdateTreeFilters is used on a useEffect, anytime that updateTreeFilters is called
-    // it will trigger a rerender, changing the navigate object, changing this function,
-    // calling useEffect again on an infinite loop.
-    // TODO: check for uses of navigate as a dependency
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  const updateTreeFilters = useCallback(
+    (selectedIndexes: number[] | null) => {
+      if (selectedIndexes !== null) {
```
MarceloRobert: there isn't an infinite loop but Copilot is right in the case of "when all trees are selected all the indexes will be shown in the url". This is technically a regression because we want to keep the url as small as possible, but since it is a minor thing, I think we should move on and fix it later

MarceloRobert (approved):
Looks like the behavior is the same as before now, all good. The opened comments still make sense, but we can work on them later; maybe make a TODO for them or open an issue if it's major. Either way, I'll approve; it's working well

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
alanpeixinho: Yes, I included a comment in the copilot comment.
Yes, it will show all indexes.
It was precisely the behavior of clearing the filters that was causing the re render loop.
If it is important to keep the query parameters clean in such scenario (I would argue it is a really low priority), we can open a new ticket.

MarceloRobert: It is important to keep the url small since we already keep a lot of the state in the url (think the filters, the specific tab, tree commit...) so we want to keep it as short as possible, and if the default state is to have all trees selected, then it can be hidden instead of showing all indices

MarceloRobert (approved):

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,199 +58,434 @@
...
+        if selected_commits:
+            for idx, head in tree_heads:
+                if idx in self.selected_commits:
+                    selected_commit = self.selected_commits.get(idx, "head")
+                    selected_commit_hashes.append(
+                        head if selected_commit == "head" else selected_commit
+                    )
+        else:
+            selected_commit_hashes = [head for (_, head) in tree_heads]
+        return selected_commit_hashes
```
gustavobtflores, line 486: We are passing `selected_commits` as a param and then using `self.selected_commits` inside the function, I'm not sure if we need the param. The only function call that uses `selected_commits` also passes self.selected_commits

backend/kernelCI_app/views/hardwareDetailsSummaryView.py
```diff
@@ -67,199 +58,434 @@
...
-    @extend_schema(
-        parameters=[HARDWARE_ID_PATH_PARAM],
-        responses=HardwareDetailsSummaryResponse,
-        request=HardwareDetailsPostBody,
-        methods=["POST"],
-    )
-    def post(self, request, hardware_id) -> Response:
+    def select_commits_hashes(
+        self,
+        tree_heads: list[(str, str)],
```
gustavobtflores, line 473: nit:


```suggestion
        tree_heads: list[tuple[str, str]],
```

gustavobtflores (approved):
Looks good, just some minor comments. I would leave some TODOs for refactoring `filter_instance` and `aggregate_summaries` in `hardwareDetails.py` to reduce repetition and branching, not a blocker though

backend/kernelCI_app/queries/hardware.py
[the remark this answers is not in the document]
alanpeixinho: Fixed

PR #1835, opened 2026-04-01
feat: add monitoring to cronjobs

### LINKED ISSUE #1792: Add healthcheck to cronjobs
Our email system runs on cronjobs and it is a very high priority, we should know when it goes down so that we can fix it as soon as possible

cc @bhcopeland

MarceloRobert:
## Changes
- Added healthcheck_id parameter to relevant commands used in the cronjobs
- Added map from healthcheck ids to healthcheck.io uuid, the ids reflect the defined cronjobs and are set via environment_variable
- Added wrapper to run functions with healthcheck.io pings for start, success and failure
- Applied healthcheck changes to main cronjob execution

## How to test
For this PR, the ideal way to test would be with a healthcheck account and uuid so you can monitor the pings. Follow the instructions there to create an account and a check.

- In the code, set one of the cronjobs to `notifications --action=fake-report` and set a small interval, such as `*/2 * * * *` (run every 2 minutes). That is:
```
(
    "*/2 * * * *",
    "django.core.management.call_command",
    [
        "notifications",
        "--monitoring-id=notifications_new_issues",
        "--action=fake_report",
    ],
),
```
- You won't need any other cronjob for testing, but if you want to make more than one check you're welcome
- **Check that the cronjobs are not going to send email reports accidentally**
- Get your healthcheck id and use it in the cronjobs
- Start the cronjobs either via `docker compose up backend --build` or locally with `poetry run ./manage.py crontab add` (docker is the recommended way)
- In the healthcheck.io dashboard, see the details of your healthcheck and monitor the received pings

The cronjobs should even if the ping fails or the monitoring_id is empty

## Required production changes
- Create 6 different checks for the 6 monitored cronjobs
- Set up on Discord a desired webhook (could reuse the already-present ones)
- Set up in the checks the integration to webhook, POSTing to the discord webhook when a check goes down. Example of body:
```
{
  "avatar_url": "https://avatars.githubusercontent.com/u/11725450?s=200&v=4",
  "username": "KernelCI Healthcheck Notifications",
  "content": "Healthcheck.io Alert: $NAME service went down!"
}
```
Also set the request headers to `Content-Type: application/json`
- Disable the default email notification integration in the checks

Closes #1792

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/healthcheck.py
+++ b/backend/kernelCI_app/management/commands/helpers/healthcheck.py
@@ -1,0 +1,69 @@
+from collections.abc import Callable
+from typing import Any, Literal
+
+from django.conf import settings
+
+import requests
+
+from kernelCI_app.helpers.logger import log_message
+
+MONITORING_ID_PARAM_HELP_TEXT = (
+    "Monitoring ID configured in settings for healthcheck.io pings "
+    "(optional, used only for monitoring the command execution over time)"
+)
+type PingStatus = Literal["start", "fail", "success"]
+
+
+def _resolve_monitoring_url(*, monitoring_id: str, status: PingStatus) -> str | None:
+    healthcheck_base_url: str = settings.HEALTHCHECK_BASE_URL
+    monitoring_path_map: dict[str, str] = settings.HEALTHCHECK_MONITORING_PATH_MAP
+    monitoring_path = monitoring_path_map.get(monitoring_id)
+
+    if not monitoring_path:
+        return None
+
+    # Success just needs to ping base healthcheck.io url + uuid, no subpath
+    status_suffix = f"/{status}" if status != "success" else ""
+
+    return f"{healthcheck_base_url.rstrip('/')}/{monitoring_path.lstrip('/')}{status_suffix}"
+
+
+def _ping_healthcheck(*, monitoring_id: str, status: PingStatus) -> None:
+    monitoring_url = _resolve_monitoring_url(monitoring_id=monitoring_id, status=status)
+    if not monitoring_url:
+        log_message(
+            "No healthcheck URL configured for monitoring_id='%s', skipping %s ping."
+            % (monitoring_id, status)
+        )
+        return
+
+    try:
+        response = requests.get(monitoring_url, timeout=10)
+        response.raise_for_status()
+        log_message(
+            "Success at pinging healthcheck '%s' with monitoring_id '%s'"
+            % (monitoring_url, monitoring_id)
+        )
+    except requests.RequestException as e:
+        log_message(
+            "ERROR: failed to ping healthcheck for monitoring_id='%s' and status='%s': %s"
+            % (monitoring_id, status, e)
+        )
+
+
+def run_with_healthcheck_monitoring(
+    *, monitoring_id: str | None, action: Callable[[], Any]
+) -> Any:
+    if not monitoring_id:
+        return action()
+
+    _ping_healthcheck(monitoring_id=monitoring_id, status="start")
+
+    try:
+        result = action()
+    except Exception:
+        _ping_healthcheck(monitoring_id=monitoring_id, status="fail")
+        raise
+
+    _ping_healthcheck(monitoring_id=monitoring_id, status="success")
+    return result
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -97,6 +97,23 @@ The monitoring system supports multi-worker Gunicorn deployments using Prometheu
 - `PROMETHEUS_METRICS_PORT`: Port for the metrics aggregator (default: `8001`)
 - `PROMETHEUS_MULTIPROC_DIR`: Directory for multiprocess metric files (default: `/tmp/prometheus_multiproc_dir`)
 
+### Cronjob Healthchecks
+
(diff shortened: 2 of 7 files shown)


nuclearcat:
I will proceed tomorrow with that, thank you very much!

docs/monitoring.md
```diff
@@ -97,6 +97,23 @@ The monitoring system supports multi-worker Gunicorn deployments using Prometheu
...
+- Each monitored cron run sends pings to `/start`, `/success`, and `/fail`.
+
+Configure these variables in `.env.backend`:
+
+- `HEALTHCHECK_MONITORING_NOTIFICATIONS_NEW_ISSUES_HOURLY`
+- `HEALTHCHECK_MONITORING_NOTIFICATIONS_SUMMARY_MICROSOFT_DAILY`
+- `HEALTHCHECK_MONITORING_NOTIFICATIONS_SUMMARY_MAESTRO_DAILY`
+- `HEALTHCHECK_MONITORING_NOTIFICATIONS_HARDWARE_SUMMARY_WEEKLY`
+- `HEALTHCHECK_MONITORING_DELETE_UNUSED_HARDWARE_STATUS_WEEKLY`
+- `HEALTHCHECK_MONITORING_NOTIFICATIONS_METRICS_SUMMARY_WEEKLY`
```
gustavobtflores: these seem different from the ones in .env.example

gustavobtflores (approved):
LGTM, just update the docs to reflect the variables set on .env.example

docs/monitoring.md
MarceloRobert: I shortened the names in settings but forgot about them here, fixed.

backend/kernelCI_app/management/commands/helpers/healthcheck.py
```diff
@@ -0,0 +1,71 @@
...
+    monitoring_url = _resolve_monitoring_url(monitoring_id)
+    if not monitoring_url:
+        out(
+            "No healthcheck URL configured for monitoring_id='%s', skipping %s ping."
+            % (monitoring_id, status)
+        )
+        return
+
+    # Success just needs to ping base healthcheck.io url + uuid, no subpath
+    monitoring_status_url = (
```
dede999: The success path special-cases `status != "success"` to avoid appending a
subpath, making the URL-building logic fragile. healthcheck.io also accepts
`/{uuid}/0` for success, but more importantly, this branching makes the code
harder to follow and test.

Consider extracting a helper:

```python
    def _build_status_url(base: str, status: str) -> str:
        if status == "success":
            return base
        return f"{base.rstrip('/')}/{status}"
```

backend/kernelCI_app/management/commands/helpers/healthcheck.py
```diff
@@ -0,0 +1,71 @@
+from collections.abc import Callable
+from typing import Any
+
+from django.conf import settings
+
+import requests
+
+from kernelCI_app.helpers.logger import out
```
dede999: nit: `out()` is used for both info and error messages. The rest of the
backend uses `logging.getLogger(__name__)` with appropriate levels. Consider:

    logger = logging.getLogger(__name__)
    # then: logger.info(...), logger.warning(...), logger.error(...)

backend/kernelCI_app/management/commands/helpers/healthcheck.py
```diff
@@ -0,0 +1,71 @@
...
+
+MONITORING_ID_PARAM_HELP_TEXT = (
+    "Monitoring ID configured in settings for healthcheck.io pings "
+    "(optional, used only for monitoring the command execution over time)"
+)
+HEALTHCHECK_BASE_URL = "https://hc-ping.com"
```
dede999: nit: `HEALTHCHECK_BASE_URL` is a module-level constant but is patched directly
in tests via `@patch(... "HEALTHCHECK_BASE_URL" ...)`. This couples tests to
internals and makes it harder to change the constant's location later.

Consider moving it to `settings.py` alongside `HEALTHCHECK_MONITORING_PATH_MAP`,
which would make `_resolve_monitoring_url` purely functional and easier to test

backend/kernelCI_app/management/commands/helpers/healthcheck.py
MarceloRobert: I included the status into the `_resolve_monitoring_url` function

backend/kernelCI_app/management/commands/helpers/healthcheck.py
MarceloRobert: I think I was having problems with a simple logger before, but I replaced `out` for `log_message` and it seems to be ok now

backend/kernelCI_app/management/commands/helpers/healthcheck.py
MarceloRobert: Done

alanpeixinho (approved):
Got the pings.

dede999 (approved):

MarceloRobert:
On hold while we wait for the healthcheck configuration and bigger PR merges (to avoid conflicts in them), at this moment it's just the hardwareDetailsSummary performance changes

MarceloRobert:
Merging because the changes won't break the cronjobs even if the healthcheck is not configured. We'll keep an eye on the configuration to finish the job though

PR #1836, opened 2026-04-02
feat: add docker-compose configuration for development environment

### LINKED ISSUE #1786: Create a docker-compose-dev file only for development
Currently we have some docker-compose files that are used for deployment, would be good to have a file only for development environment

gustavobtflores:
## Description

This PR introduces a Docker-based development environment with live reload capabilities and a Makefile for standardized project commands.

## Changes

- Add Docker development environment with `docker-compose.dev.yml`, supporting live reload for both Django backend and Vite frontend
- Add `Makefile` with common development, build, and deployment commands
- Add comprehensive documentation for the development environment in `docs/dev-environment.md`

## How to test

1. Run `docker compose -f docker-compose.dev.yml up` and verify live reload works when modifying backend or frontend files
2. Run `make help` to see available commands

Closes #1786


docs/dev-environment.md
```diff
@@ -0,0 +1,423 @@
...
+
+- **Backend** — edit any `.py` file and Django's development server (`manage.py runserver`) restarts automatically. No container rebuild needed.
+- **Frontend** — edit any file under `dashboard/src/` or `dashboard/public/` and Vite's Hot Module Replacement (HMR) pushes the change to the browser instantly, without a full page reload.
+
+## Services
+
+| Service | Image / Build | Port | Purpose |
+|---|---|---|---|
+| `backend` | `./backend` (Dockerfile) | 8000 | Django dev server with live reload |
+| `dashboard_db` | `postgres:17` | — | PostgreSQL database |
```
MarceloRobert: doesn't the database expose a port?

docs/dev-environment.md
```diff
@@ -0,0 +1,423 @@
...
+```bash
+docker compose -f docker-compose.dev.yml logs -f backend
+```
+
+When you run `makemigrations` / `migrate`, do it inside the running container so migrations are applied to the live database:
+
+```bash
+docker compose -f docker-compose.dev.yml exec backend poetry run python manage.py makemigrations
+docker compose -f docker-compose.dev.yml exec backend poetry run python manage.py migrate
+```
```
MarceloRobert: these notes could be in the `## RUnning database migrations` section

docs/dev-environment.md
```diff
@@ -0,0 +1,423 @@
...
+
+```
+[vite] hmr update /src/components/MyComponent.tsx
+```
+
+CSS-only changes are injected into the page without touching JavaScript at all.
+
+If you introduce a syntax error, Vite surfaces it in the browser overlay and in the terminal — fix the file and it recovers automatically.
+
+## Running database migrations
```
MarceloRobert, line 108: could add a note that we have a script for running migrations automatically for the default database and the sqlite database

docs/dev-environment.md
```diff
@@ -0,0 +1,423 @@
...
+
+> `--build` only rebuilds images — it never touches volumes. Your database data is safe when you rebuild. To get a clean database you must use `down -v` explicitly.
+
+## Rebuilding after dependency changes
+
+Normal source code changes never require a rebuild — they are picked up via volume mounts or live reload. A rebuild is only needed when something baked into the image changes: a `Dockerfile`, `pyproject.toml`, or `pnpm-lock.yaml`.
+
+The `--build` flag forces Compose to rebuild images even if they already exist locally. Without it, `docker compose up` reuses whatever is cached.
+
+If you change `pyproject.toml` or `pnpm-lock.yaml`, rebuild the affected image:
```
MarceloRobert: there's no need to rebuild the image before `up`ing it, I think that just telling the user to pass a `--build` flag is enough

docs/dev-environment.md
```diff
@@ -0,0 +1,423 @@
...
+docker compose -f docker-compose.dev.yml down -v
+```
+
+**What does `-v` do?** By default, everything inside a container is thrown away when the container is removed, but data you want to keep (like the database) is stored in a *volume* — a piece of storage that lives outside the container and survives restarts and rebuilds. The `dashboard_db` service uses a volume to persist the PostgreSQL data files, which is why your database survives a normal `down` + `up` cycle.
+
+The `-v` flag tells Compose to delete those volumes along with the containers, permanently wiping the database. Use it when you want a completely clean slate. It is the Docker equivalent of dropping and recreating the database.
+
+> `--build` only rebuilds images — it never touches volumes. Your database data is safe when you rebuild. To get a clean database you must use `down -v` explicitly.
+
+## Rebuilding after dependency changes
```
MarceloRobert, line 178: I'd also leave a comment about using the `--force-recreate` flag if you want to re-up the containers without having to stop them

Makefile
```diff
@@ -0,0 +1,62 @@
...
+	@test -f .env || (cp .env.example .env && echo "Created .env from .env.example")
+	@test -f .env.backend || (cp .env.backend.example .env.backend && echo "Created .env.backend from .env.backend.example")
+	@test -f $(DASHBOARD_DIR)/.env || (cp $(DASHBOARD_DIR)/.env.example $(DASHBOARD_DIR)/.env && echo "Created dashboard/.env from dashboard/.env.example")
+	$(MAKE) install
+
+install: ## Install dependencies (pnpm + poetry)
+	cd $(DASHBOARD_DIR) && pnpm install --frozen-lockfile
+	cd $(BACKEND_DIR) && poetry install
+
+lint: ## Run CI-equivalent lint checks (no autofix)
```
MarceloRobert: nit: since we are also running the ruff formatter, IMO we could call this function `check`, which will also remove ambiguity with the verb "to lint"

Makefile
```diff
@@ -0,0 +1,62 @@
...
+	cd $(BACKEND_DIR) && poetry run ruff format --check .
+
+fix: ## Apply autofixes (lint, format, prettify)
+	cd $(DASHBOARD_DIR) && pnpm lint
+	cd $(BACKEND_DIR) && poetry run ruff check . --fix
+	cd $(BACKEND_DIR) && poetry run ruff format .
+	cd $(DASHBOARD_DIR) && pnpm prettify
+
+test: ## Run fast local tests
+	cd $(DASHBOARD_DIR) && pnpm test
```
MarceloRobert: vite starts the "test watcher" which doesn't exit automatically, maybe there's a flag to disable that?

MarceloRobert (commented):
LGTM, just some comments. Also, I think we can delete the other split .env example files (like .env.backend.example) so that we don't keep useless files in the repo

docker-compose.dev.yml
```diff
@@ -0,0 +1,79 @@
...
+        environment:
+            - POSTGRES_USER=${DB_USER:-admin}
+            - POSTGRES_PASSWORD=${DB_PASSWORD:-barebare}
+            - POSTGRES_DB=${DB_NAME:-dashboard}
+        volumes:
+            - dashboard-db-data:/var/lib/postgresql/data
+        networks: [private]
+        ports: ["${DB_PORT_PUBLISHED:-5434}:5432"]
+        healthcheck:
+            test: ["CMD-SHELL", "pg_isready -U ${DB_USER:-admin}"]
```
MarceloRobert: Also, this healthcheck is fallbacking the `POSTGRES_DB` to the same value as `DB_USER`, which means it tries to access an "admin" db, which doesn't exist.

You can fix this by passing the argument `-d ${DB_NAME:-dashboard}` to this command

Makefile
gustavobtflores: I would prefer to keep something like check-lint, at least

MarceloRobert (approved):
All good!

PR #1844, opened 2026-04-08
fix: fallback to origin when lab/runtime is missing in standardize_labs

### LINKED ISSUE #1752: coalesce "lab" with "origin"
Some origins are not submitting their "lab", neither on builds or tests.

dede999:
## Summary

Closes #1752

When ingesting data via `standardize_labs`, builds and tests that have
no `lab` (builds) or `runtime` (tests) value in `misc` were silently
losing their origin context. This fix uses the `origin` field as a
fallback, ensuring entries without explicit lab information can still
be grouped and filtered by origin.

**Behavior:**
- If `lab`/`runtime` is missing and `origin` exists → `origin` is set as `lab`/`runtime` in `misc`
- If the lab matches an automatic lab pattern → existing behavior is preserved; `origin` is NOT used as fallback
- If neither `lab`/`runtime` nor `origin` is present → no change

## How to verify

Run the unit tests:

```bash
cd backend
poetry run pytest kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py::TestStandardizeLabs -v -m unit
```

All 8 test cases in `TestStandardizeLabs` should pass, including the 4 new ones added in this PR that cover the fallback scenarios.

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -93,15 +93,22 @@ def standardize_tree_names(
 
 def standardize_labs(input_data: dict[str, Any]) -> None:
     """
-    Standardize labs in data, moving automatic lab names to AUTOMATIC_LAB_FIELD
+    Standardize labs in data, moving automatic lab names to AUTOMATIC_LAB_FIELD.
+    Falls back to 'origin' when 'lab' (builds) or 'runtime' (tests) is missing.
     """
 
     builds: list[dict[str, Any]] = input_data.get("builds", [])
     for build in builds:
         lab = build.get("misc", {}).get("lab")
-        if lab and AUTOMATIC_LABS.match(lab):
+        is_automatic = lab and AUTOMATIC_LABS.match(lab)
+        if is_automatic:
             build["misc"][AUTOMATIC_LAB_FIELD] = lab
             build["misc"].pop("lab", None)
+
+        if not lab or is_automatic:
+            origin = build.get("origin")
+            if origin:
+                build.setdefault("misc", {})["lab"] = origin
 
     tests: list[dict[str, Any]] = input_data.get("tests", [])
     for test in tests:
@@ -109,6 +116,11 @@ def standardize_labs(input_data: dict[str, Any]) -> None:
         if lab and AUTOMATIC_LABS.match(lab):
             test["misc"][AUTOMATIC_LAB_FIELD] = lab
             test["misc"].pop("runtime", None)
+            lab = None
+        if not lab:
+            origin = test.get("origin")
+            if origin:
+                test.setdefault("misc", {})["runtime"] = origin
 
 
 def _extract_origins_info(data: Optional[dict[str, Any]]) -> str:
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -91,36 +91,37 @@ def standardize_tree_names(
                 checkout["tree_name"] = correct_tree
 
 
+def _standardize_lab_field(item: dict[str, Any], field: str) -> None:
+    """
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -102,13 +103,21 @@ def standardize_labs(input_data: dict[str, Any]) -> None:
         if lab and AUTOMATIC_LABS.match(lab):
             build["misc"][AUTOMATIC_LAB_FIELD] = lab
             build["misc"].pop("lab", None)
+        elif not lab:
```
alanpeixinho: should we not use the same coalesce for AUTOMATIC_LABS? Or is it correct to not do the same for them?

dashboard/src/hooks/useLogData.ts
gustavobtflores, line 59: this change is already on main, remember to rebase your branch with the latest changes

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
gustavobtflores: I think we should, AFAIK the path should be:

lab is AUTOMATIC_LAB -> fallback to origin
lab is None -> fallback to origin
lab is not None -> do nothing

@MarceloRobert what do you think?

dashboard/src/hooks/useLogData.ts
dede999: The linting change is not on the PR anymore after the rebase

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
MarceloRobert: I agree, if it is one of the automatics it should fallback to the origin otherwise it will be empty

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
MarceloRobert, line 1: if the lab is in `AUTOMATIC_LAB`s then it should also go to origin otherwise it will go to None

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -102,13 +103,23 @@ def standardize_labs(input_data: dict[str, Any]) -> None:
         if lab and AUTOMATIC_LABS.match(lab):
             build["misc"][AUTOMATIC_LAB_FIELD] = lab
             build["misc"].pop("lab", None)
+            lab = None
+        if not lab:
+            origin = build.get("origin")
+            if origin:
+                build.setdefault("misc", {})["lab"] = origin
```
gustavobtflores: nit: here I would go with: 

```suggestion
        is_automatic = lab and AUTOMATIC_LABS.match(lab)
        if is_automatic:
            build["misc"][AUTOMATIC_LAB_FIELD] = lab
            build["misc"].pop("lab", None)
        
        if not lab or is_automatic:
            origin = build.get("origin")
            if origin:
                build.setdefault("misc", {})["lab"] = origin
```

so we don't depend on a branch setting `lab = None` and we have clear independent branches

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -102,13 +103,23 @@ def standardize_labs(input_data: dict[str, Any]) -> None:
...
     for test in tests:
         lab = test.get("misc", {}).get("runtime")
         if lab and AUTOMATIC_LABS.match(lab):
             test["misc"][AUTOMATIC_LAB_FIELD] = lab
             test["misc"].pop("runtime", None)
+            lab = None
+        if not lab:
+            origin = test.get("origin")
+            if origin:
+                test.setdefault("misc", {})["runtime"] = origin
```
gustavobtflores: ditto

gustavobtflores (approved):
Looks good, just a nit, pre-approving

alanpeixinho (approved):

MarceloRobert (approved):
LGTM

PR #1851, opened 2026-04-13
fix: sort aggregation values to avoid deadlocks in process_pending

gustavobtflores:
## Description

Fixes a PostgreSQL deadlock in the `tree_listing` table caused by concurrent updates from the ingester and aggregation processor. When two transactions locked overlapping rows in different orders, PostgreSQL detected a deadlock and killed one transaction. This fix ensures all transactions acquire row locks in a deterministic order by sorting values by their lock-key columns before executing `executemany()`.

## Why the deadlock happens

Two concurrent processes update the `tree_listing` table inside separate transactions:

1. The ingester runs `aggregate_checkouts_and_pendings()` inside `transaction.atomic()`, which calls `update_tree_listing()`, an `INSERT ... ON CONFLICT DO UPDATE` on `tree_listing`.

2. The aggregation processor runs `_process_tree_listing()` inside `transaction.atomic()` — a plain `UPDATE tree_listing` via `executemany()`.

Both processes use `cursor.executemany()` which sends one statement per row. PostgreSQL acquires row-level locks incrementally as each statement executes. When two transactions lock overlapping rows in different orders, PostgreSQL detects a deadlock and kills one of them.

The iteration order came from `dict.values()` (insertion order in Python 3.7+), which depends on the arrival order of tests/builds, not on any stable sort key. So two concurrent batches that touched overlapping `tree_listing` rows could easily lock them in opposite order, causing deadlocks.

By sorting both writers by the same lock-key columns in the same deterministic order, all transactions now acquire row locks consistently, making deadlocks theoretically impossible.

## Changes

- Sort `checkout_values` in `update_tree_listing()` by the unique constraint columns `(origin, tree_name, git_repository_url, git_repository_branch, git_commit_hash)` before the INSERT ... ON CONFLICT DO UPDATE
- Sort `values` in `_process_tree_listing()` (process_pending_aggregations.py) by the UPDATE WHERE clause columns `(origin, tree_name, git_repository_branch, git_repository_url, git_commit_hash)` before executing the UPDATE

## How to test

1. Run the backend test suite: `docker compose exec test_backend poetry run pytest`
2. Verify no regressions in tree listing aggregation tests

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/tree_listing.py
+++ b/backend/kernelCI_app/management/commands/helpers/tree_listing.py
@@ -1,0 +1,52 @@
+from datetime import datetime
+from typing import Any, NamedTuple
+
+
+class CheckoutRow(NamedTuple):
+    checkout_id: str
+    origin: str
+    tree_name: str | None
+    git_repository_url: str | None
+    git_repository_branch: str | None
+    git_commit_hash: str | None
+    git_commit_name: str | None
+    git_commit_tags: list[str] | None
+    start_time: datetime | None
+
+
+class TreeListingRow(NamedTuple):
+    """Flat row passed to cursor.executemany() for UPDATE tree_listing.
+
+    Field order must match the SQL parameter placeholders in _process_tree_listing.
+    """
+
+    build_pass: int
+    build_failed: int
+    build_inc: int
+    boot_pass: int
+    boot_failed: int
+    boot_inc: int
+    test_pass: int
+    test_failed: int
+    test_inc: int
+    origin: str
+    tree_name: str | None
+    git_repository_branch: str | None
+    git_repository_url: str | None
+    git_commit_hash: str | None
+
+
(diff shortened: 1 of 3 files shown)


backend/kernelCI_app/management/commands/helpers/tree_listing.py
```diff
@@ -0,0 +1,52 @@
...
+    test_failed: int
+    test_inc: int
+    origin: str
+    tree_name: str | None
+    git_repository_branch: str | None
+    git_repository_url: str | None
+    git_commit_hash: str | None
+
+
+def tree_listing_sort_key(v: Any) -> tuple:
```
MarceloRobert, line 39: nit: avoid using `Any`

MarceloRobert (approved):
LGTM

alanpeixinho (approved):
LGTM

PR #1852, opened 2026-04-13
fix: improve icon layout in Section using float-based positioning

### LINKED ISSUE #1433: Move title status icon on details pages
The status icon at the beginning of the title on build/test details make the title have less horizontal space to occupy, which is more noticeable for long titles

Image

dede999:
## Summary         

  - Replaced the flexbox row layout in the Section header with a float-based approach to better handle icon and title positioning.                                                                                                 
  - leftIcon is now rendered as a float-left element with explicit sizing (h-12 w-12), allowing the title text to wrap naturally beside it.
  - rightIcon is now rendered as float-right before the title in the DOM, ensuring it anchors to the right edge correctly.                                                                                                         
  - The container uses overflow-hidden to contain the floated elements.                                                                                                                                                            

##  Motivation                                                                                                                                                                                                                       
                                                                                                                                                                                                                                   
  The previous flex layout caused alignment issues when the icon and title needed to behave more like an inline/text-flow layout. Float-based positioning gives more predictable results for icon+title combinations, especially   
  when the title text wraps across multiple lines.

##  Test plan       

  - Verify Section renders correctly on hardware build details

## Related issue

Closes #1433

### DIFF
--- a/dashboard/src/components/Section/Section.tsx
+++ b/dashboard/src/components/Section/Section.tsx
@@ -134,12 +134,12 @@ const Section = ({
               {leftIcon}
             </span>
           )}
-          {rightIcon && <span>{rightIcon}</span>}
           {title && (
             <span className="max-w-full text-2xl font-bold break-all">
               {title}
             </span>
           )}
+          {rightIcon && <span className="ml-1 sm:ml-2">{rightIcon}</span>}
         </div>
         {subtitle}
       </div>
--- a/dashboard/src/components/Section/Section.tsx
+++ b/dashboard/src/components/Section/Section.tsx
@@ -128,14 +128,18 @@ const Section = ({
     <div className="text-dim-gray mb-6 flex flex-col gap-4">
       <div className="flex flex-col gap-2">
         {eyebrow && <span className="text-sm">{eyebrow}</span>}
-        <div className="flex flex-row items-center gap-2">
-          {leftIcon}
+        <div className="overflow-hidden">
+          {leftIcon && (
+            <span className="float-left mr-2 -mb-2 [&_svg]:h-8 [&_svg]:w-8">
+              {leftIcon}
+            </span>
+          )}
+          {rightIcon && <span>{rightIcon}</span>}
           {title && (
             <span className="max-w-full text-2xl font-bold break-all">
               {title}
             </span>
           )}
-          {rightIcon}
         </div>
         {subtitle}
       </div>


dede999:
As it's easy to see, when the title is shorter the icon and the text are not aligned with each other, both text and icons are attached to the top of the div

image

image

image
-------

After diminishing the icon size
image

image

dashboard/src/components/Section/Section.tsx
```diff
@@ -128,14 +128,18 @@ const Section = ({
...
         {eyebrow && <span className="text-sm">{eyebrow}</span>}
-        <div className="flex flex-row items-center gap-2">
-          {leftIcon}
+        <div className="overflow-hidden">
+          {leftIcon && (
+            <span className="float-left mr-2 *:text-2xl [&_svg]:h-6 [&_svg]:w-6">
+              {leftIcon}
+            </span>
+          )}
+          {rightIcon && <span className="float-right">{rightIcon}</span>}
```
MarceloRobert: I think the rightIcon doesn't need to float-right, if I see it correctly it's only used in the testDetails statusHistory section, in which case it looks weird after the change (too far apart)

Image

dashboard/src/components/Section/Section.tsx
```diff
@@ -128,14 +128,18 @@ const Section = ({
     <div className="text-dim-gray mb-6 flex flex-col gap-4">
       <div className="flex flex-col gap-2">
         {eyebrow && <span className="text-sm">{eyebrow}</span>}
-        <div className="flex flex-row items-center gap-2">
-          {leftIcon}
+        <div className="overflow-hidden">
+          {leftIcon && (
+            <span className="float-left mr-2 *:text-2xl [&_svg]:h-6 [&_svg]:w-6">
```
MarceloRobert: `text-2xl` is already a default as the inspector shows
```suggestion
            <span className="float-left mr-2 [&_svg]:h-6 [&_svg]:w-6">
```

MarceloRobert: I'm not sure why the icon is getting so much space at the bottom, but we can increase it's size and keep it within one line with negative margin
```suggestion
            <span className="float-left mr-2 mb-[-8px] *:text-2xl [&_svg]:h-8 [&_svg]:w-8">
```

dashboard/src/components/Section/Section.tsx
dede999: Thanks for the suggestions. I have applied them all

dashboard/src/components/Section/Section.tsx
```diff
@@ -128,14 +128,18 @@ const Section = ({
...
         {eyebrow && <span className="text-sm">{eyebrow}</span>}
-        <div className="flex flex-row items-center gap-2">
-          {leftIcon}
+        <div className="overflow-hidden">
+          {leftIcon && (
+            <span className="float-left mr-2 -mb-2 [&_svg]:h-8 [&_svg]:w-8">
+              {leftIcon}
+            </span>
+          )}
+          {rightIcon && <span>{rightIcon}</span>}
```
MarceloRobert: I think the rightIcon should be reverted to its place

Image

alanpeixinho (approved):
Nice work.
But I believe this PR would be better classified as a fix, than a feat.

dashboard/src/components/Section/Section.tsx
```diff
@@ -128,14 +128,18 @@ const Section = ({
...
+              {leftIcon}
+            </span>
+          )}
           {title && (
             <span className="max-w-full text-2xl font-bold break-all">
               {title}
             </span>
           )}
-          {rightIcon}
+          {rightIcon && <span className="ml-1 sm:ml-2">{rightIcon}</span>}
```
MarceloRobert, line 142: I don't see the need for the different margin size, but ok

MarceloRobert (approved):
Fixed the problem now

PR #1853, opened 2026-04-14
feat: implement fallback when tree_tests_rollup doesn't have data

gustavobtflores:
## Description

Adds fallback logic to treeDetailsSummaryView to handle cases where the pre-aggregated tree_tests_rollup table has not yet been populated for a checkout. When rollup data is empty, the endpoint now falls back to the legacy row-by-row query approach to ensure test/boot summary data is still returned.

## Changes

- Added fallback condition in BaseTreeDetailsSummary.get() to detect empty rollup data and trigger legacy query
- Added _sanitize_test_rows_legacy() method to process legacy test/boot rows (modeled after BaseTreeDetails._sanitize_rows() but without build handling since builds are already processed separately)
- Added logging with origin, tree name, and git URL for better debugging when fallback is triggered

### DIFF
--- a/backend/kernelCI_app/helpers/treeDetails.py
+++ b/backend/kernelCI_app/helpers/treeDetails.py
@@ -478,12 +478,14 @@ def process_boots_summary(instance, row_data: dict[str, Any]) -> None:
     instance.boot_summary_typed.labs[test_lab].increment(test_status)
 
 
-def process_filters(instance, row_data: dict) -> None:
+def process_filters(instance, row_data: dict, skip_build_filters: bool = False) -> None:
+    """skip_build_filters is used on some calls to this function
+    where we know that the build filters are already processed"""
     issue_id = row_data["issue_id"]
     issue_version = row_data["issue_version"]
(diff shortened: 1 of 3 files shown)


backend/kernelCI_app/helpers/treeDetails.py
```diff
@@ -478,12 +478,12 @@ def process_boots_summary(instance, row_data: dict[str, Any]) -> None:
     instance.boot_summary_typed.labs[test_lab].increment(test_status)
 
 
-def process_filters(instance, row_data: dict) -> None:
+def process_filters(instance, row_data: dict, skip_build_filters: bool = False) -> None:
```
MarceloRobert, line 481: nit: add a comment for what the `skip_build_filters` is for

MarceloRobert (approved):
LGTM, seems to have fixed some issues such as this [mainline checkout](https://staging.dashboard.kernelci.org/tree/mainline/master/014441d1e4b2f5be728433b761df3377dee7a487)

PR #1854, opened 2026-04-14
feat: commit plotting mechanism update

dede999:
## Summary

-  Add sliding window navigation (4 commits visible at a time) to the commit history graph, allowing users to browse older builds beyond the initial 6-commit batch
- Accumulate data as the user navigates backwards, fetching additional pages from the API using cursor-based anchoring — no backend changes required
- Prevent chart flash and loading skeleton when fetching older pages by keeping buffered data visible during background fetches

## Details

The Build History graph previously showed only the latest 6 commits with no way to access older ones. This change introduces:
- Sliding window: Displays 4 commits at a time, with Older/Newer buttons for single-step navigation and First/Last buttons for jumping to the extremes
- Accumulative buffer: Data from multiple API calls is merged into a single chronological buffer, deduplicated by commit hash. New pages are fetched on demand when the user navigates past the oldest buffered commit
- Stable UX: The chart remains visible during background fetches (no loading skeleton flicker), and navigation buttons are disabled while a fetch is in flight to prevent double-triggers

## How to test
- Load a tree details page — graph shows 4 most recent commits with the current commit visible
- Click "Older" repeatedly — window slides left one commit at a time
- Navigate past the initial 6 commits — new page is fetched transparently, chart does not flash
- Click "First" — jumps to the oldest known commits (fetches if needed)
- Click "Last" — returns to the newest commits
- Switch tabs (builds/boots/tests) — graph resets to the latest window
- Verify buttons are disabled appropriately at both ends and during fetches

## Related Issue 

Closes #1840


dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
```diff
@@ -107,7 +163,131 @@ const CommitNavigationGraph = ({
...
+      return;
+    }
+    lastMergedAnchorRef.current = effectiveAnchor;
+
+    if (data.length < BACKEND_PAGE_SIZE) {
+      setExhausted(true);
+    }
+
+    // Data from API is newest-first; reverse to get chronological (oldest-first)
+    const newCommits = [...data].reverse();
```
alanpeixinho: nice touch, more sane

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
```diff
@@ -107,7 +163,131 @@ const CommitNavigationGraph = ({
...
+    const prev = allCommitsRef.current;
+
+    if (prev.length === 0) {
```
alanpeixinho: do we really need to have a special case for when prev is empty?
cant we just make prev an empty list, and let the usual flow of merging?

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
```diff
@@ -107,7 +163,131 @@ const CommitNavigationGraph = ({
...
+      return;
+    }
+
+    // Deduplicate: only prepend commits not already in the buffer
+    const existingHashes = new Set(prev.map(c => c.git_commit_hash));
+    const uniqueNew = newCommits.filter(
+      c => !existingHashes.has(c.git_commit_hash),
+    );
+
+    if (uniqueNew.length === 0) {
```
alanpeixinho: again, if no new unique elements are added, do we need to have a special case for it?

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
```diff
@@ -107,7 +163,131 @@ const CommitNavigationGraph = ({
...
+      setWindowEnd(w => w - 1);
+    }
+  }, [windowStart, exhausted, allCommits]);
+
+  const goOldest = useCallback(() => {
+    if (exhausted) {
+      setWindowEnd(WINDOW_SIZE - 1);
+    } else if (allCommits.length > 0) {
+      // Flag the merge effect to position window at oldest end after fetch
+      jumpToOldestRef.current = true;
```
alanpeixinho: if we are more than one page of distance from the oldest, this would need to recursively trigger the fetches, making this a slow fetching, and also more complicated.
I think we better keep only single page distance navigation, unless we do some changes on the backend to better suit this kind of navigation.
What you think of it?

alanpeixinho: If we do need faster navigation, a simpler solution right now, could be updating the end in steps of 3 (or 5)

alanpeixinho: Go newest seems to be safe to keep. And can help if you navigated to deep in the abyss.

MarceloRobert:
On Hold while @alanpeixinho works on changing the commitHistory endpoint for performance. His refactor will change the way the frontend requests the data, which will change the logic for the commit pagination

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
```diff
@@ -123,7 +156,63 @@ const CommitNavigationGraph = ({
...
+          Math.min(start + NUM_SELECTED_COMMITS, last - NUM_SELECTED_COMMITS),
+          Math.min(end + NUM_SELECTED_COMMITS, last),
+        ] as [number, number],
+    );
+  }, [commitsLength]);
+
+  const goOldest = useCallback(() => {
+    const last = Math.max(commitsLength - 1, 0);
+    setVisibleRange(_ => [last - NUM_SELECTED_COMMITS, last]);
+  }, [commitsLength]);
```
gustavobtflores, line 270: `goOlder`, `goOldest` and `canGoOlder` are capping by commitsLength - 1, what is causing the oldest commit to be hidden when we have > 6 commits since `.slice` excludes the end parameter

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
gustavobtflores: nit: not related to the changes, but as we are touching this file, I would refactor this section to be an object and move outside the component

dashboard/src/components/CommitNavigationGraph/CommitNavigationGraph.tsx
alanpeixinho: nice catch

gustavobtflores (approved):
LGTM

PR #1856, opened 2026-04-15
fix: handle invalid origin when navigating between pages

gustavobtflores:
## Description

Fixes the origin selector appearing blank and causing empty backend responses when navigating between Hardware and Tree pages. The issue occurred because the `origin` search parameter is global, but Hardware pages use `test_origins` while Tree pages use `checkout_origins`. When a selected origin from one page type didn't exist in the other, the selector showed nothing and the backend returned no data.

## Changes

- Extract `pageOrigins` into a separate `useMemo` to compute page-specific origins
- Update `selectItems` to use the pre-computed `pageOrigins`
- Expand the origin validation `useEffect` to check if the current origin is invalid for the page and auto-correct to `DEFAULT_ORIGIN`

## How to test

1. Go to the Hardware page and select an origin that exists in `test_origins` but not in `checkout_origins`
2. Navigate to the Trees page
3. Verify the origin selector now shows a valid fallback origin instead of being blank
4. Verify tree data loads correctly instead of showing empty results if there are results for the origin

Closes #1841

### DIFF
--- a/dashboard/src/components/TopBar/TopBar.tsx
+++ b/dashboard/src/components/TopBar/TopBar.tsx
@@ -41,15 +41,16 @@ const OriginSelect = ({
     [navigate],
   );
 
-  const selectItems = useMemo(() => {
-    if (originData === undefined) {
-      return <></>;
+  const pageOrigins = useMemo(() => {
+    if (!originData) {
+      return [];
     }
-
-    const pageOrigins = isHardwarePath
+    return isHardwarePath
       ? originData.test_origins
       : originData.checkout_origins;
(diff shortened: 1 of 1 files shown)


alanpeixinho (approved):

MarceloRobert (approved):
I see that the treeListing still makes a request with the invalid origin, so I think the best thing would be for the listings (tree and hardware) to check if the origin is valid before making the requests. But either way, a request to an invalid origin will be really fast, and this change ultimately fixes the problem so I'll approve and leave this as a technical debt

PR #1861, opened 2026-04-16
feat: add environment variables to allow configuration of django persistent connections

### LINKED ISSUE #1862: Enable Django persistent database connections for production
## Description

The backend currently defaults `CONN_MAX_AGE` to `0`, meaning every HTTP request opens a new PostgreSQL connection and closes it when the request finishes. In production, where gunicorn runs with 5 sync workers handling steady traffic, this creates unnecessary connection churn and latency overhead on every request.

Django supports persistent connections via `CONN_MAX_AGE`, which keeps connections open and reuses them across requests within the same worker. Combined with `CONN_HEALTH_CHECKS`, this avoids reusing stale connections after database restarts.

## References

- [Django docs: Persistent connections](https://docs.djangoproject.com/en/6.0/ref/databases/#persistent-connections)
- [Django docs: CONN_MAX_AGE](https://docs.djangoproject.com/en/6.0/ref/settings/#std-setting-CONN_MAX_AGE)

gustavobtflores:
## Description

Add configurable Django persistent database connections to reduce connection overhead in production environments. This allows database connections to be reused across requests instead of opening and closing a connection for each request.

## Changes

- Add `DB_CONN_MAX_AGE` environment variable to configure connection persistence (defaults to `0` for development safety)
- Add `DB_CONN_HEALTH_CHECKS` environment variable to verify connections before reuse (defaults to `False`)

Closes #1862

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,10 @@ DB_PASSWORD=CHANGE_ME
 DB_ENGINE=django.db.backends.postgresql
 DB_OPTIONS_CONNECT_TIMEOUT=16
 
+# Django persistent connections (production): enable for better performance
+# DB_CONN_MAX_AGE=60          # Seconds to keep connections open (0 = close after each request)
+# DB_CONN_HEALTH_CHECKS=True  # Verify connection is alive before reuse (recommended with persistent connections)
+
 # Optional: separate application database user/name for setup-dashboard-db.sh.
 # Defaults to DB_USER / DB_NAME if not set.
 # APP_DB_USER=dashboard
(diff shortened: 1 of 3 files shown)


tales-aparecida:
Consider how cronjobs interact with this. I recall hearing a history of a bug in CKI about it, but I don't know exactly what happened and why, so just think about it

gustavobtflores:
> Consider how cronjobs interact with this. I recall hearing a history of a bug in CKI about it, but I don't know exactly what happened and why, so just think about it

If a cronjob runs frequently enough to keep the connection open, that could be a problem. However, as far as I know, we don't have any running that frequently, so it shouldn't be an issue.

MarceloRobert (approved):
LGTM, looked at the docs and seems to be good

PR #1863, opened 2026-04-16
fix: Improve performance on TestStatusHistory endpoint

### LINKED ISSUE #1131: Improve test status history query

alanpeixinho:
* Moving to a raw SQL query, instead of ORM.
  * Force the query to use a hash based comparison, instead of a slower nested loop

    Closes #1131

### DIFF
--- a/backend/kernelCI_app/queries/test.py
+++ b/backend/kernelCI_app/queries/test.py
@@ -1,9 +1,9 @@
 from typing import Optional
 
-from django.db import connection
+from django.db import connection, transaction
 
+from kernelCI_app.cache import get_query_cache, set_query_cache
 from kernelCI_app.helpers.database import dict_fetchall
-from kernelCI_app.models import Tests
 from kernelCI_app.typeModels.databases import (
     Origin,
     Test__StartTime,
@@ -64,33 +64,66 @@ def get_test_status_history(
     field_timestamp: Timestamp,
     group_size: int,
 ):
-    query = Tests.objects.filter(
-        path=path,
-        build__checkout__origin=origin,
-        build__checkout__git_repository_url=git_repository_url,
-        build__checkout__git_repository_branch=git_repository_branch,
-        build__config_name=config_name,
-    ).values(
-        "start_time",
-        "id",
-        "status",
-        "build__checkout__git_commit_hash",
-    )
+    cache_key = "TestStatusHistory"
+    params = {
+        "path": path,
+        "origin": origin,
+        "git_repository_url": git_repository_url,
+        "git_repository_branch": git_repository_branch,
+        "config_name": config_name,
+        "group_size": group_size,
+        "field_timestamp": field_timestamp,
+        "platform": platform,
+        "test_start_time": test_start_time,
+    }
+
+    if rows := get_query_cache(key=cache_key, params=params):
+        return rows
 
     if platform is None:
-        query = query.filter(environment_misc__platform__isnull=True)
+        platform_clause = "AND T.ENVIRONMENT_MISC ->> 'platform' IS NULL"
     else:
-        query = query.filter(environment_misc__platform=platform)
+        platform_clause = "AND T.ENVIRONMENT_MISC ->> 'platform' = %(platform)s"
 
     if test_start_time is None:
         if field_timestamp is None:
-            query = query.filter(
-                start_time__isnull=True,
-                field_timestamp__isnull=True,
-            )
+            time_clause = "AND T.START_TIME IS NULL AND T._TIMESTAMP IS NULL"
+            order_clause = "ORDER BY T._TIMESTAMP DESC"
         else:
-            query = query.filter(field_timestamp__lte=field_timestamp)
-        return query.order_by("-field_timestamp")[:group_size]
+            time_clause = "AND T._TIMESTAMP <= %(field_timestamp)s"
+            order_clause = "ORDER BY T._TIMESTAMP DESC"
     else:
-        query = query.filter(start_time__lte=test_start_time)
-        return query.order_by("-start_time")[:group_size]
+        time_clause = "AND T.START_TIME <= %(test_start_time)s"
+        order_clause = "ORDER BY T.START_TIME DESC"
+
+    query = f"""
+        SELECT
+            T.START_TIME,
+            T.ID,
+            COALESCE(T.STATUS, 'NULL') AS status,
+            C.GIT_COMMIT_HASH AS build__checkout__git_commit_hash
+        FROM
+            TESTS T
+        INNER JOIN BUILDS B ON T.BUILD_ID = B.ID
+        INNER JOIN CHECKOUTS C ON B.CHECKOUT_ID = C.ID
+        WHERE
+            T.PATH = %(path)s
+            AND C.ORIGIN = %(origin)s
+            AND C.GIT_REPOSITORY_URL = %(git_repository_url)s
+            AND C.GIT_REPOSITORY_BRANCH = %(git_repository_branch)s
+            AND B.CONFIG_NAME = %(config_name)s
+            {platform_clause}
+            {time_clause}
+        {order_clause}
+        LIMIT %(group_size)s;
+    """
+
+    with transaction.atomic():
+        with connection.cursor() as cursor:
+            # do not let postgres planner use a slow nested loop in the query
+            # a hashed approach should always be faster here
+            cursor.execute("SET LOCAL enable_nestloop = off")
+            cursor.execute(query, params)
+            rows = dict_fetchall(cursor)
+            set_query_cache(key=cache_key, params=params, rows=rows)
+            return rows
--- a/backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
+++ b/backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
@@ -1,9 +1,8 @@
-from unittest.mock import patch
+from unittest.mock import Mock, patch
 
 from kernelCI_app.queries.test import get_test_details_data, get_test_status_history
 from kernelCI_app.tests.unitTests.queries.conftest import (
     setup_mock_cursor,
-    setup_mock_test_queryset,
 )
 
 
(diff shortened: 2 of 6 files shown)


backend/kernelCI_app/views/testStatusHistoryView.py
```diff
@@ -88,6 +89,10 @@ def get(self, request: HttpRequest) -> Response:
             status_history=status_history_response
         )
 
+        status_history_response = [
+            TestStatusHistoryItem(**item) for item in status_history_response
+        ]
```
MarceloRobert: This validation will already happen in the `TestStatusHistoryResponse` validation below, no need to loop through everything again

backend/kernelCI_app/typeModels/testDetails.py
```diff
@@ -69,12 +69,10 @@ class TestDetailsResponse(BaseModel):
...
-    start_time: Test__StartTime
-    id: Test__Id
-    status: Test__Status
-    git_commit_hash: Checkout__GitCommitHash = Field(
-        validation_alias="build__checkout__git_commit_hash"
-    )
+    start_time: datetime
+    id: str
+    status: str
+    git_commit_hash: str
```
MarceloRobert: apart from `id` all the other fields can still be `None`, no?

backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
MarceloRobert, line 1: some patches are unused, prefer either a `called_once` or a `not_called` assert in those cases (or just don't patch, not everything has to be patched)

MarceloRobert (commented):
Looks good, only some small comments

backend/kernelCI_app/views/testStatusHistoryView.py
alanpeixinho: isnt it expecting an TestStatusHistoryItem?
This loop is not exactly validating, we are converting the dictionary to a DTO.

backend/kernelCI_app/typeModels/testDetails.py
alanpeixinho: I can change that, but should any of them be None?

backend/kernelCI_app/views/testStatusHistoryView.py
MarceloRobert: `status_history_response` already is a `list[dict]` because of the `dict_fetchall`, so the `TestStatusHistoryResponse` will validate and already be returned, so I don't think there's a need for the `TestStatusHistoryItem` either way

backend/kernelCI_app/typeModels/testDetails.py
MarceloRobert: I think `status` could be defaulted to `NULL`, but other than I don't see a problem with them being `None`, it's just how the database is setup.

`git_commit_hash` would be a second place in this, but I still don't think it will affect the page/response that much, if a test is not related to an existing checkout we can't do anything about it.

backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
alanpeixinho: all patches here are covering some IO operation, I kept them, but included a check for calls in all cases.

backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
```diff
@@ -108,7 +185,42 @@ def test_get_test_status_history_with_null_timestamps(self, mock_tests_model):
...
+        mock_set_cache.assert_called_once()
+        mock_transaction.assert_called_once()
+
+    @patch("kernelCI_app.queries.test.set_query_cache")
+    @patch("kernelCI_app.queries.test.get_query_cache")
+    @patch("kernelCI_app.queries.test.dict_fetchall")
+    @patch("kernelCI_app.queries.test.connection")
+    def test_get_test_status_history_returns_cached_result(
+        self,
+        mock_connection,
```
MarceloRobert: nit: mock_connection still unused

backend/kernelCI_app/typeModels/testDetails.py
```diff
@@ -69,12 +69,10 @@ class TestDetailsResponse(BaseModel):
...
-    start_time: Test__StartTime
-    id: Test__Id
-    status: Test__Status
-    git_commit_hash: Checkout__GitCommitHash = Field(
-        validation_alias="build__checkout__git_commit_hash"
-    )
+    start_time: Optional[datetime]
+    id: str
+    status: Optional[str]
+    git_commit_hash: Optional[str]
```
MarceloRobert: if reverting to `optional[type]` then we can keep using the value that was here before. We are just using these types (such as `Test_StartTime`) to signal that these values come directly from the database

MarceloRobert: I thought you were going to change the `status` field though, to default to `"NULL"` when it is `None`

backend/kernelCI_app/tests/unitTests/queries/test_queries_test.py
alanpeixinho: good point

backend/kernelCI_app/typeModels/testDetails.py
alanpeixinho: I have changed back to use the database types. But in there, `Test__Status` is an Optional[Literal["FAIL", "PASS", "SKIP", "ERROR", "MISS", "DONE"]], so the null value is represented as None.
Changing that could impact all uses of this type.

MarceloRobert (approved):
LGTM, tested and it seems all good

backend/kernelCI_app/queries/test.py
```diff
@@ -64,33 +64,66 @@ def get_test_status_history(
...
-                field_timestamp__isnull=True,
-            )
+            time_clause = "AND T.START_TIME IS NULL AND T.FIELD_TIMESTAMP IS NULL"
+            order_clause = "ORDER BY T.FIELD_TIMESTAMP DESC"
         else:
-            query = query.filter(field_timestamp__lte=field_timestamp)
-        return query.order_by("-field_timestamp")[:group_size]
+            params["field_timestamp"] = field_timestamp
+            time_clause = "AND T.FIELD_TIMESTAMP <= %(field_timestamp)s"
+            order_clause = "ORDER BY T.FIELD_TIMESTAMP DESC"
```
gustavobtflores: It shouldn't be `T._TIMESTAMP` instead of `T.FIELD_TIMESTAMP` here?

backend/kernelCI_app/queries/test.py
```diff
@@ -64,33 +64,66 @@ def get_test_status_history(
...
+    """
+
+    with transaction.atomic():
+        with connection.cursor() as cursor:
+            # do not let postgres planner use a slow nested loop in the query
+            # a hashed approach should always be faster here
+            cursor.execute("SET LOCAL enable_nestloop = off")
+            cursor.execute(query, params)
+            rows = dict_fetchall(cursor)
+            set_query_cache(key=cache_key, params=params, rows=rows)
```
gustavobtflores, line 128: I would like to see some `EXPLAIN ANALYZE` with this turned on and off attached in the PR, just to ensure we aren't missing anything here

backend/kernelCI_app/queries/test.py
alanpeixinho: Good catch, I think both the frontend and tests do dot give us this case.
I will fix and include some tests to catch it.

backend/kernelCI_app/queries/test.py
alanpeixinho: For the regular queries (which are already fast) the benefit, if it exists should be minimal (with some chance of even being slightly worse).
I am including one of the slowest queries, where the benefit can be seen. The main problem is that the regular query offered by the planner is bad in its worst case scenarios.

Example obtained with [this query](http://staging.dashboard.kernelci.org/test/maestro%3A69ef87bb9b5a9683090c4fd5)

|||
|---|---|
|Before|[analyze-nested.md](https://github.com/user-attachments/files/27143913/analyze-nested.md)|
|After|[analyze-no-nested.md](https://github.com/user-attachments/files/27143916/analyze-no-nested.md)|

alanpeixinho: More regular case from [this query](http://staging.dashboard.kernelci.org/test/maestro%3A69ee97b99b5a968309072a0b)

|||
|---|---|
|Before|[analyze-nested-small.md](https://github.com/user-attachments/files/27143978/analyze-nested-small.md)|
|After|[analyze-no-nested-small.md](https://github.com/user-attachments/files/27143985/analyze-no-nested-small.md)|

gustavobtflores (approved):
LGTM

PR #1864, opened 2026-04-16
feat: add a backoff time whenever process_pending deadlock

### LINKED ISSUE #1868: process_pending_aggregations fails on PostgreSQL deadlocks without retry
The `process_pending_aggregations` management command occasionally fails when PostgreSQL detects a deadlock during aggregation batch writes. When this happens, the entire aggregation batch fails immediately without any retry mechanism, causing data processing interruptions.

gustavobtflores:
## Description
Add retry backoff handling to process_pending_aggregations so transient PostgreSQL deadlocks do not fail the aggregation batch immediately.

## Changes
- Catch deadlock-related database errors during tree listing aggregation writes
- Retry the transaction with exponential backoff before giving up
- Restructure batch transaction boundaries to reduce transaction lock time

Closes #1868

### DIFF
--- a/backend/kernelCI_app/management/commands/process_pending_aggregations.py
+++ b/backend/kernelCI_app/management/commands/process_pending_aggregations.py
@@ -9,7 +9,9 @@ from typing import Literal, Optional, Sequence, TypedDict, Union
 from django.conf import settings
 from django.core.management.base import BaseCommand
 from django.db import connection, transaction
+from django.db.utils import OperationalError
 from prometheus_client import Counter, start_http_server
+from psycopg.errors import DeadlockDetected
 
 from kernelCI_app.constants.general import MAESTRO_DUMMY_BUILD_PREFIX
 from kernelCI_app.constants.ingester import PROMETHEUS_MULTIPROC_DIR
@@ -36,6 +38,12 @@ AGGREGATION_RECORDS_WRITTEN = Counter(
     "aggregation_records_written_total",
     "Total number of records written to destination tables",
     ["table"],  # values: "tree_listing", "hardware_status", "processed_items"
+)
+
+DEADLOCK_RETRIES_TOTAL = Counter(
+    "aggregation_deadlock_retries_total",
+    "Total number of deadlock retries",
+    ["table"],  # values: "tree_listing", "hardware_status"
 )
 
 
@@ -525,6 +533,11 @@ def _fetch_test_issues(test_ids: list[str]) -> dict[str, dict]:
 
 
 class Command(BaseCommand):
+    # WARNING: This command is designed for single-worker execution only.
+    # Running multiple concurrent workers is not safe: select_for_update(skip_locked=True)
+    # releases row locks when Transaction 1 commits, but pending rows are not deleted until
+    # Transaction 2. In that window a second worker can claim and process the same rows,
+    # causing double-counting in tree_listing aggregations.
     help = """
         Process pending tests for hardware status aggregation,
         checking corresponding builds and checkouts in the database.
@@ -1075,6 +1088,7 @@ class Command(BaseCommand):
         test_builds_by_id: dict[str, Builds],
         ready_builds: Sequence[PendingBuilds],
         build_checkouts_by_id: dict[str, Checkouts],
+        max_retries: int = int(os.getenv("PROCESS_PENDING_MAX_RETRIES", "5")),
     ) -> None:
         tree_listing_data, new_processed_entries_tree = aggregate_tree_listing(
             ready_tests,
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -1127,27 +1148,28 @@ def process_pending_batch(self, batch_size: int) -> int:
...
-                        ready_tests,
-                        test_builds_by_id,
-                        ready_builds,
-                        build_checkouts_by_id,
-                    )
+            if ready_tests or ready_builds:
+                self._process_tree_listing_batch(
+                    ready_tests,
+                    test_builds_by_id,
+                    ready_builds,
```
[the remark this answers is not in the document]
gustavobtflores: The current flow is only safe under a single-worker assumption, because the select_for_update(skip_locked=True) locks are released before _process_tree_listing_batch() runs and before the pending rows are deleted.

In our deployment we only run one worker for process_pending_aggregations, so this does not currently create duplicate processing in practice. I've added an inline code comment in process_pending_aggregations.py to make that limitation explicit.

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -1075,15 +1083,33 @@ def _process_tree_listing_batch(
         test_builds_by_id: dict[str, Builds],
         ready_builds: Sequence[PendingBuilds],
         build_checkouts_by_id: dict[str, Checkouts],
+        max_retries: int = 5,
```
alanpeixinho: I think this would be better as an env var, if for some reason, we need to keep trying longer

alanpeixinho (approved):

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -1075,15 +1083,33 @@ def _process_tree_listing_batch(
...
+                with transaction.atomic():
+                    self._process_tree_listing(tree_listing_data)
+                    self._process_new_processed_entries(new_processed_entries_tree)
+                break
+            except OperationalError as e:
+                if (
+                    isinstance(e.__cause__, DeadlockDetected)
+                    and attempt < max_retries - 1
+                ):
+                    wait = random.uniform(0.5, min(30, 2 ** (attempt + 2)))
```
MarceloRobert: question: why do we need it to be `random`? I understood the logic of waiting exponentially with a cap at 30 seconds, but not why it's random between 0.5 and the cap

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -1127,27 +1153,28 @@ def process_pending_batch(self, batch_size: int) -> int:
...
+                self._process_tree_listing_batch(
+                    ready_tests,
+                    test_builds_by_id,
+                    ready_builds,
+                    build_checkouts_by_id,
+                )
 
+            with transaction.atomic():
                 tests_count += self._delete_ready_tests(ready_tests=ready_tests)
                 builds_count += self._delete_ready_builds(ready_builds=ready_builds)
```
MarceloRobert, line 1172: I see that splitting the transactions is better for deadlock, but should we separate the `delete_ready_x` like this? I'm considering that we only want to delete those items if we successfully updated the listing rows, but if we separate the transactions we could be deleting the rows without them being properly consumed I think, or am I overlooking something here?

backend/kernelCI_app/management/commands/process_pending_aggregations.py
gustavobtflores: It doesn’t need it, it's just another layer to prevent ingester and process_pending from conflicting with each other. It doesn’t hurt, though.

backend/kernelCI_app/management/commands/process_pending_aggregations.py
gustavobtflores: If any of the transactions fail before, this would never run, so the delete_ready_x wouldn't run anyway, seems safe

MarceloRobert: ah, because it would cause an exception. got it

backend/kernelCI_app/management/commands/process_pending_aggregations.py
```diff
@@ -1075,15 +1083,33 @@ def _process_tree_listing_batch(
         test_builds_by_id: dict[str, Builds],
         ready_builds: Sequence[PendingBuilds],
         build_checkouts_by_id: dict[str, Checkouts],
+        max_retries: int = int(os.getenv("PROCESS_PENDING_MAX_RETRIES", "5")),
```
MarceloRobert, line 1091: nit: environment variables should be documented somewhere

MarceloRobert (approved):
LGTM

Issue #1845, opened 2026-04-09
Replace flake8 and black with Ruff


MarceloRobert:
As following the discussion in https://github.com/kernelci/kernelci-project/issues/566, we should replace our current setup of black + flake8 with ruff. Ruff should be checked pre-commit and in the CI as well.

PR #1865, opened 2026-04-17
Refactor: ruff formatting

MarceloRobert:
This PR is only meant for applying all the ruff styling changes, so that we can later ignore the single squash-merge commit from the blame history using the [.git-blame-ignore-revs file](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-filefile)

If we had included these changes with the configuration, the squash-merge would combine everything (refactor and configs) into a single commit, and ignoring the revisions from that commit would ignore the config changes, which is not correct.

## Changes
- Applied auto-formatting
- Applied auto-linting
- Applied manual linting changes for lint rules with unsafe fixes

## How to test
This is just a refactor, the entire dashboard should still be working normally. There was no change to the frontend.

Since this PR doesn't include the ruff configs, it's going to fail all CI, but you can check that everything is right by going to the ruff config PR and running the ruff commands there.

Part of #1845

### DIFF
--- a/backend/conftest.py
+++ b/backend/conftest.py
@@ -1,8 +1,9 @@
-import tomllib
-from pytest import Item
 import os
+import tomllib
+
 import django
 from django.conf import settings
+from pytest import Item
 
 
 def pytest_addoption(parser):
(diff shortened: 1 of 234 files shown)


alanpeixinho (approved):
Tested via frontend, all services seems to be correct and with the expected performance.

gustavobtflores (approved):

PR #1866, opened 2026-04-17
Chore: batch update frontend dependencies

MarceloRobert:
## Changes
- Updates some frontend dependencies, with the focus on high-priority security updates

## How to test
These updates should not change the dashboard behavior, so navigate around the site and test components to see if they are still normal. The backend had no changes.

Closes dependabot alerts 98, 99, 100, 101, 102, 103, 104, 105, 106?, 110, 111, 118, 122, 123, 135. It also might close moderate to low-priority ones of the same libraries that were updated.


alanpeixinho (approved):
Testing the frontend, everything seems in place

gustavobtflores (approved):
Looks good, tested frontend and everything seems okay

PR #1870, opened 2026-04-22
fix: include builds from all matching checkouts in tree details queries

### LINKED ISSUE #1869: Tree details page returns empty results when multiple checkout rows exist for the same commit
## Problem

The tree details endpoints (`/api/tree/<hash>/summary` and `/api/tree/<tree>/<branch>/<hash>/summary`) return empty results for certain trees even though:

- The tree listing page correctly shows test/build counts
- Data exists in the checkouts, builds, tests, and tree_tests_rollup tables
- The origin query parameter is correctly set

## Root Cause

The get_tree_details_builds() and related queries use a RELEVANT_CHECKOUTS CTE with ORDER BY c._timestamp DESC LIMIT 1 to find the matching checkout. This limits the query to only the most recent checkout row for the given commit hash.

However, in production there can be multiple checkout rows for the same commit. When LIMIT 1 picks a checkout row that doesn't have associated builds, the tree details page appears empty even though builds exist under a different checkout row for the same commit.

gustavobtflores:
Fixes an issue where the tree details page returns empty results when multiple checkout rows exist for the same commit hash. The query was previously limited to only the most recent checkout (`LIMIT 1`), potentially missing builds associated with other checkout rows.

## Changes

- Removed `LIMIT 1` from the `RELEVANT_CHECKOUTS` CTE to return all matching checkouts
- Changed the build join from `b.checkout_id = rc.checkout_id` to `b.checkout_id IN (SELECT checkout_id FROM RELEVANT_CHECKOUTS)` to aggregate builds from all matching checkout rows

## Related Issue

Closes #1869

### DIFF
--- a/backend/kernelCI_app/queries/tree.py
+++ b/backend/kernelCI_app/queries/tree.py
@@ -801,7 +801,6 @@ def get_tree_details_builds(
                 AND c.origin = %(origin_param)s
             ORDER BY
                 c._timestamp DESC
-            LIMIT 1
         )
         SELECT
             b.id AS build_id,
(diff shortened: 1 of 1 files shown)


alanpeixinho (approved):
All good

MarceloRobert (approved):
Builds are fixed, only some tests are still not the same count but ok

PR #1873, opened 2026-04-22
feat: Add option to dump and load database snapshots, while keeping relationships on foreign tables.

### LINKED ISSUE #1884: Add database snapshot and restore functionality with foreign key preservation
Implement a command to dump and load database snapshots while preserving relationships on foreign tables. This replaces the previous direct DB-to-DB migration flow and enables safer data migration and backup workflows.

alanpeixinho:
Include a command to dump and load database snapshots.

## Summary
- Add `snapshot` and `restore` subcommands to `update_db` management command, replacing the previous direct DB-to-DB migration flow
- Snapshots are stored as `.tar.gz` archives containing per-table CSV files, preserving foreign key relationships on restore

1. Populate a dabase with `poetry run python3 manage.py seed_test_data --clear --yes`, or use an existing database.
2. Run the snapshot command, with desired parameters, as in `poetry run python3 manage.py update_db snapshot --start-interval "100 days" --end-interval "0 minute" --filepath ./my-snapshot.tar.gz`
3. Check that data is properly saved inside tables on `my-snapshot.tar.gz`.
4. Clear the database with `run python3 manage.py seed_test_data --clear --yes --checkouts 0 --issues 0`
5. Run the restore command with desired parameters, as `poetry run python3 manage.py restore snapshot --start-interval "100 days" --end-interval "0 minute" --filepath ./my-snapshot.tar.gz`
6. Check that the data on the snapshot was properly inserted in the database.

Closes #1884


backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -161,6 +209,46 @@ def handle(
...
+                    self.restore_builds()
+                case "tests":
+                    self.restore_tests()
+                case "incidents":
+                    self.restore_incidents()
+                case _:
+                    self.stdout.write(
+                        self.style.ERROR(
+                            f"""Unknown table '{table}'.
+                                Valid options are: issues, checkouts, builds, tests, incidents."""
```
MarceloRobert: You could make a template string and reuse between commands. You can save something like `myvar = "table is %(tablename)s"` and then use as `print(myvar % {"tablename": "issues"})`

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -203,7 +291,7 @@ def select_issues_data(self) -> list[tuple]:
             self.end_interval,
         ] + self.origins
 
-        with self.kcidb_connection.cursor() as kcidb_cursor:
+        with connection.cursor() as kcidb_cursor:
```
MarceloRobert: prefer `connections['default']`

MarceloRobert: also other occurrences in the file

MarceloRobert (commented):
looks ok, I'll test later

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: It is a copy/paste from the existing error message. But I plan to refactor regardless

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Coming back to it.
Given we are dealing with a single data source, what would be the benefit of using named connections?

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -51,46 +77,62 @@ def __init__(self, *args, **kwargs):
...
+                "--filepath",
+                type=str,
+                required=True,
+                help="Path to store/load the snapshot file.",
+            )
+
+    def _invalid_table_error(self, table: str) -> str:
+        return (
+            f"Unknown table '{table}'.\n"
+            "\tValid options are: issues, checkouts, builds, tests, incidents."
```
MarceloRobert, line 153: there's not really a need to make a function for this when you could use template strings but it doesn't matter either, just commenting.

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -51,46 +77,62 @@ def __init__(self, *args, **kwargs):
...
+                type=lambda s: [origin.strip() for origin in s.split(",")],
+                help="Limit database changes to specific origins (comma-separated list)."
+                " If not provided, any origin will be considered",
+                default=[],
+            )
+            p.add_argument(
+                "--filepath",
+                type=str,
+                required=True,
+                help="Path to store/load the snapshot file.",
```
MarceloRobert: does the file need to have a specific forma like .csv? I think it would be good to increment the help text here.

MarceloRobert (commented):
it worked for a small interval, but for an interval of 1 day I got an error of 
```
Processed 1610 tests records: 1.53 MiB
Error updating database: read
CommandError: Command failed
```
I assume it's because it went over the size limit of 32MiB?

Also left some other comments

gustavobtflores:
I think we could remove https://github.com/kernelci/dashboard/blob/main/backend/scripts/copy_db_data.sh as it no longer makes sense after the changes on the command

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -129,38 +173,77 @@ def handle(
...
+        finally:
+            self.snapshot_archive.close()
+
+    def restore(self, table: str, snapshot_filepath: str):
+        self.snapshot_archive = tarfile.open(snapshot_filepath, "r:*")
 
+        try:
+            match table:
+                case None:
+                    self.restore_incidents()
```
gustavobtflores: Wouldn't be better to restore incidents last here? `restore_incidents` calls `insert_incidents_data` that checks for existing issues, builds and tests and could skip incidents incorrectly

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Included information on the help, as well as a coercion to the correct extension.

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: If it is okay with you I would prefer to keep a function, it is easier to debug and include breakpoints on errors.

alanpeixinho:
> it worked for a small interval, but for an interval of 1 day I got an error of
> 
> ```
> Processed 1610 tests records: 1.53 MiB
> Error updating database: read
> CommandError: Command failed
> ```
> 
> I assume it's because it went over the size limit of 32MiB?
> 
> Also left some other comments

I have managed to find the problem, when spilling the 32MiB, the temporary file is moved to disk, and this was causing the later reads to fail.
The best solution is to actually open the file with read and write permissions.
But I also included a little more buffer (64MiB) to avoid (for most of the cases) the spilling.

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Good catch. Done.

backend/kernelCI_app/management/commands/update_db.py
MarceloRobert: > Given we are dealing with a single data source, what would be the benefit of using named connections?

It's mainly as a suggestion from django itself. The definition for `connection` has a comment above it that says "`# For backwards compatibility. Prefer connections['default'] instead.`". IIUC it's the same thing but it's more explicit and therefore more organized, especially if an app eventually has multiple connections.

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -51,46 +78,62 @@ def __init__(self, *args, **kwargs):
...
+            "snapshot", help="Save snapshot of database into file."
         )
-        parser.add_argument(
-            "--related-data-only",
-            action="store_true",
-            help="""Only retrieves data that are related to the existing data.
-                This allows to follow foreign key constraints,
-                but it almost certainly won't retrieve all data in the given interval.""",
+        restore_parser = command_parser.add_parser(
+            "restore", help="Load snapshot file onto database (WRITE)."
```
MarceloRobert: nit: saying `(WRITE)` doesn't sound very explicit to me
```suggestion
            "restore", help="Load snapshot file onto database (will INSERT into db)."
```

backend/kernelCI_app/management/commands/update_db.py
```diff
@@ -51,46 +78,62 @@ def __init__(self, *args, **kwargs):
...
+                type=str,
+                help="Start interval for filtering data ('x days' or 'x hours' format)",
+                required=True,
+            )
+            p.add_argument(
+                "--end-interval",
+                type=str,
+                help="End interval for filtering data ('x days' or 'x hours' format)",
+                required=True,
+            )
```
MarceloRobert: nit: `--start-interval` and `--end-interval` doesn't seem to make a difference for the restore command. I don't think there's a need to add a feature to "filter the snapshot file so that it only inserts data between the interval", I think you should just remove this parameter from the `restore` subparser, but it's a nitpick

MarceloRobert (approved):
Seems to be all working, just left some nits

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Fair. I agree it is better to just restore it all.

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Fair.

gustavobtflores (approved):

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Done

backend/kernelCI_app/management/commands/update_db.py
alanpeixinho: Done

MarceloRobert (approved):
Tested again and is working well

PR #1877, opened 2026-04-23
fix: Improve performance for the TreeCommitHistory endpoint

### LINKED ISSUE #1883: Optimize TreeCommitHistory endpoint database queries for better performance
The TreeCommitHistory endpoint currently has suboptimal database queries that may cause performance issues, especially on large trees. We need to optimize the database queries to improve response times.

alanpeixinho:
## Summary
  - Optimize `TreeCommitHistory` endpoint with improved database queries for better performance
  - Add new `TreeCommitsListView` endpoint for future paginated tree commit listing
  - Update frontend `CommitNavigationGraph` component and API integration
## How to Test
1. Verify backend endpoints:
   - GET /api/tree/{tree_id}/commits-history/ - optimized commit history
   - GET /api/tree/{tree_id}/commits-list/ - commit listing
2. Frontend:
   - Navigate to a tree details page
   - Verify CommitNavigationGraph renders correctly with commit history for builds, boots and tests.
   - Check that commit list view loads properly.
4. Performance:
   - Compare response times for TreeCommitHistory endpoint before/after changes on large trees.

Closes #1883

### DIFF
--- a/backend/kernelCI_app/queries/tree.py
+++ b/backend/kernelCI_app/queries/tree.py
@@ -7,6 +7,10 @@ from kernelCI_app.cache import get_query_cache, set_query_cache
 from kernelCI_app.helpers.database import dict_fetchall
 from kernelCI_app.helpers.treeDetails import create_checkouts_where_clauses
 from kernelCI_app.models import Checkouts
+from kernelCI_app.queries.duration import (
+    get_boot_test_duration_clause,
+    get_build_duration_clause,
+)
 from kernelCI_app.utils import get_query_time_interval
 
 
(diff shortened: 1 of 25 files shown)


backend/kernelCI_app/queries/tree.py
```diff
@@ -895,6 +896,198 @@ def _create_selected_checkouts_clause(
     return selected_checkouts_clause
 
 
+def get_tree_commits(
+    *,
+    origin: Optional[str],
+    git_url: Optional[str],
+    git_branch: Optional[str],
+    tree_name: Optional[str],
```
MarceloRobert, line 907: `tree_name` and `git_branch` can be null but there's no query condition in the query if they are None

backend/kernelCI_app/queries/tree.py
```diff
@@ -895,6 +896,198 @@ def _create_selected_checkouts_clause(
...
+            b.architecture,
+            b.config_name,
+            lab
+    """
+
+    boot_filter = ""
+    if include_boots and not include_tests:
+        boot_filter = "\nAND t.path LIKE 'boot%%'"
+    elif include_tests and not include_boots:
+        boot_filter = "\nAND t.path NOT LIKE 'boot%%'"
```
MarceloRobert: the boot filter should be `= boot` or `like boot.%`. This is because a test called "bootloader_something" is not considered a boot test, only those with the specific name "boot" or "subtests" of boot

backend/kernelCI_app/typeModels/treeCommits.py
```diff
@@ -63,6 +63,44 @@ class TreeCommitsQueryParameters(DirectTreeCommitsQueryParameters):
...
+        None, description=DocStrings.TREE_NAME_PATH_DESCRIPTION
+    )
+    git_branch: Optional[str] = Field(
+        None, description=DocStrings.TREE_COMMIT_GIT_BRANCH_DESCRIPTION
+    )
+    git_url: Optional[str] = Field(
+        None, description=DocStrings.TREE_COMMIT_GIT_URL_DESCRIPTION
+    )
+    commit_hashes: list[str] = Field(
+        None, description="Comma-separated list of commit hashes to get history for"
```
MarceloRobert: should move this string to the DocStrings file

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -538,3 +547,207 @@ class TreeCommitsHistory(BaseTreeCommitsHistory):
...
+            )
+
+        results = self.aggregate_commits(commit_hashes, instances)
+
+        try:
+            valid_response = TreeCommitsResponse(
+                root=[results[commit_hash] for commit_hash in commit_hashes]
+            )
+        except ValidationError as e:
+            print(e)
```
MarceloRobert: either remove the print or use `log_message`

MarceloRobert (commented):
Some filters are not working \[in treeDetails\]:
- test hardware - 0 results
- boot origin - not applying
- test inconclusive status - missing data
- boot inconclusive status - missing data
- build inconclusive status - 0 results / missing data
- test issue (uncategorized) - 0 results
- build issue (real) - 0 results
- boot issue (any) - 0 results

I'm also getting error 400 on trees without tree_name or git_branch. Maybe we could just fallback to the old endpoint in those cases.

Also also, the hardwareDetails page requests the commitHistory when there's a single tree selected but it is returning error 400 for me

backend/kernelCI_app/queries/tree.py
alanpeixinho: Nice catch

backend/kernelCI_app/queries/tree.py
alanpeixinho: My idea was to not apply filtering when not specified, do we have cases, where we want to search specifically for null branchs and tree names?

alanpeixinho:
> Some filters are not working [in treeDetails]:
> 
> * test hardware - 0 results
> * boot origin - not applying
> * test inconclusive status - missing data
> * boot inconclusive status - missing data
> * build inconclusive status - 0 results / missing data
> * test issue (uncategorized) - 0 results
> * build issue (real) - 0 results
> * boot issue (any) - 0 results
> 
> I'm also getting error 400 on trees without tree_name or git_branch. Maybe we could just fallback to the old endpoint in those cases.
> 
> Also also, the hardwareDetails page requests the commitHistory when there's a single tree selected but it is returning error 400 for me

The 400 returns was a choice for empty cases. Is it preferrable to return 200 and an empty list?

backend/kernelCI_app/queries/tree.py
MarceloRobert: It's not common but there are cases of trees without name/branch. Ex.: [broonie tree](https://dashboard.kernelci.org/tree/05a44cdbb5ca716eb8ae89058c702837992dc996?o=broonie&p=t&tri%7Cc=&tri%7Cch=05a44cdbb5ca716eb8ae89058c702837992dc996), even though the current commitHistory is not right here either, it at least lets the user navigate through the commits

MarceloRobert:
> The 400 returns was a choice for empty cases. Is it preferrable to return 200 and an empty list?

For the cases that I have seen the error 400, they should actually be returning data, not be empty. As for when they _should_ be empty, then I guess we have been using status 200 with an error message, but it can be up for debate again; my preference would be the 200 with "error" message (I know it's not a real error but the field in the response is called `error`)

backend/kernelCI_app/views/treeCommitsHistory.py
alanpeixinho: It is not necessary to log here, just a debugging slip

backend/kernelCI_app/typeModels/treeCommits.py
alanpeixinho: Done

backend/kernelCI_app/queries/tree.py
alanpeixinho: Done

backend/kernelCI_app/queries/tree.py
alanpeixinho: I am adding an option to search for exactly NULL tree names and branches.
What I dislike is that it makes this parameters no longer trully an optional filter, as null is now just another id

MarceloRobert: That's how we have to work with this database, unfortunately. I think you added the filter to the `get_tree_commit_history_hashes_aggregated` query but not for the `get_tree_commits` query

backend/kernelCI_app/queries/tree.py
MarceloRobert: Both conditions should apply at all times, so it should be
```
if include_boots and not include_tests:
    boot_filter = "\nAND (t.path ='boot' OR t.path LIKE 'boot.%%')"
elif include_tests and not include_boots:
    boot_filter = "\nAND t.path NOT LIKE 'boot.%%' AND t.path != 'boot'"
```

alanpeixinho: You are absolutely right. My oversight.

gustavobtflores: This change is still missing, no?

alanpeixinho: Yes, this change was still local. It should have been updated now.

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -538,3 +548,218 @@ class TreeCommitsHistory(BaseTreeCommitsHistory):
...
+            params.builds_related_to_filtered_tests_only
+        )
+
+        include_types = params.types
+        if params.types == ["builds"] and (
+            self.builds_related_to_filtered_tests_only or len(self.filterHardware) > 0
+        ):
+            include_types = ["builds", "boots", "tests"]
+
+        instances = get_tree_commit_history_hashes_aggregated(
```
gustavobtflores: nit: I think we could rename instances here to something more meaningful

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -538,3 +548,218 @@ class TreeCommitsHistory(BaseTreeCommitsHistory):
...
+        lab: str,
+        compiler: str,
+        architecture: str,
+        compatibles: set[str],
+        status: str,
+        known_issues: set[str],
+        is_build: bool,
+        is_boot: bool,
+        is_test: bool,
+    ) -> bool:
```
gustavobtflores, line 576: nit: `hardware_id` and `compatibles` aren't being used

backend/kernelCI_app/views/treeCommitsHistory.py
alanpeixinho: Done

backend/kernelCI_app/views/treeCommitsHistory.py
alanpeixinho: Good catch. Done

gustavobtflores (approved):
Looks good

PR #1878, opened 2026-04-24
Fix: path search on treeDetails

### LINKED ISSUE #1872: String based searches for issues aren't substring based
If I type a substring of a test name (eg, "regmap") into the path name search box for tests my expectation is that it will match as a substring. However it appears that the search is actually matching top level elements of the path, and only matching within the top level group (eg, "rt" matches "rt-tests" but nothing beyond the first . matches). I would have expected to match on any part of the test name so a search for "crc32" woudl match any test with that string in the path.

MarceloRobert:
Fixes an issue where a path filter would make the rollup rows filter the entire path group instead of only filtering individual tests. ~~For the solution, legacy rows have to be used since the aggregated rows don't have information about individual tests.~~ **We opted for a solution where the test path filter only applies to the table in the frontend, without making changes in the backend so that we don't fallback to the legacy query and the path filtering becomes faster too.**

The tests were being filtered correctly, but the summary was returning empty and was causing all the frontend components to be replaced with a "no data" too.

## Changes
- Removed test path and boot path from the mapping that ultimately goes to the backend request
- Altered the frontend testsTable and bootsTable so that they use the table state for the path filtering

Since these changes are in the core component, they work in the treeDetails page, hardwareDetails page, and buildDetails page (that has a testTable).

## How to test
Check any treeDetails page and try applying a test or boot path filter that only works on a subset of a group (ex "nfs" for "boot.nfs" or "setup" for "tast.setup"). Then compare with the production state.

Closes #1872

### DIFF
--- a/backend/kernelCI_app/views/treeDetailsSummaryView.py
+++ b/backend/kernelCI_app/views/treeDetailsSummaryView.py
@@ -265,6 +265,14 @@ class BaseTreeDetailsSummary(APIView):
             issues_dict=self.test_issues_dict
         )
 
+    def _should_fallback_to_legacy_tests(self) -> bool:
+        """Rollup rows cannot evaluate path filters with test-level precision.
+
+        tree_tests_rollup stores only `path_group` (first path segment). When a
+        path filter is present, summary correctness requires raw test rows.
+        """
+        return bool(self.filters.filterTestPath or self.filters.filterBootPath)
+
     def get(
         self,
         request: HttpRequest,
@@ -308,14 +316,24 @@ class BaseTreeDetailsSummary(APIView):
 
         self._sanitize_builds_rows(builds_rows)
 
-        if rollup_rows:
+        should_fallback_to_legacy = self._should_fallback_to_legacy_tests()
+
+        # We only use the rollup rows if they exist and we have no other constraints
+        if rollup_rows and not should_fallback_to_legacy:
             self._sanitize_rollup_rows(rollup_rows)
         else:
-            out(
-                f"Rollup data empty for commit {commit_hash} "
-                f"(origin={origin_param}, tree={tree_name}, git_url={git_url_param}), "
-                f"falling back to legacy query"
-            )
+            if should_fallback_to_legacy:
+                out(
+                    f"Path filter detected for commit {commit_hash} "
+                    f"(origin={origin_param}, tree={tree_name}, git_url={git_url_param}), "
+                    "falling back to legacy query for accurate test/boot summary"
+                )
+            else:
+                out(
+                    f"Rollup data empty for commit {commit_hash} "
+                    f"(origin={origin_param}, tree={tree_name}, git_url={git_url_param}), "
+                    f"falling back to legacy query"
+                )
             legacy_rows = get_tree_details_data(
                 origin_param=origin_param,
                 git_url_param=git_url_param,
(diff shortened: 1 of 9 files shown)


MarceloRobert:
Moved to draft because we discussed about changing the behavior of the filter to not apply the path filter on the summary endpoint, which won't lose a lot of functionality while increasing the performance massively

alanpeixinho:
When typing a search, anytime the filter is applied we loose focus on the search bar.
This is not a problem when we know the exact search string, but when we are trying to build a search string from partial filter, re-selecting the search bar is undesireable.

path_filter

backend/kernelCI_app/views/treeDetailsSummaryView.py
```diff
@@ -308,14 +316,24 @@ def get(
 
         self._sanitize_builds_rows(builds_rows)
 
-        if rollup_rows:
+        should_fallback_to_legacy = self._should_fallback_to_legacy_tests()
+
+        # We only use the rollup rows if they exist and we have no other constraints
+        if rollup_rows and not should_fallback_to_legacy:
```
alanpeixinho, line 322: nit: could we do an early return on empty rollup rows? I believe it could simplify this conditional

dashboard/src/components/Tabs/FilterList.tsx
```diff
@@ -15,6 +16,11 @@ export const createFlatFilter = (filter: TFilter): string[] => {
   const flatFilter: string[] = [];
 
   Object.entries(filter).forEach(([field, fieldValue]) => {
+    // Type guard + don't show filters outside of the expected keys
+    if (!isTFilterObjectKeys(field) && !isTFilterNumberKeys(field)) {
```
alanpeixinho: nit: can we create a helper function to check if it is a valid key of any type?

gustavobtflores:
Since we aren't filtering summary by path anymore, I would remove the cards with the filters, it causes some visual confusion for users

MarceloRobert:
> Since we aren't filtering summary by path anymore, I would remove the cards with the filters, it causes some visual confusion for users

My reasoning is that those filters stay applied even when switching tabs, so they could stay there (as a card) to represent this. However I was able to hide them by having a list of "ignored filters" that won't show up in the cards, what do you think?

backend/kernelCI_app/views/treeDetailsSummaryView.py
MarceloRobert: I didn't quite understand what you mean

alanpeixinho: Feel free to ignore, it is just a nit. But moving this logic to a separate function, we can simplify the conditional using early returns.
Something like
```
def _handle_test_data(self, rollup_rows: list, params: dict):
    should_fallback = self._should_fallback_to_legacy_tests()
    
    if rollup_rows and not should_fallback:
        self._sanitize_rollup_rows(rollup_rows)
        return

    ctx = f"(origin={params['origin_param']}, tree={params['tree_name']}, git_url={params['git_url_param']})"
    if should_fallback:
        out(f"Path filter detected for commit {params['commit_hash']} {ctx}, falling back to legacy query.")
    else:
        out(f"Rollup data empty for commit {params['commit_hash']} {ctx}, falling back to legacy query.")

    legacy_rows = get_tree_details_data(**params)
    if legacy_rows:
        self._sanitize_test_rows_legacy(legacy_rows)
```

But again, it is okay as it is.

alanpeixinho (approved):
Great Job

gustavobtflores (approved):
LGTM

PR #1880, opened 2026-04-27
Fix: topbar page title

### LINKED ISSUE #1855: Two issue links giving different output
https://dashboard.kernelci.org/issue/maestro%3A67d434ff5d6b71886d45efe758ca00ff45cb969a?iv=1
https://dashboard.kernelci.org/issue/maestro%3Ac970827f049cfd67124148e3c19abe1d4aa8347a?iv=1

One link says **_Trees_** One link says **_Issues_**

It doesn't happen from a direct link, it happens through the source you come through

Image

MarceloRobert:
Fixes the page title for the topbar, using only the route instead of the redirect state + route

## Changes
- Removes the "redirectFrom" state and uses only the route path for the page title in the topBar

## How to test
Go from treeDetails to build/test/issueDetails page and see that even though the breadcrumb is present, the title is the current page (instead of staying at "Tree" like the old behavior). The same happens for hardwareDetails. Compare with the old behavior on staging/production.

## Example
Before:
image

After:
image

Tested on [localhost issue](http://localhost:5173/issue/maestro%3Ada694c56147298d223ee432ad8d6a8ee311b773a?iv=1) using submissions data.

Closes #1855

### DIFF
--- a/dashboard/src/components/TopBar/TopBar.tsx
+++ b/dashboard/src/components/TopBar/TopBar.tsx
@@ -1,11 +1,6 @@
 import { FormattedMessage } from 'react-intl';
 
-import {
-  useSearch,
-  useNavigate,
-  useRouterState,
-  useMatches,
-} from '@tanstack/react-router';
+import { useSearch, useNavigate, useMatches } from '@tanstack/react-router';
 
 import { useCallback, useEffect, useMemo, useState, type JSX } from 'react';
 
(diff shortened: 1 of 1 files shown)


alanpeixinho (approved):
Tested on local database. OK.

gustavobtflores (approved):
LGTM

PR #1889, opened 2026-05-05
Fix: change frontend cache time

MarceloRobert:
The frontend query cache can stale some requests, which in turn can show outdated numbers between treeListing, treeDetails and treeCommitHistory. Lowering the cache time will technically increase the number of requests, but this can be done without much concern since we have improved endpoint performance. Besides, without this risk for outdated numbers there is less chance for confusion in users when they see trees that are receiving data in real time.

## Changes
- Changed frontend request cache from 5 minutes to 30 seconds

Also removes eslint rule ignore

## How to test
Go to any frontend page, check that if you make the same request within 30s it will get the data from cache (it won't perform the request again) and after 30s it will redo the request.

Closes #1890


alanpeixinho (approved):
LGTM

gustavobtflores (approved):

PR #1895, opened 2026-05-12
workflows: add dashboard team management file

victor-accarini:
To add and remove members of the dashboard team we use a specific membership file, this way it's easier to onboarding and offboard new members.

Github has a policy that only members of an organization can be part of a team, this change automates that invitation and team setup.

### DIFF
--- a/.github/dashboard-team
+++ b/.github/dashboard-team
@@ -1,0 +1,5 @@
+nuclearcat
+bhcopeland
+victor-accarini
+alanpeixinho
+gustavobtflores
--- a/.github/dashboard-team
+++ b/.github/dashboard-team
@@ -3,3 +3,4 @@ bhcopeland
 victor-accarini
 alanpeixinho
 gustavobtflores
+tales-aparecida
--- a/.github/scripts/sync-dashboard-team.js
+++ b/.github/scripts/sync-dashboard-team.js
@@ -1,0 +1,198 @@
+#!/usr/bin/env node
+
+const fs = require("fs");
+const { Octokit } = require("@octokit/rest");
+const path = require("path");
+
+const octokit = new Octokit({
+  auth: process.env.GITHUB_TOKEN,
+});
+
+const ORG = process.env.ORG;
+const TEAM_SLUG = process.env.TEAM_SLUG;
+const DRY_RUN = process.env.DRY_RUN === "true";
+
+const TEAM_FILE = path.join(
+  __dirname,
+  "..",
+  "dashboard-team",
+);
+
+const PROTECTED_USERS = ["nuclearcat", "bhcopeland", "victor-accarini"];
+
+const USER_RE = /^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/i;
+
+function validateEnv() {
+  const required = ["ORG", "TEAM_SLUG", "GITHUB_TOKEN"];
+
+  const missing = required.filter((v) => !process.env[v]);
+
+  if (missing.length > 0) {
+    throw new Error(`Missing required env vars: ${missing.join(", ")}`);
+  }
+}
+
+function readDesiredUsers() {
+  if (!fs.existsSync(TEAM_FILE)) {
+    throw new Error(`Team file not found: ${TEAM_FILE}`);
+  }
+
+  const content = fs.readFileSync(TEAM_FILE, "utf8");
+
+  const users = [
+    ...new Set(
+      content
+        .split("\n")
+        .map((line) => line.trim().toLowerCase())
+        .filter((line) => line && !line.startsWith("#")),
+    ),
+  ];
+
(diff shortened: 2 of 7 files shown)


.github/dashboard-team
```diff
@@ -0,0 +1,4 @@
+nuclearcat
+victor-accarini
+alanpeixinho
+gustavobtflores
```
bhcopeland, line 5: Can you add me too?

bhcopeland: Also all commits should have a signed off tag. But I noticed you guys don't do that, is that something we can change going forward?

victor-accarini: Done

victor-accarini: For the workflow to work we need to set the ORG_ADMIN_TOKEN secret, or if we already have another valid secret I can update to use that instead

.github/scripts/sync-dashboard-team.js
```diff
@@ -0,0 +1,202 @@
...
+
+const PROTECTED_USERS = ["nuclearcat", "bhcopeland", "victor-accarini"];
+
+const USER_RE = /^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/i;
+
+function validateEnv() {
+  const required = ["ORG", "TEAM_SLUG"];
+
+  if (!DRY_RUN) {
+    required.push("GITHUB_TOKEN");
```
alanpeixinho: even for a dry run, I believe we still need to provide the GITHUB_TOKEN, better to always check for it

.github/scripts/sync-dashboard-team.js
```diff
@@ -0,0 +1,202 @@
...
+        .split("\n")
+        .map((line) => line.trim().toLowerCase())
+        .filter((line) => line && !line.startsWith("#")),
+    ),
+  ];
+
+  const invalidUsers = users.some((u) => !USER_RE.test(u));
+  if (invalidUsers) {
+    throw new Error(
+      `Found an invalid GitHub username in array: [${users.join(", ")}]`,
```
alanpeixinho: As we are not providing the reason a username is invalid, we should at least filter to show here only the list of invalid usernames.

alanpeixinho (approved):
Tested on dry run only. But it is okay.

.github/scripts/sync-dashboard-team.js
victor-accarini: Yep, you are right, updated

.github/scripts/sync-dashboard-team.js
victor-accarini: Done

.github/scripts/sync-dashboard-team.js
victor-accarini: I also update the CODEOWNERS file to ensure we can't change the script behavior without one of the owners knowing @bhcopeland

alanpeixinho: > I also update the CODEOWNERS file to ensure we can't change the script behavior without one of the owners knowing @bhcopeland

Good point

.github/workflows/sync-dashboard-team.yaml
```diff
@@ -0,0 +1,91 @@
...
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/dashboard-team"
+      - ".github/scripts/sync-dashboard-team.js"
+      - ".github/scripts/package.json"
+      - ".github/scripts/package-lock.json"
+      - ".github/workflows/sync-dashboard-team.yml"
```
gustavobtflores: I think the extension should be .yaml here

```suggestion
      - ".github/workflows/sync-dashboard-team.yaml"
```

.github/workflows/sync-dashboard-team.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+  pull_request:
+    paths:
+      - ".github/dashboard-team"
+      - ".github/scripts/sync-dashboard-team.js"
+      - ".github/scripts/package.json"
+      - ".github/scripts/package-lock.json"
+      - ".github/workflows/sync-dashboard-team.yml"
```
gustavobtflores: ditto

```suggestion
      - ".github/workflows/sync-dashboard-team.yaml"
```

gustavobtflores (commented):
I think we should also have a .gitignore for the /scripts node_modules

tales-aparecida (approved):

gustavobtflores (approved):

.github/dashboard-team
```diff
@@ -0,0 +1,5 @@
+nuclearcat
+bhcopeland
+victor-accarini
+alanpeixinho
+gustavobtflores
```
tales-aparecida, line 5: Can I be added as well? :)

gustavobtflores: Done

PR #1897, opened 2026-05-12
fix: move development Nginx configuration to another directory

gustavobtflores:
## Description

Keep the dev-only nginx template out of the staging/default proxy image so the proxy does not try to load a `dashboard_dev` upstream outside the development compose stack.

## How to test

1. Run `docker compose build proxy`.
2. Run `docker compose up -d proxy`.
3. Confirm the proxy logs no longer show `Running envsubst on /etc/nginx/templates/dev.conf.template`.
4. Confirm the proxy no longer fails with `host not found in upstream "dashboard_dev"`.
5. For the dev stack, run `docker compose -f docker-compose.dev.yml up -d proxy` and confirm the Vite proxy path still uses the dev template.


alanpeixinho (approved):
LGTM

PR #1902, opened 2026-05-13
feat/search commit intent filter hardware

alanpeixinho:
Include an option to do commit searches on the search bar of the hardware listing page.

## How to test
  * Open the hardware listing page and type a valid commit hash, it should list include only the status of the specified commit.
  * When using regular text (not containing any valid commit hashes), the regular text search should take place.
  * In the case of using both, the page should join the behaviours, listing the selected commits, and applygin textual search for the extra text.
  
  Closes #1900

### DIFF
--- a/dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
+++ b/dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
@@ -24,10 +24,13 @@ import { useHardwareDetailsCommitHistory } from '@/api/hardwareDetails';
 
 import type {
   CommitHead,
+  CommitHistory,
   CommitHistoryTable,
   PreparedTrees,
   HardwareTrees,
 } from '@/types/hardware/hardwareDetails';
+
+import { parseSearchIntent } from '@/lib/intent';
 
 import MemoizedCompatibleHardware from '@/components/Cards/CompatibleHardware';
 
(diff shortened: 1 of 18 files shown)


backend/kernelCI_app/tests/unitTests/queries/hardware_queries_test.py
```diff
@@ -36,6 +36,30 @@ def test_get_hardware_listing_data_success(self, mock_connection):
...
+            commits_list=["a" * 40],
+        )
+
+        sql, exec_params = mock_cursor.execute.call_args[0]
+        assert "selected_checkouts AS" in sql
+        assert "JOIN selected_checkouts AC ON b.checkout_id = AC.id" in sql
+        assert "ANY(%(commits_list)s)" in sql
+        assert "git_commit_tags" in sql and "&&" in sql
+        assert "SELECT DISTINCT" not in sql
+        assert exec_params["commits_list"] == ["a" * 40]
```
gustavobtflores: I don't feel that these tests are valuable, tbh

dashboard/src/pages/hardwareDetails/HardwareDetails.tsx
```diff
@@ -88,27 +119,41 @@ const prepareTreeItems = ({
...
-    const result: PreparedTrees = {
+    const rows = commitHistoryData?.[treeIdentifier] ?? [];
+    const excludeFromCommitIntentDefaultSelection =
+      !!commitIntentTokens?.length &&
+      !isCommitHistoryDataLoading &&
+      !findMatchingCheckout(rows, commitIntentTokens) &&
+      !commitIntentTokens.some(t =>
+        matchesCommitToken(
+          tree.head_git_commit_hash ?? '',
+          tree.head_git_commit_tags,
```
gustavobtflores: The field sent from the backend is `head_git_commit_tag` not `head_git_commit_tags`

alanpeixinho: Thank you, it was a typo

backend/kernelCI_app/tests/unitTests/queries/hardware_queries_test.py
alanpeixinho: I agree with that.

gustavobtflores (approved):

PR #1904, opened 2026-05-15
feat: add checkout selectors on hardware page

gustavobtflores:
Allows the user to select a specific tree/branch/commit on the hardware page to filter data

Closes #1901


backend/kernelCI_app/queries/hardware.py
```diff
@@ -179,6 +179,134 @@ def get_hardware_listing_data(
...
+                    qr.git_repository_url,
+                    qr.git_repository_branch
+            ) AS branch_latest_start_time
+        FROM
+            qualified_revisions qr
+        ORDER BY
+            qr.tree_name ASC,
+            branch_latest_start_time DESC,
+            qr.latest_start_time DESC
+    """
```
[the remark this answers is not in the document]
gustavobtflores: Kind of a design decision for now, we may look into optimized approaches later on

backend/kernelCI_app/queries/hardware.py
```diff
@@ -179,6 +179,134 @@ def get_hardware_listing_data(
         return cursor.fetchall()
 
 
+def get_hardware_selectors(origin: str) -> list[dict]:
```
alanpeixinho, line 204: Cant we make this faster, if we just fetch the list of trees associated with a single hardware?
And lazily fetch branches and revisions as needed?

dashboard/src/locales/messages/index.ts
```diff
@@ -214,9 +214,20 @@ export const messages = {
     'hardwareDetails.timeFrame':
       'Results from {startDate} and {startTime} to {endDate} {endTime}',
     'hardwareListing.bannerTitle': 'Hardware Listing',
+    'hardwareListing.branchSelectorLabel': 'Branch',
     'hardwareListing.description': 'List of hardware from kernel tests',
     'hardwareListing.notFound': 'No hardware information available',
+    'hardwareListing.revisionCapNote': 'Showing latest 50 revisions',
```
alanpeixinho: are we limiting the list of revisions?

backend/kernelCI_app/views/hardwareByRevisionView.py
```diff
@@ -0,0 +1,90 @@
...
+
+        return hardwares
+
+    @extend_schema(
+        parameters=[HardwareListingByRevisionQueryParamsDocumentationOnly],
+        responses=HardwareListingResponse,
+    )
+    def get(self, request: Request):
+        try:
+            query_params = HardwareListingByRevisionQueryParams(
```
alanpeixinho, line 66: we might have similar endpoints here, from the #1902 , we will have to check if we can reuse them

backend/kernelCI_app/queries/hardware.py
alanpeixinho: In worst case scenario, we could just return a list of all tree/branches, and let the filter do its job and return empty if the user selects a combination not available for the hardware listing.

dashboard/src/locales/messages/index.ts
gustavobtflores: We are not, forgot to remove after some changes

backend/kernelCI_app/queries/hardware.py
gustavobtflores: The problem is that, for origins with a low volume of tests, the user would end up with many empty tree/branch/revision combinations. I think the approach is reasonable while we work on other improvements that will make these queries easier.

backend/kernelCI_app/views/hardwareByRevisionView.py
gustavobtflores: I would prefer to keep them separated for now as they have kind of different meanings, and these endpoints are not being limited by a time-window

dashboard/src/pages/Hardware/HardwareTable.tsx
```diff
@@ -65,10 +74,18 @@ interface IHardwareTable {
...
+  selectors?: HardwareSelectorTree[];
+  selectedTree?: HardwareSelectorTree | null;
+  selectedBranch?: HardwareSelectorBranch | null;
+  selection?: HardwareRevisionSelection | null;
+  onTreeChange?: (nextSelection: HardwareRevisionSelectorValue) => void;
 }
 
 type HardwareListingRoutes = '/hardware' | '/hardware/v1' | '/hardware/v2';
 
+const noop = (): void => {};
```
alanpeixinho: I am usually for the null object pattern, but isnt more common in javascript we use an undefined for "nothing done here"?

dashboard/src/utils/search.ts
```diff
@@ -151,6 +151,10 @@ const generalMinifiedParams: Record<SearchParamsKeys, string> = {
   endTimestampInSeconds: 'et',
   issueVersion: 'iv',
   logOpen: 'l',
+  treeName: 't',
+  gitRepositoryUrl: 'hgu',
```
alanpeixinho: nit: why 'h' for prefixing the minified parameters?

backend/kernelCI_app/queries/hardware.py
```diff
@@ -201,6 +201,134 @@ def get_hardware_listing_data(
...
+    params = {"origin": origin}
+
+    query = """
```
alanpeixinho, line 214: since we are already joining with builds and tests, cant we bring only the revisions associated with the specified hardware we are tracking?

alanpeixinho: I believe this might help as well, pre select only a subset of builds.

```
WITH hwbuilds AS (
    SELECT DISTINCT 
        builds.id, 
        builds.checkout_id,
        tests.environment_misc->>'platform' as platform
    FROM tests
    INNER JOIN builds ON builds.id = tests.build_id 
    WHERE tests.origin = 'maestro' 
      AND tests.start_time > (NOW() - INTERVAL '5 days')
)
SELECT * FROM checkouts 
INNER JOIN hwbuilds ON hwbuilds.checkout_id = checkouts.id;
```

alanpeixinho: Sorry, forgot to retrieve all commits, we need a full scan on checkouts at least.

```
WITH hwbuilds AS (
    SELECT DISTINCT 
        builds.id, 
        builds.checkout_id,
        tests.environment_misc->>'platform' as platform
    FROM tests
    INNER JOIN builds ON builds.id = tests.build_id 
    WHERE tests.origin = 'maestro' 
      AND tests.start_time > (NOW() - INTERVAL '5 days')
),
hwtrees as (select hwbuilds.platform, checkouts.git_repository_url, checkouts.tree_name  FROM checkouts 
INNER JOIN hwbuilds ON hwbuilds.checkout_id = checkouts.id)
select distinct hwtrees.platform, checkouts.git_repository_url, checkouts.tree_name, checkouts.git_repository_branch, checkouts.git_commit_hash
	from checkouts inner join hwtrees on hwtrees.tree_name = checkouts.tree_name and hwtrees.git_repository_url = checkouts.git_repository_url;
```

alanpeixinho: It is just a rough idea what what we could do, it can certainly be improved.

gustavobtflores: But this way we would be limiting by builds tested within the last 5 days, no?

dashboard/src/utils/search.ts
gustavobtflores: I think there was some conflict happening before, but we can remove this

backend/kernelCI_app/queries/hardware.py
alanpeixinho: I am actually selecting hardware tested in the last 5 days (something we are already doing in the listing). From those, we are selecting all trees related with those tests, and all commits from those trees.

dashboard/src/api/hardware.ts
```diff
@@ -57,52 +58,83 @@ export const useHardwareListing = (
...
     },
   );
 
   return data;
 };
 
-export const useHardwareListingV2 = (
-  startTimestampInSeconds: number,
-  endTimestampInSeconds: number,
+export const useHardwareSelectors = (
```
alanpeixinho, line 76: Is it feasible to make this call async?
This way the page could load and it might reduce the perceived latency on the more expensive endpoint.

backend/kernelCI_app/queries/hardware.py
gustavobtflores: Yeah, I was thinking about scenarios where some origins don't have tests for 5 days. In that case, the page could end up looking strangely empty, since we would still have old commits to show. But it's still a nice workaround.

alanpeixinho: Got it. I was under the impression that we were already limited by the trees with checkouts in the last few days.

dashboard/src/api/hardware.ts
gustavobtflores: We can't, cause the table below depends on this selector to be ready before fetching data

backend/kernelCI_app/queries/hardware.py
alanpeixinho: I got a slighlty better query with gemini, going roughly from 36s -> 23s. Mine had some unecessary steps.

```
SELECT DISTINCT
    c.git_repository_url,
    c.tree_name,
    c.git_repository_branch,
    c.git_commit_hash,
    c.git_commit_name,
    c.start_time
FROM checkouts c
INNER JOIN (
    -- Step 1: Find ONLY unique active stream keys first
    SELECT 
        c2.tree_name, 
        c2.git_repository_url
    FROM tests t
    INNER JOIN builds b ON b.id = t.build_id
    INNER JOIN checkouts c2 ON b.checkout_id = c2.id
    WHERE t.origin = 'maestro'
      AND t.start_time > (NOW() - INTERVAL '5 days')
      AND (t.environment_misc -> 'platform') IS NOT NULL
    GROUP BY c2.tree_name, c2.git_repository_url
) active_streams 
  ON c.tree_name = active_streams.tree_name 
 AND c.git_repository_url = active_streams.git_repository_url;
```

backend/kernelCI_app/queries/hardware.py
```diff
@@ -201,6 +201,104 @@ def get_hardware_listing_data(
         return cursor.fetchall()
 
 
+def get_hardware_selectors(origin: str) -> list[dict]:
```
alanpeixinho, line 204: we should include a redis cache

backend/kernelCI_app/views/hardwareSelectorsView.py
```diff
@@ -0,0 +1,97 @@
...
+            query_params = HardwareSelectorsQueryParams(
+                origin=request.GET.get("origin")
+            )
+        except ValidationError as e:
+            return Response(data=e.json(), status=HTTPStatus.BAD_REQUEST)
+
+        selectors_raw = get_hardware_selectors(origin=query_params.origin)
+
+        try:
+            result = self._sanitize_records(selectors_raw=selectors_raw)
```
alanpeixinho, line 93: I am getting some sanitization error for hardware listing
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for HardwareSelectorBranch
git_repository_branch
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
```

alanpeixinho (approved):
Nice work.

PR #1905, opened 2026-05-18
feat: Include hardware registry info into DB via django command

alanpeixinho:
How to test:
   1. Run migrations `./migrate-app-db.sh`. (Using a local database with write permissions)
   2. Run update_hardware_registry `poetry run python3 manage.py update_hardware_registry ti.yaml` command passing an example of valid hardware registry manifest (https://github.com/kernelci/kernelci-pipeline/blob/main/config/hardware_registry/ti.yaml).
   3. The information should be available in the tables hardware_registry_platform_vendors, hardware_registry_platforms, hardware_registry_processors, hardware_registry_silicon_vendors and hardware_registry_system_modules.

Closes #1899


backend/kernelCI_app/management/commands/update_hardware_registry.py
```diff
@@ -0,0 +1,103 @@
...
+                self.style.WARNING(
+                    "Missing typical hardware-registry sections : {', '.join(missing)}"
+                )
+            )
+
+        silicon_vendors = data["silicon_vendors"]
+        platform_vendors = data["platform_vendors"]
+        processors = data["processors"]
+        system_modules = data["system_modules"]
+        platforms = data["platforms"]
```
gustavobtflores: Is it safe to access these keys even if data may be missing?

backend/kernelCI_app/management/commands/update_hardware_registry.py
```diff
@@ -0,0 +1,103 @@
...
+        if not isinstance(data, dict):
+            raise CommandError(
+                f"Expected a YAML mapping at root, got {type(data).__name__}."
+            )
+
+        missing = [k for k in EXPECTED_SECTIONS if k not in data]
+        if missing:
+            self.stderr.write(
+                self.style.WARNING(
+                    "Missing typical hardware-registry sections : {', '.join(missing)}"
```
gustavobtflores: Missing `f` prefix in the string?

backend/kernelCI_app/management/commands/update_hardware_registry.py
alanpeixinho: I believe we could trust this data, as it is a manifest, and not data directly from user, but you are correct, we should be careful.

gustavobtflores (approved):
LGTM

PR #1908, opened 2026-05-18
fix: Use same sort keys for listings of trees on HardwareDetails page.

alanpeixinho:
There were instances of sorting the lists of trees in the hardware
  page using a different set of keys (due to mistake on the performance
  improvements of #1832).
  This case might have not triggered, because the sorting provided by
  the first 2 keys tree_name/branch, might suffice for basically most of
  the comparisons.
  * All of them are now using the keys (in order of importance):
    1. tree_name
    2. branch,
    3. git_repository_url
    4. git_commit_hash
    
    Closes #1907


gustavobtflores (approved):
Looks good

PR #1912, opened 2026-05-20
fix: move secret check from job-level to step-level in sync-dashboard-team

gustavobtflores:
Fix GitHub Actions workflow failure caused by using `secrets.ORG_ADMIN_TOKEN` in job-level `if`.

GitHub Actions does not allow `secrets` context in `jobs.<job_id>.if`, so PR validation now gates only on pull request event. Secret-dependent dry run step checks `env.ORG_ADMIN_TOKEN` instead.

Ref: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-secrets#using-secrets-in-a-workflow

### DIFF
--- a/.github/workflows/sync-dashboard-team.yaml
+++ b/.github/workflows/sync-dashboard-team.yaml
@@ -29,8 +29,10 @@ permissions:
 jobs:
   # Add validation checks for Pull Requests
   validate:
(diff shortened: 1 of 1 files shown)


alanpeixinho (approved):
LGTM

PR #1913, opened 2026-05-20
Expose notification metrics as an endpoint

### LINKED ISSUE #1763: Create data metrics api endpoint
Create an endpoint that returns the same data that is being sent as an email to the mailing list. The metrics in this case refer to the data that is being received (how many new checkouts, builds, tests, regressions, lab tests...), not related to website usage (user requests, status codes, response time...).

Example of the data that should be returned by the endpoint: https://groups.io/g/kernelci-results/message/60587
New format example at #1761

alanpeixinho:
## What is done
* Include /api/metrics that receives start and end day for aggregation
* Concurrent run of metrics queries to alleviate the latency

## How to test
Start backend server.
Test the endpoint /api/metrics, make sure they match the counting on the email notifications (for the specific range dates)

Closes #1763

### DIFF
--- a/backend/kernelCI_app/urls.py
+++ b/backend/kernelCI_app/urls.py
@@ -175,4 +175,5 @@ urlpatterns = [
     path("proxy/", views.ProxyView.as_view(), name="proxyView"),
     path("origins/", views.OriginsView.as_view(), name="originsView"),
     path("tree-report/", views.TreeReport.as_view(), name="treeReportView"),
+    path("metrics/", views.MetricsView.as_view(), name="metricsView"),
 ]
--- a/backend/kernelCI_app/urls.py
+++ b/backend/kernelCI_app/urls.py
@@ -175,5 +175,5 @@ urlpatterns = [
     path("proxy/", views.ProxyView.as_view(), name="proxyView"),
     path("origins/", views.OriginsView.as_view(), name="originsView"),
     path("tree-report/", views.TreeReport.as_view(), name="treeReportView"),
-    path("metrics/", views.MetricsView.as_view(), name="metricsView"),
+    path("metrics/", view_cache(views.MetricsView), name="metricsView"),
 ]
--- a/backend/kernelCI_app/views/metricsView.py
+++ b/backend/kernelCI_app/views/metricsView.py
@@ -1,0 +1,51 @@
+from http import HTTPStatus
+
+from drf_spectacular.utils import extend_schema
+from pydantic import ValidationError
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from kernelCI_app.constants.localization import ClientStrings
+from kernelCI_app.helpers.errorHandling import create_api_error_response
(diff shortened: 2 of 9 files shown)


backend/kernelCI_app/views/metricsView.py
```diff
@@ -0,0 +1,51 @@
...
+            )
+
+        try:
+            data = get_metrics_data(
+                start_days_ago=query_parameters.start_days_ago,
+                end_days_ago=query_parameters.end_days_ago,
+            )
+            valid_response = metrics_report_data_to_response(data)
+        except ValidationError as e:
+            return Response(data=e.json(), status=HTTPStatus.INTERNAL_SERVER_ERROR)
```
gustavobtflores: We shouldn't return an error other than 500 here for validation?

backend/kernelCI_app/urls.py
```diff
@@ -175,4 +175,5 @@ def view_cache(view):
     path("proxy/", views.ProxyView.as_view(), name="proxyView"),
     path("origins/", views.OriginsView.as_view(), name="originsView"),
     path("tree-report/", views.TreeReport.as_view(), name="treeReportView"),
+    path("metrics/", views.MetricsView.as_view(), name="metricsView"),
```
gustavobtflores: Include cache for the endpoint?

backend/kernelCI_app/views/metricsView.py
```diff
@@ -0,0 +1,51 @@
...
+            HTTPStatus.OK: MetricsResponse,
+            HTTPStatus.BAD_REQUEST: dict[str, str],
+        },
+        methods=["GET"],
+    )
+    def get(self, request) -> Response:
+        try:
+            query_parameters = MetricsQueryParameters.model_validate(request.GET.dict())
+        except ValidationError as e:
+            return Response(data=e.json(), status=HTTPStatus.INTERNAL_SERVER_ERROR)
```
gustavobtflores: Same here for validation error

alanpeixinho: Changed to a bad request

backend/kernelCI_app/views/metricsView.py
alanpeixinho: For what I see, for validation errors on response we are usually returning 500 error. And I think it makes some sense. An error in the response could be seen as something wrong with the endpoint.

alanpeixinho: I will, however, keep this try catch to the response validation alone, this way we are more certain it shouldnt be triggered incorrectly.

backend/kernelCI_app/urls.py
alanpeixinho: Done

backend/kernelCI_app/views/metricsView.py
gustavobtflores: Oh, my bad, I was thinking about the request validation, this one is correct

gustavobtflores (approved):

PR #1914, opened 2026-05-21
fix: remove obsolete e2e test from tree listing

gustavobtflores:
After we removed date inputs on tree-listing and hardware-listing in #1898, we no longer need to test these inputs. This PR removes the test from tree-listing e2e tests and fix CI.


alanpeixinho (approved):
LGTM

victor-accarini (approved):

PR #1915, opened 2026-05-21
feat: improve tests table hierarchy nesting

gustavobtflores:
## Description

Improve tests table readability by collapsing empty single-child group chains into a single displayed path segment.

## How to test

1. Open a tests table with deeply nested single-child groups
2. Confirm empty intermediate groups are shown as one combined path segment

Closes #1906


alanpeixinho (approved):
Nice way to solve

PR #1919, opened 2026-05-25
feat: Using stable key on tree lists on Hardware Details Page

alanpeixinho:
## What it is

Changes the default selection tree of hardware details page, from using index based selection, to a stable (hash based) selection.

## How to test

---

1. Open the hardware details page.
2. Select a subset of trees.
3. The query param x should be composed of hash based list of trees, and should match the selection based on index (production).

---
1. Open the hardware details page.
2. include a query parameter to select using old numeric indexes `?x[]=[0,1]`
1. Any url from the previous implementation (using indexes instead of hashes) should still work and lead to equivalent results.

Closes #1909


backend/kernelCI_app/constants/hardwareDetails.py
```diff
@@ -1 +1,9 @@
+import hashlib
+
 SELECTED_HEAD_TREE_VALUE = "head"
+TREE_KEY_HASH_LENGTH = 14
+
+
+def make_tree_key(tree_name: str, branch: str, url: str) -> str:
+    raw = f"{tree_name}|{branch}|{url}"
+    return hashlib.sha256(raw.encode()).hexdigest()[:TREE_KEY_HASH_LENGTH]
```
felipebergamin: shouldn't we consider origin as well?

backend/kernelCI_app/queries/hardware.py
felipebergamin: shouldn't we change the cache key? this will keep old query cached after deploy.

same for get_hardware_trees_data

dashboard/src/api/hardwareDetails.ts
felipebergamin: in your PR body you say the support for numeric indexed trees would be kept. But your second commit removes the support.

PR #1920, opened 2026-05-27
feat: dynamic metrics page

### LINKED ISSUE #1782: Create a dynamic page for the "weekly report"
Take the same schema of data we submit to the mailing list in our weekly report and show with dynamic filtering.

The goal here is to give users a way to get an updated report at any time, with clickable links to see it in more detail.

gustavobtflores:
Adds an interactive Metrics page at `/metrics` for CI coverage and build regression data from `GET /api/metrics/`.

## How to test

1. Open `/metrics` and confirm tables load for 7/14/30-day periods
2. Expand a build regression row and verify new issue details

Closes #1782

### DIFF
--- a/backend/kernelCI_app/queries/notifications.py
+++ b/backend/kernelCI_app/queries/notifications.py
@@ -790,6 +790,42 @@ def get_metrics_data(
     WHERE r.ranked <= 3 AND n.total_incidents > 0
     """
 
+    new_build_issues_query = """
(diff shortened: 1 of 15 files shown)


alanpeixinho (approved):
I think we could improve readability on lists of regressions, they have long strings and it can make listings a little bit dense. Maybe more spacing or better truncation could help.
But functionwise seems to be okay.

PR #1925, opened 2026-06-08
Reports regzbot tracking

bhcopeland:
We need to sanity-check the SQL command that returns a sensible "last good build".

Other than that fixes are pretty simple:

- Cleaned up whitespaces in email
- Dropped broken cc, and updated cc list (as it was bouncing)
- Added regzbot _tags_ into email template


backend/kernelCI/settings.py
```diff
@@ -178,7 +178,7 @@ def get_json_env_var(name, default):
                 "--monitoring-id=notifications_new_issues",
                 "--action=new_issues",
                 "--to=kernelci-results@groups.io",
-                "--cc=gus@collabora.com",
+                "--cc=linux@leemhuis.info, ben.copeland@linaro.org",
```
knurd: Please use regressions@leemhuis.info for me instead.

Or maybe even better: regressions@lists.linux.dev – I monitor that closely, so it's just as good as mailing me directly. And that way other can join in easily if they want to.

backend/kernelCI_app/management/commands/templates/issue_build.txt.j2
```diff
@@ -10,17 +9,25 @@ New build issue found on {{ issue["tree_name"] }}/{{ issue["git_repository_branc
...
+Please include the KernelCI and regzbot tags when submitting a fix:
+
+Reported-by: kernelci.org bot <bot@kernelci.org>
+
+{% if builds and builds[0]["last_pass_commit"] -%}
+#regzbot introduced: {{ builds[0]["last_pass_commit"] }}..{{ issue["git_commit_hash"] }}
+{% endif -%}
+#regzbot title: build regression on {{ issue["tree_name"] }}/{{ issue["git_repository_branch"] }}
+#regzbot link: https://d.kernelci.org/i/{{ issue["id"] }}
+
```
knurd: Without an "#regzbot introduced: …" regzbot won't act (at least for now, but changing that would not be easy), because it needs a commit or an range where the regressions started. So I'd the "{% endif -%}" needs to move down three lines afaics.

backend/kernelCI_app/queries/notifications.py
```diff
@@ -367,6 +367,51 @@ def kcidb_last_test_without_issue(issue, incident):
...
+        FROM builds b
+        JOIN checkouts c ON b.checkout_id = c.id
+        WHERE c.git_repository_url = %(giturl)s
+        AND c.git_repository_branch = %(branch)s
+        AND b.config_name = %(config_name)s
+        AND b.architecture = %(architecture)s
+        AND b.compiler = %(compiler)s
+        AND b.status = 'PASS'
+        AND c.origin = %(origin)s
+        AND b._timestamp >= NOW() - INTERVAL %(interval)s
```
alanpeixinho: might be safer to just use `start_time` instead of `_timestamp`, as it will retrieve all builds compiled in the given interval.

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -222,6 +223,12 @@ def generate_issue_report(
     if issue["build_id"]:
         issue_type = "build"
         incidents = kcidb_build_incidents(issue_id)
+        for incident in incidents:
+            last_build = kcidb_last_build_without_issue(issue, incident)
+            if last_build:
+                incident["last_pass"] = last_build[0]["start_time"]
+                incident["last_pass_commit"] = last_build[0]["git_commit_hash"]
+                incident["last_pass_id"] = last_build[0]["id"]
```
alanpeixinho: could we call it last_pass_build_id? just to be clear on which id we are talking here.

backend/kernelCI_app/management/commands/notifications.py
alanpeixinho, line 586: this call is missing a required parameter group_size that limits how far in history we go. We might need to either make this an optional parameter (allowing to return full history) or pass a value large enough here.

backend/kernelCI/settings.py
bhcopeland: Thanks! It already goes out to this list, so I have just removed the CC here.

backend/kernelCI_app/management/commands/templates/issue_build.txt.j2
bhcopeland: Have changed in the second review, I have moved the endif down three lines. It will only trigger when there's an introduced change (the full introduced, title and link) and when there's no last-good commit (the regzbot block is omitted). So it'll only appear when the last-good/test query finds a prior passing. If nothing is found within the window (18 days) there's no range so there won't be a regzbot block.

backend/kernelCI_app/queries/notifications.py
bhcopeland: Good idea, have switched to start_time thanks!

backend/kernelCI_app/management/commands/notifications.py
bhcopeland: Updated thanks!

backend/kernelCI_app/management/commands/notifications.py
bhcopeland: Added group_size=10. Does that fit? A fail streak longer than that will probably lead to no range, which I feel is a safe outcome here. Can bump more though if you like.

alanpeixinho: I think it might suffice, the only possible downside of the group size might be that the bigger it is, the slower that query might get.
But since this is a notification command, we are probably not that worried about bigger latency.

alanpeixinho (approved):
LGTM

PR #1932, opened 2026-06-09
Add githooks to manage commit DCO sign-off

mentonin:
Adds a local commit-msg git-hook to check for commit messages without the DCO "Signed-off-by:" trailer.

Closes: #1922
Related: #1930, #1931

## Caveats
- Does not check that the DCO is valid (same email as committer)
- Only works as a dev convenience. Actual repo verification should be enabled via GH settings/CI (see #1930)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,3 +49,18 @@ repos:
         language: system
         pass_filenames: false
         stages: [pre-push]
+
+      - id: check-dco
+        name: Check DCO Sign-off
+        stages: [commit-msg]
+        entry: sh
+        args:
+          - -c
+          - |
+            if ! ( git interpret-trailers --parse "$1" | grep -q "Signed-off-by:" ); then
+              echo 'Missing DCO "Signed-off-by:" line. Try committing with "--signoff|-s"'
+              exit 1
+            fi
+          - --
+        language: system
+        always_run: true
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -49,3 +49,43 @@ repos:
...
         stages: [pre-push]
+
+      - id: check-dco
+        name: Check DCO Sign-off
+        stages: [commit-msg]
+        entry: sh
+        args:
+          - -c
+          - |
+            if ! ( git interpret-trailers --parse "$1" | grep -q "Signed-off-by:" "$1" ); then
```
bhcopeland: You want to remove the trailing $1 here. This will read it from stdin and exit 0 if it exists.

.pre-commit-config.yaml
```diff
@@ -49,3 +49,43 @@ repos:
...
+              echo 'Missing DCO "Signed-off-by:" line. Try committing with "--signoff|-s"'
+              exit 1
+            fi
+          - --
+        language: system
+        always_run: true
+
+      - id: prepare-dco
```
alanpeixinho: I think I am with @bhcopeland, might be safer if we just use the pre commit check for DCO, and leave the user to add it by themselves.
I am not sure if we have permissions to add check on CI as well, but if we do, might be a good idea.

alanpeixinho (approved):
LGTM

Issue #1929, opened 2026-06-08
Report about dashboard usage separated by user-agent


gustavobtflores:

alanpeixinho:
Closed by #1933

Issue #1591, opened 2025-10-28
Add deploy labels to Grafana dashboard


gustavobtflores:
Currently, Grafana metrics reset after each deploy because Prometheus counters restart when the process restarts, this is expected behavior. 

However, it would be useful to have labels for each deploy to improve visibility and tracking of changes.

PR #1933, opened 2026-06-09
feat: add middleware backend analytics

### LINKED ISSUE #1928: Track dashboard user behavior

alanpeixinho:
## What it is
  * Adds a middleware custom metric to log anonymous user visitor metrics for the endpoint access.
   
## How to test

  * Start a docker compose dev stack
  * Check prometheus metric port (default 8081) to check that metrics are being stored (mainly unique visitors, and endpoints_by_client).

Closes #1928 #1929

### DIFF
--- a/backend/kernelCI/settings.py
+++ b/backend/kernelCI/settings.py
@@ -88,7 +88,8 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
-    "kernelCI_app.middleware.logServerErrorMiddleware.LogServerErrorMiddleware",
+    "kernelCI_app.middleware.LogServerErrorMiddleware",
+    "kernelCI_app.middleware.BackendRequestMetricsMiddleware",
     "django_prometheus.middleware.PrometheusAfterMiddleware",
 ]
 
--- a/backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
+++ b/backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
@@ -1,0 +1,316 @@
+"""Privacy-preserving client analytics for ``/api/`` requests.
+
+This middleware records anonymous, aggregate usage metrics. No personal data
+(raw IP, raw User-Agent, full referrer URL) is stored or exposed as a metric
+label.
+
+Collected as aggregate Prometheus counters only:
+  * Request attributes: endpoint, method, status_class, and coarse client
+    buckets (browser, os, device) derived from the User-Agent. Referrer is
+    reduced to its external domain (or ``direct_or_internal``).
+  * Daily unique-visitor estimates (total and per-endpoint).
+
+Visitor anonymization: a fingerprint is computed as
+``HMAC-SHA256(daily_salt, "<ip>|<user_agent>")``. The ``daily_salt`` is a
+random 32-byte secret generated per UTC day, kept only in the cache with a
+~25h TTL, and rotated daily so hashes cannot be linked across days. Only the
+hash is used as a de-duplication cache key; raw IP/User-Agent are discarded
+immediately and never persisted.
+
+See ``docs/monitoring.md`` ("Client Analytics & Privacy") for full details and
+compliance notes.
+"""
+
+import hashlib
+import hmac
+import logging
+import re
+import secrets
+from dataclasses import dataclass
+from datetime import UTC, datetime
+from urllib.parse import urlparse
+
+from django.core.cache import cache
+from django.core.exceptions import DisallowedHost
+from prometheus_client import Counter
+
+UNKNOWN = "unknown"
+DIRECT_OR_INTERNAL = "direct_or_internal"
+UNIQUE_VISITOR_TTL_SECONDS = 25 * 60 * 60  # 25h
+UNIQUE_VISITOR_SALT_BYTES = 32
+
+logger = logging.getLogger(__name__)
+
+DASHBOARD_BACKEND_REQUESTS_BY_CLIENT = Counter(
+    "dashboard_backend_requests_by_client_total",
+    "Backend requests grouped by endpoint and client attributes",
+    [
+        "endpoint",
+        "method",
+        "status_class",
+        "browser",
+        "os",
+        "device",
+        "referrer_domain",
+    ],
+)
+
+DASHBOARD_UNIQUE_VISITORS_TOTAL = Counter(
+    "dashboard_unique_visitors_total",
+    "Daily unique backend visitors",
+)
+
+DASHBOARD_UNIQUE_VISITORS_BY_ENDPOINT_TOTAL = Counter(
+    "dashboard_unique_visitors_by_endpoint_total",
+    "Daily unique backend visitors deduplicated per endpoint by rotated Redis salt",
+    ["endpoint"],
+)
+
+
+@dataclass(frozen=True)
+class ClientInfo:
+    browser: str
+    os: str
+    device: str
+
+
+class BackendRequestMetricsMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if request.path.startswith("/api/"):
+            labels = get_backend_request_labels(request, response)
+            record_client(**labels)
+            record_unique_visitor(request=request, endpoint=labels["endpoint"])
+        return response
+
+
+def record_client(
+    *,
+    endpoint: str,
+    method: str,
+    status_class: str,
+    browser: str,
+    os: str,
+    device: str,
+    referrer_domain: str,
+) -> None:
+    DASHBOARD_BACKEND_REQUESTS_BY_CLIENT.labels(
+        endpoint=endpoint,
+        method=method,
+        status_class=status_class,
+        browser=browser,
+        os=os,
+        device=device,
+        referrer_domain=referrer_domain,
+    ).inc()
+
+
+def record_unique_visitor(*, request, endpoint: str) -> None:
+    try:
+        analytics_date = get_analytics_date()
+        visitor_hash = get_daily_visitor_hash(request, analytics_date=analytics_date)
+        if visitor_hash is None:
+            return
+
+        visitor_key = f"analytics:unique-visitors:{analytics_date}:{visitor_hash}"
+        endpoint_visitor_key = (
+            f"analytics:unique-visitors:{analytics_date}:"
+            f"endpoint:{endpoint}:{visitor_hash}"
+        )
+
+        if cache.add(visitor_key, "true", timeout=UNIQUE_VISITOR_TTL_SECONDS):
+            DASHBOARD_UNIQUE_VISITORS_TOTAL.inc()
+
+        if cache.add(endpoint_visitor_key, "true", timeout=UNIQUE_VISITOR_TTL_SECONDS):
+            DASHBOARD_UNIQUE_VISITORS_BY_ENDPOINT_TOTAL.labels(endpoint=endpoint).inc()
+    except Exception as exc:
+        logger.debug("Failed to record unique visitor metric: %s", exc)
+
+
+def get_daily_visitor_hash(request, *, analytics_date: str) -> str | None:
+    user_agent = request.headers.get("User-Agent", "")
+    client_ip = get_client_ip(request)
+    if not client_ip:
+        return None
+
+    daily_salt = get_daily_salt(analytics_date)
+    if daily_salt is None:
+        return None
+
+    message = f"{client_ip}|{user_agent}".encode()
+    return hmac.new(daily_salt.encode(), message, hashlib.sha256).hexdigest()
+
+
+def get_daily_salt(analytics_date: str) -> str | None:
+    salt_key = f"analytics:unique-visitors:salt:{analytics_date}"
+    daily_salt = cache.get(salt_key)
+    if daily_salt is not None:
+        return daily_salt
+
+    candidate_salt = secrets.token_hex(UNIQUE_VISITOR_SALT_BYTES)
+    cache.add(salt_key, candidate_salt, timeout=UNIQUE_VISITOR_TTL_SECONDS)
+
+    daily_salt = cache.get(salt_key)
+    return daily_salt
+
+
+def get_analytics_date() -> str:
+    return datetime.now(UTC).date().isoformat()
+
+
+def get_client_ip(request) -> str:
+    forwarded_for = request.headers.get("X-Forwarded-For", "")
+    if forwarded_for:
+        return forwarded_for.split(",", maxsplit=1)[0].strip()
+
+    forwarded = request.headers.get("Forwarded", "")
+    if forwarded:
+        forwarded_ip = parse_forwarded_for(forwarded)
+        if forwarded_ip:
+            return forwarded_ip
+
+    return request.META.get("REMOTE_ADDR", "").strip()
+
+
+def parse_forwarded_for(forwarded: str) -> str:
+    first_hop = forwarded.split(",", maxsplit=1)[0]
+    for part in first_hop.split(";"):
+        key, _, value = part.strip().partition("=")
+        if key.lower() == "for":
+            return normalize_forwarded_node(value.strip().strip('"'))
+    return ""
+
+
+def normalize_forwarded_node(node: str) -> str:
+    if node.startswith("["):  # IPV6 [2001:db8::1]:8080
+        return node[1 : node.find("]")] if "]" in node else node[1:]
+    if node.count(":") == 1:  # IPV4 192.0.2.60:8080
+        return node.split(":", maxsplit=1)[0]
+    return node
+
+
+def get_backend_request_labels(request, response) -> dict[str, str]:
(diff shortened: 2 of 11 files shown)


backend/kernelCI/settings.py
```diff
@@ -89,6 +89,7 @@ def get_json_env_var(name, default):
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "kernelCI_app.middleware.logServerErrorMiddleware.LogServerErrorMiddleware",
+    "kernelCI_app.middleware.backendRequestMetricsMiddleware.BackendRequestMetricsMiddleware",
```
mentonin: We should probably use a module and re-export the middleware classes to improve readability

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+from urllib.parse import urlparse
+
+from django.core.cache import cache
+from django.core.exceptions import DisallowedHost
+from prometheus_client import Counter
+
+UNKNOWN = "unknown"
+DIRECT_OR_INTERNAL = "direct_or_internal"
+UNIQUE_VISITOR_TTL_SECONDS = 48 * 60 * 60
+UNIQUE_VISITOR_SALT_BYTES = 32
```
mentonin: Why 2 days?

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+    "Backend requests grouped by endpoint and client attributes",
+    [
+        "endpoint",
+        "method",
+        "status_class",
+        "browser",
+        "os",
+        "device",
+        "referrer_domain",
+    ],
```
mentonin, line 55: This may result in very high cardinality

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+    "dashboard_unique_visitors_total",
+    "Daily unique backend visitors",
+)
+
+DASHBOARD_UNIQUE_VISITORS_BY_ENDPOINT_TOTAL = Counter(
+    "dashboard_unique_visitors_by_endpoint_total",
+    "Daily unique backend visitors deduplicated per endpoint by rotated Redis salt",
+    ["endpoint"],
+)
+
```
mentonin, line 68: I think the name might be confusing. I associate "visitors" with website access, while this tracks api usage. I would rather have a name using "backend" or "api", or maybe replacing "visitors" with something like "consumers"

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+        response = self.get_response(request)
+        if request.path.startswith("/api/"):
+            labels = get_backend_request_labels(request, response)
+            record_client(**labels)
+            record_unique_visitor(request=request, endpoint=labels["endpoint"])
+        return response
+
+
+def record_client(**kwargs) -> None:
+    DASHBOARD_BACKEND_REQUESTS_BY_CLIENT.labels(**kwargs).inc()
```
mentonin: I don't like kwargs here, the dict keys are known and must be the same as DASHBOARD_BACKEND_REQUESTS_BY_CLIENT labels.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+    return daily_salt
+
+
+def get_analytics_date() -> str:
+    return datetime.now(UTC).date().isoformat()
+
+
+def get_client_ip(request) -> str:
+    forwarded_for = request.headers.get("X-Forwarded-For", "")
+    if forwarded_for:
```
mentonin, line 166: Should we handle `Forwarded` header as well?

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+            request_host=get_request_host(request),
+        ),
+    }
+
+
+def get_request_host(request) -> str:
+    try:
+        return request.get_host()
+    except DisallowedHost:
+        return ""
```
mentonin: Maybe return `UNKNOWN` here

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+    return UNKNOWN
+
+
+def get_device(normalized_user_agent: str) -> str:
+    if "ipad" in normalized_user_agent or "tablet" in normalized_user_agent:
+        return "tablet"
+    if "mobile" in normalized_user_agent or "iphone" in normalized_user_agent:
+        return "mobile"
+    if "android" in normalized_user_agent:
+        return "mobile"
```
mentonin:
```suggestion
    if any(s in normalized_user_agent for s in ["mobile", "iphone", "android"]):
      return true
```

mentonin: This is clearer in my opinion, and keeps each specific return value easily traceable

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+    if "chrome/" in normalized_user_agent or "crios/" in normalized_user_agent:
+        return "Chrome"
+    if "safari/" in normalized_user_agent:
+        return "Safari"
+    if "msie" in normalized_user_agent or "trident/" in normalized_user_agent:
+        return "Internet Explorer"
+    if "curl/" in normalized_user_agent or "wget/" in normalized_user_agent:
+        return "HTTP Client"
+    if "python-requests/" in normalized_user_agent:
+        return "HTTP Client"
```
mentonin: see comment in get_device

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+        "method": request.method.upper(),
+        "status_class": get_status_class(response.status_code),
+        "browser": client_info.browser,
+        "os": client_info.os,
+        "device": client_info.device,
+        "referrer_domain": get_referrer_domain(
+            referrer=request.headers.get("Referer", ""),
+            request_host=get_request_host(request),
+        ),
+    }
```
mentonin, line 209:
```suggestion
    return client_info | {
        "endpoint": get_endpoint(request),
        "method": request.method.upper(),
        "status_class": get_status_class(response.status_code),
        "referrer_domain": get_referrer_domain(
            referrer=request.headers.get("Referer", ""),
            request_host=get_request_host(request),
        ),
    }
```

Matter of preference, but I like dictionary union here

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+def get_daily_salt(analytics_date: str) -> str | None:
+    salt_key = f"analytics:unique-visitors:salt:{analytics_date}"
+    daily_salt = cache.get(salt_key)
+    if daily_salt is not None:
+        return daily_salt
+
+    candidate_salt = secrets.token_hex(UNIQUE_VISITOR_SALT_BYTES)
+    cache.add(salt_key, candidate_salt, timeout=UNIQUE_VISITOR_TTL_SECONDS)
```
mentonin, line 154: Can the cache be evicted before timeout? We would lose the salt then.

proxy/etc/nginx/templates/default.conf.template
```diff
@@ -6,6 +6,7 @@ server {
         proxy_send_timeout 240s;
         send_timeout 240s;
     }
+
```
mentonin: Formatting change for an otherwise unchaged file

mentonin (commented):
Overall looks pretty good and fairly comprehensive for backend visitor metrics, my comments are mostly nits or suggestions.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: I have mostly overshoot the necessary time (maybe too much), just to be sure we don't loose cached values.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Do you have any suggestions?

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: I will disagree on this one, despite liking to keep the code more declarative, a nested loop might be less readable

backend/kernelCI/settings.py
alanpeixinho: I prefer explicit naming the class, module exporting in python can lead to far more confusion

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Took some time to look it up, and the default for redis has eviction disabled, so it should not be a problem.
Anyway, I made it explicit to make sure we dont change this behaviour.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: The point here is, client_info is actually a dataclass, not a dict, and mapping it to a dict for dictionary union might be more confusing.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: ditto

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Done

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: @tales-aparecida do you have any thoughts here? Do you think using "consumers" to address what we have been calling "visitors" would be clearer for the intended audience?

proxy/etc/nginx/templates/default.conf.template
alanpeixinho: Done

backend/kernelCI/settings.py
mentonin: I don't see how `kernelCI_app.middleware.backendRequestMetricsMiddleware.BackendRequestMetricsMiddleware` is better than `kernelCI_app.middleware.BackendRequestMetricsMiddleware`, but not a blocker.

docs/monitoring.md
```diff
@@ -118,3 +118,51 @@ Configure these variables in `.env.backend`:
 - **Target**: `host.docker.internal:8001` (backend running locally)
 - **Metrics Path**: `/metrics/`
 - **Scrape Interval**: 15 seconds
+
+## Client Analytics & Privacy
```
mentonin: I believe we must move this into a proper Privacy Policy file and link to it from the frontend for full compliance. I suggest going through with these changes and tracking the Privacy Policy as a new issue.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
mentonin: I think we can test it and deal with it later when/if issues arise.
For sanity checks, we could look into the cardinality and relevance of each item and see if we should drop anything:

- endpoint: around 30?, grows with API surface, very relevant, could be filtered or processed into smaller bins if needed
- method: 9, I think only one value is valid for most endpoints, which endpoints need this? Are invalid requests filtered out before or after this point? Could bloat cardinality unnecessarily if we track invalid requests.
- status_class: 5, relevant for tracking availability and server errors
- browser: 8. Relevant for tracking build targets and separating browsers, bots and non-browser consumers (kci-dev). We could reduce browser granularity, or track specific tools (kci-dev sets a specific user-agent).
- os: 7, I don't think it is very relevant besides device type
- device: 3 (no "unknown"), very relevant
- referrer_domain: virtually infinite, a handful in practice. I think we could reduce to `internal`, `direct`, `external` and maybe add specific tracking of other sources later?

Potential cardinality of over 2 million with 10 referrers, 75600 if we only have one method per endpoint and 3 referrers

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
mentonin: I would still prefer a single `return "mobile"` exit point

backend/kernelCI/settings.py
alanpeixinho: Because it is explicit, changing module exports in runtime introduces quite a lot of noise for a simple string. Or you think differently?

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: I think the more sound thing to do here is to start collecting some information. This way we have a proper idea of which columns might shown problems, and then act on them.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Done

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: I believe it is not necessary, as nginx uses x-forwarded-for, but still valid to keep.
Done.

docs/monitoring.md
alanpeixinho: Done.

alanpeixinho: Extending on this, we might need to include contact information on the PRIVACY.md doc, which contact should we include here @bhcopeland ?

tales-aparecida: There's should be some tracker for this already. I vaguely recall that we had planned to use the same framework the Linux Foundation website is using

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
tales-aparecida: I'll say: copy the names from well-established frameworks

alanpeixinho: For what I see, visitors is by far the most used term here, even for backend-only tracking.
Client is sometimes adopted as well, and could be an alternative.

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: I did it your way

backend/kernelCI/settings.py
alanpeixinho: Did it your way

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Keeping visitors here. All docs are clear on what we are tracking

backend/kernelCI_app/middleware/backendRequestMetricsMiddleware.py
alanpeixinho: Done

docs/monitoring.md
alanpeixinho: I followed the same as Linux Foundation. But I will consult on this regardless.

mentonin (approved):

alanpeixinho:
@bhcopeland @nuclearcat @tales-aparecida 
I re-checked #1933 for GDPR before we take this to production.
We should be covering the technical aspect: aggregate metrics only, ~25h pseudonymised de-dup in cache, no cookies, and no long-term per-user IDs.
 
I included some description on PRIVACY.md to make it more explicit that we are using pseudonymisation of IP/User-Agent. Also included a footer notice, and a link to our PRIVACY.md and the Linux Foundation Projects privacy policy.

I found one important problem (#1591), which causes prometheus aggregated metrics to be restarted on new deploys. I'm working on that right now, as this might defeat the purpose of starting to collect metrics.

If you think we are safe to deploy, I am happy to merge.

tales-aparecida:
In general, prometheus' metric-resetting is dealt with querying the increase/delta, it this not applicable here?

alanpeixinho:
> In general, prometheus' metric-resetting is dealt with querying the increase/delta, it this not applicable here?

Yes, after checking the issue on #1951, I believe this should not be a problem in this scenario as I initially considered.
Despite having the counters reset. Since we have no total counters, we are already using "reset-safe" panels here.

tales-aparecida (approved):

PR #1934, opened 2026-06-10
feat: increase metrics cache time

gustavobtflores:
Increase the metrics endpoint cache time. Right now, metrics queries are expensive, and since they are aggregated over multiple days, we can keep them cached for longer.

### DIFF
--- a/backend/kernelCI_app/queries/notifications.py
+++ b/backend/kernelCI_app/queries/notifications.py
@@ -16,6 +16,8 @@ from kernelCI_app.typeModels.metrics_notifications import (
     TopIssue,
 )
 
+METRICS_CACHE_TIMEOUT = 60 * 60 * 6  # 6 hours
+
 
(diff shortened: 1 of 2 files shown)


backend/kernelCI_app/queries/notifications.py
```diff
@@ -16,6 +16,8 @@
     TopIssue,
 )
 
+METRICS_CACHE_TIMEOUT = 60 * 60  # 1 hour
```
alanpeixinho: I think we can go even higher here. Considering this page is still a prototype for design iteration.

alanpeixinho (approved):
LGTM

backend/kernelCI_app/queries/notifications.py
gustavobtflores: Yeah, for sure, increased this to 6h

PR #1937, opened 2026-06-12
fix: correct ordering on revisions selector on hardware page

gustavobtflores:
Fixes inconsistent ordering in the hardware page revision selector by sorting trees, branches, and revisions in the hardware selectors API response before returning them to the frontend.

## How to test

1. Open the hardware page.
2. Open the revision selector dropdown.
3. Confirm trees appear in alphabetical order, branches are grouped and ordered consistently, and revisions within each branch show newest commits first.

### DIFF
--- a/backend/kernelCI_app/views/hardwareSelectorsView.py
+++ b/backend/kernelCI_app/views/hardwareSelectorsView.py
@@ -53,6 +53,21 @@ class HardwareSelectorsView(APIView):
                 }
             )
 
+        for tree in trees:
+            tree["branches"].sort(
+                key=lambda branch: (
+                    branch["git_repository_url"],
+                    branch["git_repository_branch"],
+                )
+            )
+            for branch in tree["branches"]:
+                branch["revisions"].sort(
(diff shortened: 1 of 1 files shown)


backend/kernelCI_app/views/hardwareSelectorsView.py
```diff
@@ -53,6 +53,21 @@ def _sanitize_records(self, selectors_raw: list[dict]) -> HardwareSelectorsRespo
                 }
             )
 
+        for tree in trees:
```
alanpeixinho, line 56: nit: because the difference might be negligible, but with this, we might want to remove the order by on the database query

alanpeixinho (approved):

backend/kernelCI_app/views/hardwareSelectorsView.py
gustavobtflores: The ordering in the query is different from what we want (newest-first) and is needed for the `DISTINCT ON`, we could move all the ordering to the query, but it would increase complexity a little

PR #1941, opened 2026-06-15
fix/hardware revision selector optional

alanpeixinho:
## What it is

* Make the revision selectors in hardware listing page optional

## How to test it

* Go to the hardware listing page
* The hardware listing should load the full hardware listing.
* When selecting a revision, the hardware listing should be filtered to count only instances of that revision.
* When deselecting the filter the page behaviour should go back to full listing.
* Simplify query to use revisions built on last 30 days, instead of tests ran on last 15 days.

Closes #1939


felipebergamin (approved):

backend/kernelCI_app/queries/hardware.py
```diff
@@ -135,11 +135,13 @@ def get_hardware_selectors(origin: str) -> list[dict]:
...
         FROM checkouts c
         INNER JOIN builds b ON b.checkout_id = c.id
-        INNER JOIN tests t ON t.build_id = b.id
         WHERE
-            t.origin = %(origin)s
-            AND t.start_time > (NOW() - INTERVAL '15 days')
-            AND t.environment_misc ->> 'platform' IS NOT NULL
+            b.origin = %(origin)s
+            AND b.start_time > (NOW() - INTERVAL '30 days')
+            AND c.git_repository_url IS NOT NULL
```
mentonin, line 141: I think this should also check for empty strings and for `c.commit_name`

alanpeixinho: Good point on empty strings. But I am not sure if we should ignore rows without a commit_name, as we can still use the commit_hash when commit_name is missing.

mentonin: Does the commit selector fall back to commit hash when name is null/empty string?

alanpeixinho: Thanks @mentonin helped us track that there are no empty strings for this case.
On the commit_name, the HardwareRevisionSelectors seems to fallback to the commit hash if not available, so we better keep it.

mentonin (approved):

PR #1936, opened 2026-06-10
fix: Include aggregate tables on snapshot/restore update_db commands

alanpeixinho:
## What it is
Include the missing tables TreeTestsRollup/TreeListing tables on the snapshot tools.

## How to test
* From an already populated database. Executes command `update_db snapshot` with any parameters combination.
* In a new empty database perform the `update_db restore` command with the file generated by previous command.
* The new database should contain the information retrieved from original database.

Closes #1935

Signed-off-by: Alan Peixinho <alan.peixinho@profusion.mobi>

PR #1923, opened 2026-06-03
refactor: Unversion pages/endpoints for tree/hardware listing

alanpeixinho:
## What it is
  * Unify the v1/v2 pages and endpoints for tree/pages listing.
  * Remove CheckoutsCache.

## How to test
  * Start backend server and frontend
  * make a request for the unversioned pages (as the examples):
      * `/tree?origin=maestro&intervalInDays=7`
      * `/hardware?origin=maestro&intervalInDays=7`
  * make a request for backend endpoints (as the examples):
      * `/api/tree/?origin=maestro&interval_in_days=7`
      * `/api/hardware/?origin=maestro&startTimestampInSeconds=1741192200&endTimestampInSeconds=1741624200`
  * make sure older versioning no longer exists (as the examples):
      * `/api/tree-v2/`
      * `/api/hardware-v2/`
      
Closes #1917

PR #1942, opened 2026-06-17
fix/weekly cache metrics endpoint

alanpeixinho:
## What it is

Makes the metrics page cache its response for a week, so we can:
  1. Keep parity with email
  2. Avoid slow endpoint calls

### Test warming cache crontab
  1. First install the crontabs via `poetry run python3 manage.py crontab add`
  2. Make sure they are installed `poetry run python3 manage.py crontab show`
  3. You can execute the crontab via `poetry run python3 manage.py crontab run <hash of cronjob>`
    * It is important to point here that the warm is supposed to run at the same time as email metrics (saturday), so they might not be aligned to the same days as frontend requests.
    * Alternatively to this, we could change the system clock.
  4. Verify the function that queries and stores on redis runs properly.
  5. Verify redis now has updated cache on the keys (`metricsTotalObjects`, `metricsBuildIncidents`, `metricsNewBuildIncidents`, `metricsLabSummary`) (via redis-cli) or via python script.
  5. Remember to remove the crontab from your system `poetry run python3 manage.py crontab remove`

### Testing endpoint
  1. Go to the metrics page
  2. Make sure it loads data from the range [lastSaturday - 7 days, lastSaturday] for the "previous week" and [lastSaturday - 14, lastSaturday] for the "2 previous weeks".
  3. Make sure subsequent access are cached.

  Closes #1940

- **feat: add checkout selectors on hardware page (#1904)**
- **feat: dynamic metrics page (#1920)**
- **refactor: unversion tree and hardware listing flows (#1923)**
- **feat: increase metrics cache time (#1934)**
- **Add githooks to manage commit DCO sign-off (#1932)**
- **Reports regzbot tracking (#1925)**
- **fix: Include aggregate tables on snapshot/restore update_db commands (#1936)**
- **fix(metrics): stabilize query cache and add weekly warm cron**

### DIFF
--- a/backend/kernelCI_app/queries/notifications.py
+++ b/backend/kernelCI_app/queries/notifications.py
@@ -1,5 +1,6 @@
 import sys
 from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime, time, timedelta, timezone
 from typing import Any
 
 from django.db import connection, connections
@@ -16,7 +17,9 @@ from kernelCI_app.typeModels.metrics_notifications import (
     TopIssue,
 )
 
+METRICS_CACHE_WARM_PERIODS = (7, 14)
 METRICS_CACHE_TIMEOUT = 60 * 60 * 6  # 6 hours
+METRICS_CACHE_WARM_TIMEOUT = 60 * 60 * 24 * 8  # 8 days
 
 
 def kcidb_execute_query(query, params=None):
@@ -698,17 +701,24 @@ def query_fetchone_work(
     query: str,
     params: dict[str, Any],
     timeout: int = METRICS_CACHE_TIMEOUT,
-):
-    rows = get_query_cache(key=cache_key, params=params)
-    if rows is not None:
-        return rows
+    use_cache: bool = True,
+) -> Any:
+    if use_cache:
+        cached = get_query_cache(key=cache_key, params=params)
+        if cached is not None:
+            return cached
     try:
         with connections["default"].cursor() as cursor:
             cursor.execute(query, params)
             rows = cursor.fetchone()
     finally:
         connections["default"].close()
-    set_query_cache(key=cache_key, params=params, rows=rows, timeout=timeout)
+    set_query_cache(
+        key=cache_key,
+        params=params,
+        rows=rows,
+        timeout=timeout,
+    )
     return rows
 
 
@@ -718,24 +728,50 @@ def query_fetchall_work(
     query: str,
     params: dict[str, Any],
     timeout: int = METRICS_CACHE_TIMEOUT,
-):
-    rows = get_query_cache(key=cache_key, params=params)
-    if rows is not None:
-        return rows
+    use_cache: bool = True,
+) -> Any:
+    if use_cache:
+        cached = get_query_cache(key=cache_key, params=params)
+        if cached is not None:
+            return cached
     try:
         with connections["default"].cursor() as cursor:
             cursor.execute(query, params)
             rows = cursor.fetchall()
     finally:
         connections["default"].close()
-    set_query_cache(key=cache_key, params=params, rows=rows, timeout=timeout)
+    set_query_cache(
+        key=cache_key,
+        params=params,
+        rows=rows,
+        timeout=timeout,
+    )
     return rows
+
+
+def interval_params(start_days_ago: int, end_days_ago: int) -> dict[str, str]:
+    """Build [start_date, end_date) bounds from UTC day offsets."""
+    today = datetime.now(timezone.utc).date()
+    start_datetime = datetime.combine(
+        today - timedelta(days=start_days_ago), time.min, tzinfo=timezone.utc
+    )
+    end_datetime = datetime.combine(
+        today - timedelta(days=end_days_ago),
+        time.min,
+        tzinfo=timezone.utc,
+    )
+    return {
+        "start_date": start_datetime.isoformat(),
+        "end_date": end_datetime.isoformat(),
+    }
 
 
 def get_metrics_data(
     *,
     start_days_ago: int,
     end_days_ago: int,
+    use_cache: bool = True,
+    cache_timeout: int = METRICS_CACHE_TIMEOUT,
 ) -> MetricsReportData:
 
     if start_days_ago > 30:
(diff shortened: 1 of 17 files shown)


backend/kernelCI_app/queries/notifications.py
```diff
@@ -718,24 +730,51 @@ def query_fetchall_work(
...
+
+
+def interval_params(start_days_ago: int, end_days_ago: int) -> dict[str, str]:
+    today = datetime.now(timezone.utc).date()
+    start_datetime = datetime.combine(
+        today - timedelta(days=start_days_ago), time.min, tzinfo=timezone.utc
+    )
+    end_datetime = datetime.combine(
+        today - timedelta(days=end_days_ago),
+        time.max,
```
bhcopeland: Trying to work out the SQL below, I think this will be an 8-day window? Because it causes SQL to be at the end of the 7-day window (i.e. on the last day it will be at midnight).

alanpeixinho: yes, you are correct.
The cached query is happening at the start of the day (just like the mail notification), so the 8th day would end up empty.
But still, it is not robust to rely on time of the call.

backend/kernelCI_app/queries/notifications.py
```diff
@@ -747,40 +787,34 @@ def get_metrics_data(
...
-            NOW() - INTERVAL %(start_days_ago)s
-            AND NOW() - INTERVAL %(end_days_ago)s)
+        (SELECT COUNT(DISTINCT tree_name) FROM checkouts WHERE _timestamp >=
+            %(start_date)s::timestamptz
+            AND _timestamp < %(end_date)s::timestamptz)
             AS n_trees,
-        (SELECT COUNT(*) FROM checkouts WHERE _timestamp BETWEEN
-            NOW() - INTERVAL %(start_days_ago)s
-            AND NOW() - INTERVAL %(end_days_ago)s)
+        (SELECT COUNT(*) FROM checkouts WHERE _timestamp >=
```
mentonin: I believe you can merge this SELECT with the previous one for (minor) performance gains

backend/kernelCI_app/tests/factories/seed_time.py
```diff
@@ -0,0 +1,9 @@
+from datetime import datetime, time, timedelta, timezone
+
+from django.utils import timezone as django_timezone
+
+
+def seeded_timestamp(*, days_ago: int = 1) -> datetime:
```
mentonin, line 10: It would be great if we could include some specific tests related to the timestamp: out-of-order rows, cases before/after query intervals, check open or closed interval boundaries, etc.

dashboard/src/pages/Metrics/MetricsPage.tsx
```diff
@@ -542,21 +550,44 @@ function LabsSection({ labs }: { labs: LabData[] }): JSX.Element {
   );
 }
 
+const getMetricsQueryParams = (
+  activeDays: number,
+): { startDaysAgo: number; endDaysAgo: number } => {
+  const today = new Date();
+  const weekEndDaysAgo = (today.getUTCDay() + 1) % DAYS_IN_WEEK;
+  const endDaysAgo = Math.max(weekEndDaysAgo - 1, 0);
+  const startDaysAgo = endDaysAgo + activeDays + (weekEndDaysAgo > 0 ? 1 : 0);
```
mentonin: IIUC this gets an interval of `activeDays + 1` complete days (e.g. it includes 2 Fridays instead of 1 when `activeDays=7`)

mentonin: On a broader note, I don't like the interface here. `/metrics?i=n` returns metrics from `n` days before the last Saturday, which is not really intuitive. I think you can only set i=7 or i=14 from the UI, but then why do we expose that as an integer for the user to edit?

dashboard/src/locales/messages/index.ts
```diff
@@ -286,6 +286,10 @@ export const messages = {
       'This is the new, optimized version of the {page}. If you find any bugs, please file a {gitHubLink} and you can still access the old version {oldVersionLink}. Please note that some historical data might be missing, but it should be updated with recent data.',
     'messages.olderPageVersion':
       'This is the legacy version of the {page}, please refer to the new, optimized version {newPageLink}. If you find any bugs or divergences, please report to {gitHubLink}.',
+    'metricsPage.computedAt': 'Computed {computedAt}',
+    'metricsPage.period.previousTwoWeeks': 'Previous 2 weeks',
+    'metricsPage.period.previousWeek': 'Previous week',
```
mentonin:
```suggestion
    'metricsPage.period.previousWeek': 'Last week',
```

mentonin (commented):
Added some comments. I think this works in general, but should be seen as a palliative solution. A better solution might be a rollup table. I think we could also use a set of prometheus metrics (we are tracking difference between timestamps of time series with low cardinality).

backend/kernelCI_app/queries/notifications.py
alanpeixinho: Good point

dashboard/src/pages/Metrics/MetricsPage.tsx
alanpeixinho: Nice catch, I changed the backend behaviour and forgot to update here at some point.

alanpeixinho: On the topic of the url query param, I think I am with you.
I am removing this parameter altogether, unless users actually show need to share specific ranges via url. And even if they do, we might limit the options.

dashboard/src/pages/Metrics/MetricsPage.tsx
```diff
@@ -542,33 +550,42 @@ function LabsSection({ labs }: { labs: LabData[] }): JSX.Element {
...
-            });
-          }}
-        />
+      <div className="flex flex-col gap-3 sm:flex-row sm:items-start sm:justify-between">
+        <div className="space-y-1 text-sm text-gray-500">
+          {data ? (
+            <p>
+              {formatMessage(
+                { id: 'metricsPage.computedAt' },
+                { computedAt: formatDate(data.created_at, false, true) },
```
mentonin: Should we display "Created at" or the actual interval spanned by the data?

backend/kernelCI_app/tests/factories/seed_time.py
```diff
@@ -0,0 +1,20 @@
+from datetime import datetime, time, timedelta, timezone
+
+from django.utils import timezone as django_timezone
+
+# Seeded rows are spread across this many days (days_ago 0..SEED_DAY_SPAN-1), all at
+# UTC midnight, so each lands exactly on an interval_params day boundary. This lets
+# adjacent metrics windows return distinct, non-zero counts to exercise the half-open
+# [start, end) boundaries on both sides.
+SEED_DAY_SPAN = 8
```
mentonin: IIUC, this creates data at 00:00:00 at current day up to 7 days ago. The default query fetches events from 7 days ago, 00:00:00 up to yesterday, 23:59:59. So this does not test start date

backend/kernelCI_app/tests/factories/seed_time.py
mentonin: GH won't let me mark this as resolved, but tests are looking very robust now 👍🏼

backend/kernelCI_app/tests/integrationTests/metrics_test.py
```diff
@@ -28,32 +28,49 @@
...
-    "n_tests": 30,
-    "n_issues": 20,
-    "n_incidents": 8,
+    "n_trees": 9,
+    "n_checkouts": 50,
+    "n_builds": 24,
+    "n_tests": 26,
+    "n_issues": 17,
+    "n_incidents": 7,
     "prev_n_trees": 0,
```
mentonin: previous values should probably also be seeded and tested for

dashboard/src/locales/messages/index.ts
```diff
@@ -286,6 +286,10 @@ export const messages = {
       'This is the new, optimized version of the {page}. If you find any bugs, please file a {gitHubLink} and you can still access the old version {oldVersionLink}. Please note that some historical data might be missing, but it should be updated with recent data.',
     'messages.olderPageVersion':
       'This is the legacy version of the {page}, please refer to the new, optimized version {newPageLink}. If you find any bugs or divergences, please report to {gitHubLink}.',
+    'metricsPage.computedAt': 'Computed {computedAt}',
+    'metricsPage.period.previousTwoWeeks': 'Previous 2 weeks',
```
mentonin:
```suggestion
    'metricsPage.period.previousTwoWeeks': 'Last 2 weeks',
```
For consistency, since you accepted my suggestion

dashboard/src/pages/Metrics/MetricsPage.tsx
alanpeixinho: In this case I really want to show the time data was computed, mostly because it is a long cache.
We could also show the range of the last week, but I am not sure if we gain much from doing it.

backend/kernelCI_app/tests/factories/seed_time.py
alanpeixinho: Good point. I have moved the span further, not only to include prev responses, but also to help the test be more robust, and fail if we do not properly filter start date.

dashboard/src/pages/Metrics/MetricsPage.tsx
mentonin: Wince we are slicing by `_timestamp`, I don't think the computed time is useful: the metrics for a specific time interval (in the past) should be constant

mentonin (approved):

dashboard/src/pages/Metrics/MetricsPage.tsx
alanpeixinho: Fair. I am removing, since it was introducing unecessary complexity

backend/kernelCI_app/tests/integrationTests/metrics_test.py
mentonin: 👍🏼

dashboard/src/locales/messages/index.ts
mentonin: 👍🏼

PR #1955, opened 2026-06-24
fix: communicate fixed listing filters on tree and hardware (#1924)

mentonin:
## Description

Removes unnecessary time filtering from `/trees` and add label to indicate time filtering to `/hardware`.

## Related Issues
- Depends: #1941 - the `/hardware` label incorporates changes in filtering happening in that PR
- Closes: #1924

### DIFF
--- a/dashboard/src/components/FilterLabel/FilterLabel.tsx
+++ b/dashboard/src/components/FilterLabel/FilterLabel.tsx
@@ -1,0 +1,16 @@
+import type { JSX } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export function FilterLabel({ days }: { days: number }): JSX.Element {
+  return (
+    <p
+      className="text-dim-gray text-left text-xs"
+      data-test-id="hardware-filter-label"
+    >
+      <FormattedMessage
+        id="hardwareListing.latestCheckoutFilterLabel"
+        values={{ days }}
+      />
+    </p>
+  );
+}
--- a/dashboard/src/components/FilterLabel/FilterLabel.tsx
+++ b/dashboard/src/components/FilterLabel/FilterLabel.tsx
@@ -5,10 +5,10 @@ export function FilterLabel({ days }: { days: number }): JSX.Element {
   return (
     <p
       className="text-dim-gray text-left text-xs"
-      data-test-id="hardware-filter-label"
+      data-test-id="listing-filter-label"
     >
       <FormattedMessage
-        id="hardwareListing.latestCheckoutFilterLabel"
+        id="filter.latestCheckoutFilterLabel"
         values={{ days }}
       />
     </p>
--- a/dashboard/src/components/TreeListingPage/TreeTable.tsx
+++ b/dashboard/src/components/TreeListingPage/TreeTable.tsx
@@ -34,6 +34,7 @@ import { usePaginationState } from '@/hooks/usePaginationState';
 import BaseTable, { TableHead } from '@/components/Table/BaseTable';
 import { TableBody, TableCell, TableRow } from '@/components/ui/table';
 import { ConditionalTableCell } from '@/components/Table/ConditionalTableCell';
+import { FilterLabel } from '@/components/FilterLabel/FilterLabel';
 
 import { BaseGroupedStatusWithLink } from '@/components/Status/Status';
 import { TableHeader } from '@/components/Table/TableHeader';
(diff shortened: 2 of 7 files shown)


dashboard/src/utils/constants/general.ts
```diff
@@ -9,4 +9,4 @@ export const ItemsPerPageValues = [5, 10, 20, 30, 40, 50];
 export const DEFAULT_TIME_SEARCH = 7;
 export const REDUCED_TIME_SEARCH = 5;
 export const DEFAULT_LISTING_ITEMS = 10;
-export const DEFAULT_LISTING_INTERVAL_IN_DAYS = 30;
+export const DEFAULT_LISTING_INTERVAL_IN_DAYS = 36500; // 100 years, effectively no filter
```
alanpeixinho: If we want to remove this interval, we should remove the parameter from the backend as well. We also need to check if this can affect performance.
Wouldnt it be safer to do this in a new issue?

alanpeixinho (commented):
Nice job

dashboard/src/utils/constants/general.ts
alanpeixinho: Another important point, we should check if this can impact kci-dev.

dashboard/src/components/FilterLabel/FilterLabel.tsx
```diff
@@ -0,0 +1,16 @@
+import type { JSX } from 'react';
+import { FormattedMessage } from 'react-intl';
+
+export function FilterLabel({ days }: { days: number }): JSX.Element {
```
alanpeixinho, line 4: Is it worth to add a new component, just for a a formatted message?

dashboard/src/utils/constants/general.ts
mentonin: removed this change and opened #1963

dashboard/src/components/FilterLabel/FilterLabel.tsx
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <p
+      className="text-dim-gray text-left text-xs"
+      data-test-id="hardware-filter-label"
+    >
+      <FormattedMessage
+        id="hardwareListing.latestCheckoutFilterLabel"
```
alanpeixinho: especially now, that we are adding this component on tree listing, the hardwareListing prefix might make us believe this component is only used in hardware page.

dashboard/src/utils/constants/general.ts
alanpeixinho: Fair.

dashboard/src/components/TreeListingPage/TreeTable.tsx
```diff
@@ -395,11 +397,16 @@ export function TreeTable({
...
         </BaseTable>
       </QuerySwitcher>
-      <PaginationInfo
-        table={table}
-        intlLabel="global.trees"
-        onPaginationChange={navigateWithPageSize}
-      />
+      <div className="flex flex-wrap items-start justify-between gap-4">
+        <FilterLabel days={intervalInDays ?? DEFAULT_TIME_SEARCH} />
+        <div style={{ marginLeft: 'auto' }}>
```
MarceloRobert: Avoid using `style` since we have tailwind. In this case, prefer `className="ml-auto"`

dashboard/src/pages/Hardware/HardwareTable.tsx
```diff
@@ -546,11 +550,18 @@ export function HardwareTable({
...
-      <PaginationInfo
-        table={table}
-        intlLabel="global.hardware"
-        onPaginationChange={navigateWithPageSize}
-      />
+      <div className="flex flex-wrap items-start justify-between gap-4">
+        {!selection && (
+          <FilterLabel days={intervalInDays ?? REDUCED_TIME_SEARCH} />
+        )}
+        <div style={{ marginLeft: 'auto' }}>
```
MarceloRobert: ditto `style`

dashboard/src/components/TreeListingPage/TreeTable.tsx
MarceloRobert: Resolved (I don't have permission to mark as resolved so I'll just comment)

dashboard/src/pages/Hardware/HardwareTable.tsx
MarceloRobert: Resolved

alanpeixinho (approved):
LGTM

PR #1965, opened 2026-06-30
Feat/1956/issue details last seen

felipebergamin:
- Implements last_seen information on issue details

Closes issue #1956 

Example of how it looks:

image

### DIFF
--- a/backend/kernelCI_app/helpers/issueExtras.py
+++ b/backend/kernelCI_app/helpers/issueExtras.py
@@ -3,10 +3,14 @@ from typing import List, Optional, Tuple
 
 from kernelCI_app.constants.general import UNCATEGORIZED_STRING
 from kernelCI_app.helpers.logger import log_message
-from kernelCI_app.queries.issues import get_issue_first_seen_data, get_issue_trees_data
+from kernelCI_app.queries.issues import (
+    get_issue_first_seen_data,
+    get_issue_last_seen_data,
+    get_issue_trees_data,
+)
 from kernelCI_app.typeModels.issues import (
     ExtraIssuesData,
-    FirstIncident,
+    Incident,
     IssueWithExtraInfo,
     ProcessedExtraDetailedIssues,
     TreeSetItem,
@@ -29,6 +33,19 @@ def parse_issue(issue_str: Optional[str]) -> tuple[str, Optional[int]]:
     return (issue_id, issue_version)
 
 
+def _incident_from_record(record: dict) -> Incident:
+    return Incident(
+        first_seen=record["first_seen"],
+        git_commit_hash=record["git_commit_hash"],
+        git_repository_url=record["git_repository_url"],
+        git_repository_branch=record["git_repository_branch"],
+        git_commit_name=record["git_commit_name"],
+        tree_name=record["tree_name"],
+        issue_version=record["issue_version"],
+        checkout_id=record["checkout_id"],
+    )
+
+
 def process_issues_extra_details(
     *,
     issue_key_list: List[Tuple[str, int]],
@@ -38,7 +55,7 @@ def process_issues_extra_details(
         return
 
     # TODO: combine both queries into one
-    assign_issue_first_seen(
+    assign_issue_incidents(
         issue_key_list=issue_key_list,
         processed_issues_table=processed_issues_table,
     )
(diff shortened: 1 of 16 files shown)


dashboard/src/locales/messages/index.ts
```diff
@@ -260,6 +261,7 @@ export const messages = {
     'issueDetails.issueDetails': 'Issue Details',
     'issueDetails.issueListingInfo':
       'The culprit for all issues listed is code',
+    'issueDetails.lastIncident': 'Last incident',
```
alanpeixinho: Just to keep the same pattern, we could go with 'Last Incident Data'

dashboard/src/types/issueExtras.ts
```diff
@@ -21,8 +21,20 @@ export type FirstIncident = {
   issue_version?: string;
 };
 
+export type LastIncident = {
```
alanpeixinho: Do we need a new type here? If there is no new data, we could just change to a type named Incident

backend/kernelCI_app/typeModels/issues.py
```diff
@@ -78,8 +78,20 @@ class FirstIncident(BaseModel):
     checkout_id: Optional[str]
 
 
+class LastIncident(BaseModel):
```
alanpeixinho: Do we need to differentiate types for first and last incident?

backend/kernelCI_app/queries/issues.py
```diff
@@ -293,6 +293,71 @@ def get_issue_first_seen_data(*, issue_id_list: list[str]) -> list[dict]:
...
+
+    cache_key = "issue_last_seen"
+    params = {"issue_id_list": issue_id_list}
+    records = get_query_cache(key=cache_key, params=params)
+
+    if records is None:
+        if len(issue_id_list) == 1:
+            comparison = "= %s"
+        else:
+            placeholders = ", ".join(["%s"] * len(issue_id_list))
```
alanpeixinho: I think we dont need to do this trick here. We might be able to pass as a list parameter.
There is an example of the commits parameter (a list) on get_hardware_details_summary.

backend/kernelCI_app/helpers/issueExtras.py
```diff
@@ -92,6 +101,39 @@ def assign_issue_first_seen(
...
 
+def assign_issue_last_seen(
+    *,
+    issue_key_list: List[Tuple[str, int]],
+    processed_issues_table: ProcessedExtraDetailedIssues,
+) -> None:
+    """
+    Assigns the last seen data to the processed_issues_table by querying with the issue_key_list.
+    """
+    issue_id_set: set[str] = set()
```
alanpeixinho: just a nit: we might be able to get the same with `issue_id_set = {*issue_key_list.keys()}`

backend/kernelCI_app/typeModels/issues.py
felipebergamin: not exactly. Now the only difference are the "last_seen" and "first_seen" properties. But I agree we can have an unified `Incident` type with this data

backend/kernelCI_app/queries/issues.py
```diff
@@ -293,6 +293,65 @@ def get_issue_first_seen_data(*, issue_id_list: list[str]) -> list[dict]:
     return records
 
 
+def get_issue_last_seen_data(*, issue_id_list: list[str]) -> list[dict]:
```
mentonin: This can be factored out with `get_issue_first_seen_data` into a single function with a toggle for asc/desc

backend/kernelCI_app/typeModels/issues.py
```diff
@@ -79,7 +79,8 @@ class FirstIncident(BaseModel):
 
 
 class ExtraIssuesData(BaseModel):
-    first_incident: FirstIncident
+    first_incident: Incident
+    last_incident: Optional[Incident] = None
```
mentonin: last incident should not be optional

dashboard/src/components/Section/FirstIncidentSection.tsx
```diff
@@ -1,78 +1,117 @@
...
+      <TooltipDateTime
+        dateTime={lastIncident.first_seen}
+        lineBreak={true}
+        showRelative={true}
+      />
+    ),
+  },
+];
 
 export const getFirstIncidentSection = ({
```
mentonin: lingering "First" in name - this gets first and last incidents

dashboard/src/locales/messages/index.ts
```diff
@@ -260,6 +261,7 @@ export const messages = {
     'issueDetails.issueDetails': 'Issue Details',
     'issueDetails.issueListingInfo':
       'The culprit for all issues listed is code',
+    'issueDetails.lastIncident': 'Last Incident Data',
```
mentonin, line 265: nit: I prefer "Latest Incident", as in "the issue might or might not be finished"

dashboard/src/locales/messages/index.ts
```diff
@@ -242,6 +242,7 @@ export const messages = {
     'hardwareListing.treeSelectorSearchPlaceholder': 'Search tree...',
     'issue.alsoPresentTooltip': 'Issue also present in {tree}',
     'issue.firstSeen': 'First seen',
+    'issue.lastSeen': 'Last seen',
```
mentonin, line 245: No need to specify "last" when we are inside a "last incident" section. Ditto for "first seen"

backend/kernelCI_app/queries/issues.py
felipebergamin: done

backend/kernelCI_app/typeModels/issues.py
felipebergamin: > last incident should not be optional

Agreed. It was optional before because `ExtraIssuesData` was being instantiated without `last_incident`, which was assigned later on. I changed `backend/kernelCI_app/helpers/issueExtras.py` to assign `last_incident` from the start.

alanpeixinho (approved):
Looks great

backend/schema.yml
```diff
@@ -648,15 +648,17 @@ paths:
           minimum: 0
           title: End Days Ago
           type: integer
-        description: Number of days ago that marks the end of the metrics interval
```
mentonin, line 651: Unrelated schema fixes should be in a separate commit, referencing the commits or PRs that introduced those schema changes

mentonin (changes requested):

mentonin (approved):

PR #1968, opened 2026-07-01
feat/labs migration

alanpeixinho:
## Summary

Migrates all read paths for lab information from JSONB `misc` fields (`builds.misc->>'lab'`,
`tests.misc->>'runtime'`) to the new `labs` table via `lab_id` foreign key, with JSONB fallback for rows not yet backfilled.
Every SQL query and Python helper that previously extracted lab names by parsing JSONB now uses `COALESCE(labs.name, misc_fallback)` through a `LEFT JOIN` on the `labs` table. This ensures:

* New data (with `lab_id` populated by the  ingester) reads from the structured FK
* Historical data (without `lab_id`) still works via the JSONB fallback
* After a full backfill, the JSONB fallbacks can be removed (all marked with `TODO` comments)

### legacy database

 - Use a database where `lab_id` is NULL on existing rows
 - Verify all pages display lab names as before:
   - Tree details page (boots/tests tabs) — lab column in test history
   - Tree commits history page — build lab and test lab in filters and rows
   - Hardware details page — lab in test/build summaries, history, and filters
   - Build details page — lab column in test list
   - Issue details page — lab column in test list
   - Notifications/metrics endpoint — lab summary counts

### Migrated database (test in a local database)

  * Run the migrations available
  * Run the ingester (which already fills lab_id)
    - Verify all pages display lab names as before:
    - Tree details page (boots/tests tabs) — lab column in test history
    - Tree commits history page — build lab and test lab in filters and rows
    - Hardware details page — lab in test/build summaries, history, and filters
    - Build details page — lab column in test list
    - Issue details page — lab column in test list
    - Notifications/metrics endpoint — lab summary counts


backend/kernelCI_app/queries/build.py
```diff
@@ -61,7 +69,12 @@ def get_build_tests(build_id: str) -> Optional[list[dict]]:
             "environment_compatible",
             "environment_misc",
             "build__status",
-            "lab",
+            "lab_name",
         )
     )
-    return list(result)
+    tests = []
+    for test in result:
```
mentonin: Why not assign directly to lab like before?

backend/kernelCI_app/views/treeCommitsHistory.py
```diff
@@ -84,12 +84,7 @@ def sanitize_rows(self, rows: dict) -> list:
         result = []
         for row in rows:
             build_misc = row[11]
-            sanitized_build_misc = sanitize_dict(build_misc)
```
mentonin, line 87: Why does this not use the fallback?

backend/kernelCI_app/models.py
```diff
@@ -119,6 +132,9 @@ class Builds(models.Model):
     log_url = models.TextField(blank=True, null=True)
     log_excerpt = models.CharField(max_length=16384, blank=True, null=True)
     misc = models.JSONField(blank=True, null=True)
+    lab = models.ForeignKey(
+        Labs, db_constraint=False, null=True, on_delete=models.DO_NOTHING
```
mentonin: Shouldn't we use db_constraint=True?

mentonin: ditto for Tests

mentonin: I see all other tables are using db_constraint=False, so there must be a good reason for this, but I do not see it documented

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -188,6 +191,39 @@ def prepare_file_data(
         }
 
 
+def assign_lab_ids(builds_buf: list[Builds], tests_buf: list[Tests]) -> None:
+    """Resolve each instance's real lab name to a labs.id and set its lab_id FK.
```
MarceloRobert, line 195: If the lab name is unique in the database, why not use that as a primary key? In that case you'd make a first query to check which labs are not in the database yet, a second query to update the table if needed, but you wouldn't need the third query to get the id of the newly added labs

backend/kernelCI_app/queries/tree.py
```diff
@@ -227,6 +227,7 @@ def get_tree_details_data(
                 tests.number_value AS tests_number_value,
                 tests.misc AS tests_misc,
                 tests.environment_compatible AS tests_environment_compatible,
+                COALESCE(test_labs.name, tests.misc ->> 'runtime') AS tests_lab,
```
MarceloRobert, line 231: I think you forgot a `TODO` here

backend/kernelCI_app/helpers/hardwareDetails.py
```diff
@@ -458,18 +450,16 @@ def handle_build_summary(
...
-        setattr(
-            builds_summary.labs[lab],
-            status_key,
-            getattr(builds_summary.labs[lab], status_key) + 1,
-        )
+    lab = record.get("build_lab") or UNKNOWN_STRING
+    build_lab_summary = builds_summary.labs.get(lab)
+    if not build_lab_summary:
+        build_lab_summary = StatusCount()
+        builds_summary.labs[lab] = build_lab_summary
```
mentonin, line 458: Copy pattern from previous chunk for consistence and readability

mentonin (dismissed):

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
alanpeixinho: Mostly to avoid string keys, since this is an internal key, we might benefit more from integer indices.

backend/kernelCI_app/models.py
alanpeixinho: For the kci keys, we unfortunately have some missing keys in a few rows. But since lab has only internal keys, I believe we can safely enforce constraints on them. Good point.

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
MarceloRobert: Makes sense. I am worried that we are adding one query here, one query there and soon we will have a bloat of unnecessary queries in the ingester, but it's fine for now

alanpeixinho: It is a valid concern. If we plain to move some operations to ingestion time, we certainly are going to end up increasing "ingestion complexity". But I believe that, as long as we keep the ingester fast enough, simplifying the analysis step should be the goal.

backend/kernelCI_app/queries/build.py
alanpeixinho: Mostly because lab is now the column name, but thinking again. I might change the column name for lab_id, this way we keep the lab name free to use, and follow more closely the the foreign key nomenclature.

backend/kernelCI_app/helpers/hardwareDetails.py
alanpeixinho: The point is, this is the original implementation (pre-refactor)

```python

    misc = sanitize_dict(build.misc) or {}
    lab = misc.get("lab", UNKNOWN_STRING)
    if lab:
        build_lab_summary = builds_summary.labs.get(lab)
        if not build_lab_summary:
            build_lab_summary = StatusCount()
            builds_summary.labs[lab] = build_lab_summary
        setattr(
            builds_summary.labs[lab],
            status_key,
            getattr(builds_summary.labs[lab], status_key) + 1,
        )

```
The conditional for lab is never false, because we are assigning UNKNOWN_STRING for lab.
However, it seems that this function is no longer used, and might be a legacy function used before performance improvements on HardwareDetails endpoints.
I will confirm it, and if this is the case, remove it.
We might as well plan to perform some dead code elimination, and check for similar cases.

backend/kernelCI_app/queries/build.py
alanpeixinho: Changed mind, to really avoid rename hacking on the ORM, it was easier to just add a proper sql query instead.

backend/kernelCI_app/views/treeCommitsHistory.py
alanpeixinho: Done

backend/kernelCI_app/queries/tree.py
alanpeixinho: Done

backend/kernelCI_app/queries/build.py
```diff
@@ -1,9 +1,11 @@
+import json
 from typing import Optional
 
-from django.db.models.expressions import F
+from django.db import connection
```
MarceloRobert: very nit: prefer `connections['default']`

MarceloRobert (commented):
Haven't tested yet, but looks good so far

backend/kernelCI_app/queries/build.py
alanpeixinho: Done

backend/kernelCI_app/helpers/treeDetails.py
```diff
@@ -495,8 +493,9 @@ def process_filters(instance, row_data: dict, skip_build_filters: bool = False)
         instance.global_architectures.add(row_data["build_architecture"])
         instance.global_compilers.add(row_data["build_compiler"])
         instance.unfiltered_origins["build"].add(row_data["build_origin"])
-        if (build_misc := row_data["build_misc"]) is not None:
-            instance.unfiltered_labs["build"].add(build_misc.get("lab", UNKNOWN_STRING))
+        instance.unfiltered_labs["build"].add(
+            row_data.get("build_lab") or UNKNOWN_STRING
+        )
```
felipebergamin, line 499: here the code we talked about. Not sure if this conditional should be removed

backend/kernelCI_app/queries/notifications.py
```diff
@@ -926,19 +926,20 @@ def get_metrics_data(
     ORDER BY inc.origin, total DESC
     """
 
+    # TODO: remove t.misc.runtime after backfill on lab_id columns
     lab_summary_query = """
-    -- get count of tests of each lab and how many builds are related to those tests
     SELECT
-        t.misc->>'runtime' AS lab,
+        -- TODO remove misc->>'runtime' fallback after lab backfill
+        COALESCE(l.name, t.misc->>'runtime', t.origin) AS lab,
```
felipebergamin: why t.origin here ?
If we have neither a laboratory nor a runtime, won't the `WHERE` clause you wrote will filter out the record?

backend/kernelCI_app/helpers/treeDetails.py
alanpeixinho: This one is tricky.
Post-migration we no longer read build_misc, so we can't distinguish "no misc at all" from "misc present but no lab" — we only have the lab field. And it's odd to branch on how lab is missing.
I think we should standardize: any missing lab as UNKNOWN.

alanpeixinho: If we do find a reason to treat them differently, we should include an explicit rule on this.

backend/kernelCI_app/queries/notifications.py
alanpeixinho: Nice catch. Missed removing some test here.

alanpeixinho: Fixed

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -188,6 +191,39 @@ def prepare_file_data(
...
+                    "INSERT INTO labs (name) VALUES (%s) ON CONFLICT (name) DO NOTHING",
+                    [(name,) for name in missing],
+                )
+                cursor.execute(
+                    "SELECT id, name FROM labs WHERE name = ANY(%s)", [missing]
+                )
+                id_map.update({name: lab_id for lab_id, name in cursor.fetchall()})
+
+    for obj in objs:
+        obj.lab_id = id_map.get(obj._lab_name) if obj._lab_name else None
```
tales-aparecida, line 224: I was a bit concerned with the 3 queries, so I've asked gemini to try to fold them into a single query, please review

```suggestion
def assign_lab_ids(builds_buf: list[Builds], tests_buf: list[Tests]) -> None:
    """Resolve each instance's real lab name to a labs.id and set its lab_id FK.

    Uses a single optimized CTE join to insert missing records and fetch all IDs 
    in 1 round-trip without sequence waste or ORM overhead.
    """
    objs = [*builds_buf, *tests_buf]
    names = list({obj._lab_name for obj in objs if getattr(obj, "_lab_name", None)})

    if not names:
        for obj in objs:
            obj.lab_id = None
        return

    query = """
        WITH input_names AS (
            -- 1. Unnest and deduplicate inputs in Postgres C-memory
            SELECT DISTINCT unnest(%s::text[]) AS name
        ),
        inserted AS (
            -- 2. Anti-join via LEFT JOIN / IS NULL (Faster than NOT EXISTS)
            INSERT INTO labs (name)
            SELECT i.name 
            FROM input_names i
            LEFT JOIN labs l ON l.name = i.name
            WHERE l.name IS NULL
            RETURNING id, name
        )
        -- 3. Combine newly inserted IDs with existing IDs
        SELECT id, name FROM inserted
        UNION ALL
        SELECT l.id, l.name
        FROM labs l
        JOIN input_names i ON l.name = i.name;
    """

    with connections["default"].cursor() as cursor:
        cursor.execute(query, [names])
        # Unpacks (id, name) tuples into a {name: id} dictionary directly
        id_map = dict(cursor.fetchall())

    for obj in objs:
        lab_name = getattr(obj, "_lab_name", None)
        obj.lab_id = id_map.get(lab_name) if lab_name else None
```

felipebergamin (approved):

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
felipebergamin: great optimization but I think it might burn the id sequence in some cases, right? which is the concern of the original approach.

tales-aparecida: Actually, it doesn't...

but it's great that you mentioned that! We still need need `ON CONFLICT (name) DO NOTHING` like we have in the original code to avoid the race condition.

Only when the race condition happens there will be ID burning on both the original and using the single query

alanpeixinho: This is tricky, because in the original solution, despite having tree queries, only the first simpler query would hit most of the time (since we expect to have the number of labs being significantly smaller than the number of builds or tests ).
With the unified query we are running 100% of the time a more complex query (I will take a look how this impact performance).

alanpeixinho: On the topic of race condition (thanks for catching that), I believe this can be solved with a lock in the solution with split queries.

alanpeixinho: I was reviewing this PR again, and realized that we are in fact safe from race condition here.
The do nothing on conflict seems to really be enough.

tales-aparecida: > The do nothing on conflict seems to really be enough.

Oh yes, sorry if I was not clear! My point was that including the `DO NOTHING` in my suggestion would prevent the race condition, like it is preventing it in the original code

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -245,6 +281,8 @@ def flush_buffers(
     if total == 0:
         return
 
+    assign_lab_ids(builds_buf, tests_buf)
```
felipebergamin: `assign_lab_ids` handles database operations. Shouldn't it be inside a try/catch to be properly handled? I don't see a try/catch that will capture this error, so an error here might kill the process

alanpeixinho: Done

backend/kernelCI_app/models.py
alanpeixinho: Done

felipebergamin:
@alanpeixinho I think we still have one place using old misc->lab

aggregation_helpers.py:55

alanpeixinho:
> @alanpeixinho I think we still have one place using old misc->lab
> 
> aggregation_helpers.py:55

Thanks. I fixed this one, and also got another case on process_submissions.py

felipebergamin (approved):
LGTM. tested dashboard + ingester + migrating with a db snapshot.
also couldn't find any other place using labs from misc

PR #1970, opened 2026-07-03
feat/prune age

alanpeixinho:
## Summary

Add `prune_db`, a manual management command to delete old data and reclaim space.

* Deletes checkouts, builds and tests older than a given age (e.g. `--older-than "30 days"`).
* Manual cascade (models use `DO_NOTHING`): pruning a checkout removes its builds, pruning a build removes its tests — even when children are newer.
* `--origins` scopes the age filter; cascade stays origin-agnostic so no orphans are left.
* Batched deletes (`--batch-size`) that each commit on their own to keep locks short.
* `--dry-run` and a per-table confirmation prompt showing exactly how many rows will be affected (skip with `--yes`).
* Extracts `parse_interval` from `update_db` into a shared helper.

## Usage

```bash
python manage.py prune_db --older-than "30 days" --dry-run                                                                                                                                                         
python manage.py prune_db --older-than "30 days" --origins maestro                                                                                                                                          
```

## How to Test

  * In a local database create newly seeded data `poetry run python3 manage.py seed_data --clear --yes`
  * Run command `poetry run python3 manage.py prune_db --dry-run` with different arguments for older-than parameter.
  * Verify the number of selected rows matches the database.
  * Run command `poetry run python3 manage.py prune_db` with different arguments for older-than parameter.
  * Verify the rows are properly deleted.

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/intervals.py
+++ b/backend/kernelCI_app/management/commands/helpers/intervals.py
@@ -1,0 +1,25 @@
+from datetime import datetime, timedelta
+
+from django.utils import timezone
+
+
+def parse_interval(interval_str: str) -> datetime:
+    """Turns an 'x days'/'x hours'/'x minutes' interval into an absolute
+    timestamp relative to now (now - interval)."""
+    parts = interval_str.split()
+    if len(parts) != 2:
+        raise ValueError(f"Invalid interval format: {interval_str}")
+
+    value, unit = parts
+    value = int(value)
+
+    if unit.lower() in ["minute", "minutes"]:
+        delta = timedelta(minutes=value)
+    elif unit.lower() in ["hour", "hours"]:
+        delta = timedelta(hours=value)
+    elif unit.lower() in ["day", "days"]:
+        delta = timedelta(days=value)
+    else:
+        raise ValueError(f"Unsupported time unit: {unit}")
+
+    return timezone.now() - delta
--- a/backend/kernelCI_app/management/commands/prune_db.py
+++ b/backend/kernelCI_app/management/commands/prune_db.py
@@ -1,0 +1,220 @@
+"""
+Management command to prune old builds, tests and checkouts.
+
+Removes checkouts, builds and tests older than a given age. To keep referential
+integrity, deletion cascades manually (models use DO_NOTHING): a removed
+checkout drags its builds, and a removed build drags its tests, even when those
+children are newer than the cutoff.
+
+Rows linked to an incident (an issue) are kept by default, together with their
+ancestors so nothing is orphaned; pass --skip-issue-protection to prune them too.
+
+Only checkouts, builds and tests are touched. Aggregate and derived tables (e.g.
+tree_tests_rollup, hardware_status, latest_checkout) are left untouched and must
+be cleaned up separately.
+"""
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import connections
+
+from kernelCI_app.management.commands.helpers.intervals import parse_interval
+
+PRUNABLE_TABLES = ("checkouts", "builds", "tests")
+
+
+class Command(BaseCommand):
+    help = "Prune checkouts, builds and tests older than a given age"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--older-than",
+            type=str,
+            required=True,
+            help="Delete rows older than this age ('x days' or 'x hours' format, "
+            "e.g. '30 days')",
+        )
+        parser.add_argument(
+            "--origins",
+            type=lambda s: [origin.strip() for origin in s.split(",")],
+            default=[],
+            help="Limit age-based pruning to specific origins (comma-separated). "
+            "Children of pruned parents are removed regardless of origin. "
+            "If not provided, any origin is considered.",
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be deleted without actually deleting",
+        )
+        parser.add_argument(
+            "--yes",
+            action="store_true",
+            help="Skip the confirmation prompt and delete right away",
+        )
+        parser.add_argument(
+            "--batch-size",
+            type=int,
+            default=10000,
+            help="Number of rows to delete per batch (default: 10000)",
+        )
+        parser.add_argument(
+            "--tables",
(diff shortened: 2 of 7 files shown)


backend/kernelCI_app/management/commands/helpers/intervals.py
```diff
@@ -0,0 +1,25 @@
+from datetime import datetime, timedelta
```
mentonin, line 1: Django has a built-in parser `django.utils.dateparse.parse_duration` that could be used here

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,148 @@
...
+            cutoff = parse_interval(options["older_than"])
+        except ValueError as e:
+            raise CommandError(str(e)) from e
+
+        dry_run = options["dry_run"]
+        origins = options["origins"]
+
+        # Values are always passed as query parameters; only static SQL fragments are
+        # interpolated. Named parameters let us reuse the same dict across every query.
+        params = {"cutoff": cutoff, "batch_size": options["batch_size"]}
```
felipebergamin: a minor but I think that would be nice to validate the batch_size. Currently we are able to pass a negative value to the database.

```
django.db.utils.DataError: LIMIT must not be negative
```

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,148 @@
+"""
+Management command to prune old builds, tests and checkouts.
+
+Removes checkouts, builds and tests older than a given age. To keep referential
+integrity, deletion cascades manually (models use DO_NOTHING): a removed
+checkout drags its builds, and a removed build drags its tests, even when those
+children are newer than the cutoff.
+"""
```
felipebergamin, line 15: Worth noting in docs that aggregate tables are not touched by this command.

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Done.

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Done

backend/kernelCI_app/management/commands/helpers/intervals.py
alanpeixinho: I took a look on the built-in parser. Unfortunately it has different standards from what we are following.
  * standard: "30 00:00:00" (%d %H:%M:%S)
  * ISO 8601: "P30D"
  * postgres: "3 days 04:05:06" (only works for days)
I believe we might benefit more of keeping our custom parser, we could even extend it to include weeks or months, especially in a scenario where we are working with longer ranges.
What you think?

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,259 @@
...
+                if protect_incidents:
+                    lines.append("Note: rows linked to an incident are kept.")
+                self.stdout.write("\n".join(lines))
+
+                if total == 0:
+                    self.stdout.write(self.style.SUCCESS("Nothing to prune."))
+                    return
+
+                if dry_run:
+                    self.stdout.write(
```
mentonin: Should the warning go to stderr instead?

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,259 @@
...
+                            "execute."
+                        )
+                    )
+                    return
+
+                if not options["yes"]:
+                    summary = ", ".join(f"{counts[t]} {t}" for t in selected_tables)
+                    try:
+                        answer = (
+                            input(f"Delete {summary} ({total} rows total)? [y/N] ")
```
mentonin: We already printed counts + total, can we make this shorter?

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,259 @@
...
+                            .strip()
+                            .lower()
+                        )
+                    except EOFError:
+                        answer = ""
+                    if answer not in ("y", "yes"):
+                        self.stdout.write("Aborted.")
+                        return
+
+                # Delete child-first (reverse of PRUNABLE_TABLES order): each batch
```
mentonin, line 159: Document the strict parent-before-child order of PRUNABLE_TABLES at its definition

backend/kernelCI_app/management/commands/prune_db.py
```diff
@@ -0,0 +1,259 @@
...
+    def _count(self, cursor, temp_table):
+        cursor.execute(f"SELECT COUNT(*) FROM {temp_table}")
+        return cursor.fetchone()[0]
+
+    def _batch_delete(self, cursor, table, temp_table, batch_size):
+        sql = (
+            f"WITH batch AS ("
+            f"DELETE FROM {temp_table} WHERE id IN "
+            f"(SELECT id FROM {temp_table} LIMIT %(batch_size)s) RETURNING id"
+            f") DELETE FROM {table} WHERE id IN (SELECT id FROM batch)"
```
felipebergamin: even we are filtering unknown tables from command args with PRUNABLE_TABLES I think we should go safer here and quote the identifiers. DELETE FROM "{table}"

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Done

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Good point. Changed to simple message to confirm.

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Done

backend/kernelCI_app/management/commands/prune_db.py
alanpeixinho: Done

mentonin (approved):
LGTM

small detail:
c74f7134 is missing an empty line between message header and body (it is all header)

felipebergamin (approved):

PR #1977, opened 2026-07-06
fix: Handle undefined histories in test details

### LINKED ISSUE #1167: Bug: TestStatusHistory not shown for independent tests
Tests occur in a build, and builds happen in a checkout, and checkouts have an origin. If tests have a build_id that doesn't belong to any build in the database, then we can't tell which origin a test is from.

**Problem:** When tests are not related to an existing build (or are related to a build but the build is not related to an existing checkout) - here called as an "independent test" - they don't have a direct way to check for its origin, and the test status history doesn't load.

**Expected behaviour:** the test status history should load even if it doesn't have an origin

**Example:** [test details page](https://dashboard.kernelci.org/test/redhat%3A1755381452-ppc64le-kernel-debug_upt_15)

**Possible solution:** change the test status history endpoint to allow for a None origin and don't join for origins when performing the query

mentonin:
## Summary

This PR fixes the dashboard behavior on `/test/<test_id>` for tests without a meaningful history.

Current behavior: Shows "Status History" as loading for a long while, hides on failure, and retries many times.

New behavior: Shows a message under "Status History" indicating that it is not available for this test

## Testing

Navigate to `test/<test_id>` and compare with production for tests with complete or missing data

Example: [test missing git_repository_url](https://dashboard.kernelci.org/test/broonie:0a40c6fbd105802fbbcaadca249e0948fbf8095a-sirena-2791999)

## Context

Tests in the same history must match (at least) on origin, tree, path, and build config for their comparison to be meaningful. Many tests do not have all of this information available (because of a missing build, checkout, or incomplete submission), so we can't track a meaningful history for them.

## Related Issues
- Closes #1167
- Related to #1976

### DIFF
--- a/dashboard/src/api/testDetails.ts
+++ b/dashboard/src/api/testDetails.ts
@@ -1,5 +1,5 @@
 import type { UseQueryResult } from '@tanstack/react-query';
-import { useQuery } from '@tanstack/react-query';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
 
 import type {
   TTestDetails,
@@ -9,6 +9,7 @@ import type {
 
 import type { TIssue } from '@/types/issues';
 import type { ApiUseQueryOptions } from '@/types/api';
+import { isBadRequestError, DEFAULT_QUERY_RETRY_COUNT } from '@/utils/query';
 
 import { RequestData } from './commonRequest';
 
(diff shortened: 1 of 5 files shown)


dashboard/src/utils/query.ts
```diff
@@ -19,3 +23,16 @@ export const retryHandler = (retryCount: number | boolean = 3) => {
...
+
+export const isBadRequestError = (error: unknown): boolean => {
+  if (error instanceof AxiosError) {
+    return error.response?.status === HttpStatusCode.BadRequest;
+  }
+
+  if (error instanceof Error) {
+    const statusCode = Number(error.message.split(':')[0]);
+    return statusCode === HttpStatusCode.BadRequest;
+  }
```
felipebergamin: is this really necessary? I don't like relying on the message string to infer http statuses

dashboard/src/utils/query.ts
```diff
@@ -1,3 +1,7 @@
+import { AxiosError, HttpStatusCode } from 'axios';
+
+export const DEFAULT_QUERY_RETRY_COUNT = 3;
+
 export const retryHandler = (retryCount: number | boolean = 3) => {
```
alanpeixinho: might be a nit: but we are using a hardcoded default argument here, that might differ from the constant we introduced.

dashboard/src/api/testDetails.ts
```diff
@@ -64,9 +65,30 @@ const fetchTestStatusHistory = async (
 export const useTestStatusHistory = (
   params?: TestStatusHistoryParams,
 ): UseQueryResult<TestStatusHistory> => {
+  const queryClient = useQueryClient();
+
   return useQuery({
     queryKey: ['testStatusHistoryData', params],
     queryFn: () => fetchTestStatusHistory(params),
+    retry: (failureCount, error) => {
+      if (isBadRequestError(error)) {
```
alanpeixinho: This might be a sensible default (to not retry on bad request).
If so, maybe we could just include the check for isBadRequest on the default retryHandler?

alanpeixinho (approved):
Pretty nice

felipebergamin (approved):
lgtm.

PR #1978, opened 2026-07-06
fix: remove generated schema.yml file from git

mentonin:
This file is being removed as it can be generated from the backend code and is an unecessary burden to manually sync in git.

- Closes #1910


MarceloRobert (commented):
I don't see much reason to remove this from the git but it's fine for me. I think that we can rephrase the readme though, it currently says "Please use it whenever the ins and outs of endpoints change." and I guess that's not going to be needed anymore.

PR #1981, opened 2026-07-08
chore: manage docker bloat on deployments

mentonin:
automatically prune old images and raise warnings for dangling volumes and exited containers on GitHub Actions workflow deployments.

Closes #1918

### DIFF
--- a/.github/workflows/deploy-production.yaml
+++ b/.github/workflows/deploy-production.yaml
@@ -144,7 +144,7 @@ jobs:
         run: |
           eval $(ssh-agent -s)
           echo "$SSH_KEY" | ssh-add - >/dev/null
-          ssh "${SSH_USER}@${SSH_HOST}" <<<'EOF'
+          ssh "${SSH_USER}@${SSH_HOST}" <<'EOF'
             set -eu
             docker image prune -fa --filter "until=48h"
             docker builder prune -f --filter "until=48h"
--- a/.github/workflows/deploy-production.yaml
+++ b/.github/workflows/deploy-production.yaml
@@ -88,7 +88,8 @@ jobs:
     needs: start-notification
     runs-on: ubuntu-latest
     steps:
-      - name: Configure host authenticity
+      - &ssh-host-auth
+        name: Configure host authenticity
         run: |
           mkdir -p ~/.ssh/ && chmod 700 ~/.ssh/
           touch ~/.ssh/known_hosts && chmod 600 ~/.ssh/known_hosts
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-production.yaml
```diff
@@ -106,8 +106,10 @@ jobs:
             cp ~/.env-production dashboard-production/.env &&
             cd dashboard-production &&
             git checkout ${GITHUB_SHA} &&
-            docker compose -f docker-compose-next.yml pull &&
-            docker compose -f docker-compose-next.yml up -d
+            docker compose up --pull=always --remove-orphans --detach &&
```
alanpeixinho: why are we changing the docker compose file here?
Shouldn't we be pointing to `docker-compose-next.yml`?

.github/workflows/deploy-staging.yaml
```diff
@@ -67,9 +67,10 @@ jobs:
             cp ~/.env-staging dashboard-staging/.env &&
             cd dashboard-staging &&
             git checkout ${GITHUB_SHA} &&
-            docker compose down &&
-            docker compose build --no-cache &&
```
alanpeixinho, line 71: Even with the policy of always pulling, isn´t it safer to build with no-cache here?

.github/workflows/deploy-production.yaml
alanpeixinho, line 100: I am just starting a discussion, do not take it as the best approach.
Wouldn´t it be better to leave the docker prune as a separate step here?
This way could even make it as an optional step, and if the cleanup fails the deploy process can still take place. Or would be better to fail imediatelly in such case?

alanpeixinho (commented):
Nice job

.github/workflows/deploy-production.yaml
alanpeixinho: Also, should we also remove docker builder cache? `docker builder prune -f --filter "until=48h"`

.github/workflows/deploy-production.yaml
mentonin: fixed

.github/workflows/deploy-staging.yaml
mentonin: Fixed

.github/workflows/deploy-production.yaml
mentonin: I tried to avoid opening a second SSH connection, a separate step is indeed better. I moved them to another job altogether, so the status of the deployment job actually means whether deployment happened.

.github/workflows/deploy-production.yaml
```diff
@@ -129,3 +130,24 @@ jobs:
...
+      - *ssh-host-auth
+      - name: Prune docker artifacts on remote
+        env:
+          SSH_USER: ${{ secrets.STAGING_USER }}
+          SSH_HOST: ${{ secrets.STAGING_HOST }}
+          SSH_KEY: ${{ secrets.STAGING_KEY }}
+        run: |
+          eval $(ssh-agent -s)
+          echo "$SSH_KEY" | ssh-add - >/dev/null
+          ssh "${SSH_USER}@${SSH_HOST}" <<<'EOF'
```
alanpeixinho: just checking: is the triple `<<<` correct here?

alanpeixinho (approved):
Just left one comment to confirm.
Overall, it looks fine.

.github/workflows/deploy-production.yaml
mentonin: Nope, should be `<<'EOF'`, I added a commit with the fix

alanpeixinho (approved):
Nice job

PR #1984, opened 2026-07-10
chore: update pyproject.toml to tool-agnostic standard

mentonin:
Uses the PEP621-defined `pyproject.toml` format. This is a tool-agnostic, poetry-compatible (since version 2) format. Advantages include better compatibility with other tools and no lock-in to poetry. poetry (through Poetry.lock) are still the sources of truth for versions we deploy/test against


backend/pyproject.toml
```diff
@@ -46,6 +46,9 @@ markers = [
     "performance: marks tests as performance/benchmark tests (very slow, requires full setup)"
 ]
 
+[tool.poetry]
+package-mode = false
+
 [build-system]
-requires = ["poetry-core"]
+requires = ["poetry-core>=1.0.0"]
```
alanpeixinho: shouldnt we go for poetry-core>=2.0?
it seems they implemented some of the pep621 tags only at that version (`project` for instance)
https://python-poetry.org/blog/announcing-poetry-2.0.0

backend/pyproject.toml
```diff
@@ -1,41 +1,41 @@
-[tool.poetry]
+[project]
 name = "backend"
 version = "0.1.0"
 description = "Django Backend"
-authors = ["Your Name <you@example.com>"]
```
alanpeixinho, line 5: nice catch, didnt even notice this before.

backend/pyproject.toml
```diff
@@ -1,41 +1,41 @@
...
+requires-python = ">=3.12,<4"
```
alanpeixinho, line 5: This might be a problem.
There is some dependency failing for python 3.14 that has not been addressed yet.
```
  - Installing rpds-py (0.23.1): Failed
```
Which seems to be a problem with pyo3 ffi module.
So, it might be safer to not allow newer python versions until we fix it.

backend/pyproject.toml
```diff
@@ -46,6 +46,9 @@ markers = [
     "performance: marks tests as performance/benchmark tests (very slow, requires full setup)"
 ]
 
+[tool.poetry]
+package-mode = false
```
alanpeixinho, line 50: nice

backend/pyproject.toml
mentonin: I set it to 2.4.0 instead as that is what we are using, and version 2.2 would be needed for `[dependency-groups]` table support anyway. In truth, the `[build-system]` table is not strictly necessary (we are not packaging the project), but it makes our project clearer and supports installation with pip/setup-tools if that is wanted for any reason.

backend/pyproject.toml
mentonin: That is a dependencies issue, really;

rpds-py supports python 3.14 starting in version 0.25.0 (we have 0.23.1 in lock-file). We depend on it through `jsonschema`, which is required by `drf-spectacular` and `kcidb-io`. `jsonschema` requires `rpds-py>=0.25.0` as of version `4.26.0`, so we actually should require versions of `drf-spectacular` and `kcidb-io` which in turn require `jsonschema>=4.26.0`.

felipebergamin: so we set `requires-python = ">=3.12,<3.14"` and open a new issue to fix the deps?

alanpeixinho (approved):
LGTM

backend/pyproject.toml
mentonin: > so we set `requires-python = ">=3.12,<3.14"` and open a new issue to fix the deps?

Actually, the last commit in this PR already addresses that

alanpeixinho: @mentonin Tried it here, but unfortunately it still failed to install dependencies on the python3.14 even after the bump.

```
Package operations: 1 install, 0 updates, 0 removals                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                               
  - Installing rpds-py (0.23.1): Failed                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                               
PEP517 build of a dependency failed                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                               
Backend subprocess exited when trying to invoke build_wheel   
```

We can compare our environments and check if there is still some mismatch we need to address.

felipebergamin (approved):

backend/pyproject.toml
mentonin: @alanpeixinho I somehow did not commit the poetry lock update for rpds.py; just force-pushed the correct poetry lock

PR #1987, opened 2026-07-13
reports: limit regzbot tags and lore link to upstream trees

bhcopeland:
Regression reports currently carry the #regzbot tags and a lore link to their own Message-ID regardless of the tree, but only mainline and next reports are CC'd to the regressions list. Reports for other trees (android, chromeos and so on) are out of regzbot's scope, and their lore link can point at a mail that never reached lore.

Gate the regzbot tags and the lore link on the same tree check that adds the regressions list CC, through a shared is_regzbot_tree() helper, so the tags only appear on reports that regzbot will actually see and the link only appears on mails guaranteed to be archived. Reports for other trees keep the Reported-by attribution and the

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/common.py
+++ b/backend/kernelCI_app/management/commands/helpers/common.py
@@ -13,7 +13,7 @@ KERNELCI_RESULTS = "kernelci-results@groups.io"
 KERNELCI_REPLYTO = "kernelci@lists.linux.dev"
 REGRESSIONS_LIST = "regressions@lists.linux.dev"
 
-REGZBOT_TREES = {"mainline", "next"}
+REGZBOT_TREES = {"mainline", "next", "stable"}
 
 
 def is_regzbot_tree(tree_name) -> bool:
--- a/backend/kernelCI_app/management/commands/helpers/common.py
+++ b/backend/kernelCI_app/management/commands/helpers/common.py
@@ -13,6 +13,18 @@ KERNELCI_RESULTS = "kernelci-results@groups.io"
 KERNELCI_REPLYTO = "kernelci@lists.linux.dev"
 REGRESSIONS_LIST = "regressions@lists.linux.dev"
 
+REGZBOT_TREES = {"mainline", "next"}
+
+
+def is_regzbot_tree(tree_name) -> bool:
+    """Trees whose regression reports are tracked by regzbot.
+
+    Reports for these trees are CC'd to the regressions list, which is
+    archived on lore and followed by regzbot, so only they carry the
+    #regzbot tags and a lore link to their own Message-ID.
+    """
+    return tree_name in REGZBOT_TREES
+
 
 def setup_jinja_template(template_name: str) -> Template:
     """Gets the template file from management/commands/templates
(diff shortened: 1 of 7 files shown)


alanpeixinho (approved):
Tested for different issues from different trees, seems ok.

bhcopeland:
@alanpeixinho I thought it would make sense to put stable in here. I've added it as an additional commit, since it isn't much to add.

backend/kernelCI_app/management/commands/helpers/common.py
```diff
@@ -13,6 +13,18 @@
 KERNELCI_REPLYTO = "kernelci@lists.linux.dev"
 REGRESSIONS_LIST = "regressions@lists.linux.dev"
 
+REGZBOT_TREES = ("mainline", "next", "stable")
```
tales-aparecida: nit https://docs.astral.sh/ruff/rules/literal-membership/

```suggestion
REGZBOT_TREES = {"mainline", "next", "stable"}
```

tales-aparecida (approved):

backend/kernelCI_app/management/commands/helpers/common.py
bhcopeland: Thanks! Amended and squashed; appreciated.

tales-aparecida (approved):

mentonin (approved):

backend/kernelCI_app/management/commands/notifications.py
```diff
@@ -158,11 +159,13 @@ def ask_ignore_issue():
 
 def generate_build_issue_report(issue, incidents, message_id=None):
     template = setup_jinja_template("issue_build.txt.j2")
+    regzbot = is_regzbot_tree(issue["tree_name"])
```
knurd: Nitpicking: wondering if this broad variable name might become somewhat ambiguous/confusing over time if the code gains more regzbot features

backend/kernelCI_app/tests/unitTests/commands/fixtures/report_notifications_data.py
```diff
@@ -66,10 +66,10 @@ def make_build_issue(*, with_range: bool = True):
     return issue, [incident]
 
 
-def make_boot_issue(*, with_range: bool = True):
+def make_boot_issue(*, with_range: bool = True, tree_name: str = "mainline"):
     """Returns an (issue, incidents) pair for a boot regression report."""
     issue = {
-        "tree_name": "mainline",
+        "tree_name": tree_name,
         "git_repository_branch": "master",
```
knurd: This got me thinking: is this "master" in places like this (there is another one a little bit earlier) still correct with the stable support added in the second patch? Wondering, as the various stable series are maintained in different branches of the same tree.

knurd (approved):
Don't really know this codebase, but looks fine from what I see. Just spotted two thing I wanted to mention, but they are not ciritical.

backend/kernelCI_app/management/commands/notifications.py
bhcopeland: Good point, I'll rename to regzbot_tracked

backend/kernelCI_app/tests/unitTests/commands/fixtures/report_notifications_data.py
bhcopeland: Thats a good point yes. Folded into the second commit (now fixtures take a branch, defaults to master) and the stable test now renders stable/linux-6.15.y. So both should be caught. If you could re-review

knurd: LGTM (with my limited knowledge of the codebase), thx

Issue #1969, opened 2026-07-02
Records pruning by age


nuclearcat:
As per discussion during meetings about storage policy we keep data not longer than 6 month. Recently we also hit storage limit on Azure database, so this question becoming high priority. 
I increased storage twice, but we need to implement records cleanup, so we delete records older than 6 month.

alanpeixinho:
Closed by #1970

PR #1990, opened 2026-07-14
fix(hardware): use hardware_status window for listing and details trees

### LINKED ISSUE #1264: Bug: Hardware details doesn't show tests from older checkouts
Detached from #1207 

Since we have a link to the hardware details page from the test details page, it is possible that a test is related to a tree in which the latest checkout has no tests for that hardware.

Tale example: test A was related to tree X in the checkout C1 with the hardware H, if the tree X moves on to checkout C2 but C2 has no tests with the hardware H, then when I access the hardware details for H it will show nothing, seemingly "not related to any checkout" because the latest checkouts from the trees that that hardware appears have 0 tests with H.

Real example: [test details in localhost](http://localhost:5173/test/ti%3Aae81ca814a39c130c9f62930) and [hardware details in localhost](http://localhost:5173/hardware/k3-am625-sk?et=1746457200&st=1746025200) - The database must be playground_kcidb.

### LINKED ISSUE #1705: dashboard test  result is not in sync with Maestro
dashboard test  result is not in sync with Maestro
one example :
https://kernelci-api.westus3.cloudapp.azure.com/viewer?node_id=696208facbfd84c3cde77145
this is not visble on https://dashboard.kernelci.org/hardware-new?hs=qcom&i=30

As per @AmadeusK525 Seems like this is an issue with the Hardware itself? https://dashboard.kernelci.org/hardware/x1e80100?et=1768912200&st=1768480200
@Gustavo Flores @Marcelo Robert will look into it.

### LINKED ISSUE #1983: Hardware missing from Hardware table
From discord:

Ben Copeland — Yesterday at 17:24
How come https://dashboard.kernelci.org/tree/mainline/master/8cdeaa50eae8dad34885515f62559ee83e7e8dda?df%7Ch%7Cexynos850-e850-96=true&o=linaro_pull_labs&p=bt no longer shows up on hardware? It used to, when I search https://dashboard.kernelci.org/hardware?hs=exynos850 on hardware I get no results anymore

Tales — Yesterday at 18:04
some quirk with origin filtering maybe? linaro_pull_labs in the build and test getting dropped from the search? It appears when using the new filters, so I don't think it's not present in the table https://dashboard.kernelci.org/hardware?ch=8cdeaa50eae8dad34885515f62559ee83e7e8dda&gb=master&gu=https%3A%2F%2Fgit.kernel.org%2Fpub%2Fscm%2Flinux%2Fkernel%2Fgit%2Ftorvalds%2Flinux.git&t=mainline

Let me know if you want any more details

alanpeixinho:
### What it is
* Hardware listing no longer joins `latest_checkout`, so boards tested inside the selected time window still appear even when they were not on the newest tip.
* Hardware details tree heads no longer use absolute tree tips. `get_hardware_trees_head_commits` / `get_hardware_trees_data` resolve the latest checkout per tree from `hardware_status` in the window, so boards with tests on older checkouts still show data.
* Expands the hardcoded Hardware `TEST_ORIGINS` list so origins such as `linaro_pull_labs` (and other active pull-lab origins) show in the origin dropdown.

### How to test
* Make sure origin dropdown now includes `linaro_pull_labs`, `pullab_cloud_aws_arm64`, and `pull_labs_aws_ec2`.
* Make sure hardware that are not tested on the latest checkout, but were tested in the specified time window (default is 5 days, but can be changed via `?i` query parameter) are listed. (exynos on `linaro_pull_labs` is the example highlighted on issue #1983)
* Open hardware details for a board that has tests on an older checkout but not on the current tip (e.g. cases from #1264 / #1705). Trees and summary should show that older checkout’s data instead of empty / “no commits”.

Closes #1983
Closes #1264
Closes #1705


backend/kernelCI_app/queries/hardware.py
```diff
@@ -1,7 +1,7 @@
 from datetime import datetime
 from typing import Optional, TypedDict
 
-from django.db import connection
+from django.db import connection, connections
```
mentonin, line 4: If we are going from connection to connections["default"] in lines that would otherwise be left unchanged, why not do it everywhere (or at least everywhere on the touched file)? Or use a `connection = connections["default"]` approach, if you prefer not to touch too many lines

Otherwise, why not keep using `connection`?

backend/kernelCI_app/queries/hardware.py
```diff
@@ -370,16 +364,10 @@ def get_hardware_listing_data_from_status_table(
...
-                hardware_status.checkout_id = latest_checkout.checkout_id
-            AND
-                latest_checkout.start_time >= %(start_date)s
-            AND
-                latest_checkout.start_time <= %(end_date)s
         WHERE
             hardware_status.test_origin = %(origin)s
+            AND hardware_status.start_time >= %(start_date)s
+            AND hardware_status.start_time <= %(end_date)s
         GROUP BY
```
mentonin, line 397: I think this will overcount tests (getting all builds tests in the time range, including repeats/redos).

mentonin (changes requested):
I think we might need to iterate a little more on what should be counted as "current" for a given hardware platform or tree.

Deciding which builds and tests to show for a given platform should be the "easier" part, but deciding which hardwares, builds and tests to show for a given tree is complex

<details>
<summary>Example complications</summary>

Consider the following commit graph, where commits are marked according to the key:

- #: checkout with tests on platform X
- *: checkout with tests on platform Y
- $: checkout with tests on platform Z
- branch names in parentheses

```
  G  (dog)
 /
A#*$ ---------B#*  (cat)
 \
  C#*$ --- D* --- E  (cow)
            \
             F#$ --- H#  (elk)
```

Assuming every checkout has at least one test, `latest_checkouts` maps: `cat -> B`, `elk -> F` and `cow -> D` (no checkout on `E` yet). `dog` points to nothing (checkout on `A` declared `cat` for its branch)

### Old implementation:
Only tests at `latest_checkouts`. Tests shown:

*By platform:*
- X: `cat@B`, `elk@H`
- Y: `cat@B`, `cow@D`
- Z: none

*By tree:*
- `cat`: `X@B`, `Y@B`
- `cow`: `Y@D`
- `elk`: `X@H`

### New implementation:
Gets all tests in range. Tests shown:

*By platform:*
- X: `cat@(A,B)`, `cow@C`, `elk@(F,H)`
- Y: `cat@(A,B)`, `cow@(C,D)`, `elk@F
- Z: `cat@A`, `cow@C`, `elk@F`

*By tree:*
- `cat`: `X@(A,B)`, `Y@(A,B)`, `Z@A`
- `cow`: `X@C`, `Y@(C,D)`, `Z@F`
- `elk`: `X@(F,H)`, `Z@F`

### My expectation:
Gets latest checkout per (branch, platform) pair:

*By platform:*
- X in `cat@B`, `cow@C`, `elk@H`
- Y in `cat@B`, `cow@D`
- Z in `cat@A`, `cow@C`, `elk@F`

*By tree:*
- `cat`: `(X,Y)@B`, `Z@A`
- `cow`: `Y@D`, `(X,Z)@C`
- `hawk`: `X@H`, `Y@F`

Optionally, also fall back to checkouts at git topological (first) parents (would require major changes):
- `dog` would show `(X,Y,Z)@A`
- `hawk` would show `Z@D`

</details>

Some questions:

1. When looking at/aggregating a platform, which builds/tests should be included?
    a) in the latest checked out commit per tree (current)
    b) in any commit (this PR)
    d) in the latest checked out commit with at least one build in that platform
    c) in the latest checked out commit per {tree, identifier} (my suggestion)

2. When looking at/aggregating a tree, which builds/tests should be included?
    a) in the latest checked out commit (current)
    b) in the latest commit and same tree
    c) latest by identifier
    d) latest by platform by identifier

backend/kernelCI_app/queries/hardware.py
alanpeixinho: I dont have many problems with using `connection` per se, it is just that, from previous PRs, we are preferring to use connections and be more explicit.
But, at the same time, I was trying to not go beyond scope on this task and touch many lines.
But I would argue against creating an alias for connection here. The django `connection` is already an alias for the default connection, so it would defeat the purpose of being explicit.
https://github.com/django/django/blob/main/django/db/__init__.py

backend/kernelCI_app/queries/hardware.py
alanpeixinho: You have a good point here.
We might need to go for the **latest checkout where that hardware is present**. What you think?
I think this is another point in favor of #1947.

alanpeixinho: I have changed so we get the latest checkout on that hardware, without overcounting tests made in multiple checkouts within the time window.

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
```diff
@@ -54,88 +55,86 @@ def _run_action(self, options):
...
+        self.stdout.write(
+            f"Found {stale_hardware_count} HardwareStatus entries "
+            f"and {stale_processed_count} ProcessedListingItems entries "
+            f"older than {settings.HARDWARE_STATUS_RETENTION_DAYS} days."
+        )
+
+        total_hardware_deleted = 0
+        total_processed_deleted = 0
+        with transaction.atomic():
             while True:
```
felipebergamin: this is actually killing the batch strategy, right? since it will hold the same transaction (and table locks) until the loop finishes.

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
```diff
@@ -54,88 +55,86 @@ def _run_action(self, options):
...
-                    checkout_id__in=valid_checkout_ids
-                )
-            ).values_list("listing_item_key", flat=True)
+        stale_hardware = HardwareStatus.objects.filter(start_time__lt=cutoff)
+        recent_checkout_ids = Checkouts.objects.filter(start_time__gte=cutoff).values(
+            "id"
+        )
+        stale_processed = ProcessedListingItems.objects.exclude(
+            checkout_id__in=recent_checkout_ids
+        )
```
felipebergamin: I'm concerned about this. It is removing everything that is not "new". So we are removing start_dates and checkout_ids with `null` values?

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
alanpeixinho: This was the strategy in the original code, but I agree with you here. It is safer to release the locks after every batch, instead of accumulating.
Changed to a per batch lock.

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
alanpeixinho: For what I see `checkout_id` on `ProcessedListingItems` is  a non nullable field.
The start_time from checkouts is nullable in the model. However, checking the production database, we do not have any instance of null start_times on checkout.
So I believe we are safe regarding this. (But we might as well enforce non nullable on `start_time` for `checkouts`)

backend/kernelCI_app/management/commands/delete_unused_hardware_status.py
felipebergamin, line 1: I'm wondering, isn't ProcessedListingItems reused by HardwareStatus, TreeListing, and the tree tests rollup?

Since we are only removing HardwareStatus, we might run into issues if the tests get re-ingested, right? I don't actually know if that can happen, though.

alanpeixinho: You are correct. That is also what main is doing today.
In this PR we are actually relaxing the deletion, to only delete after many days. So it might be a little bit safer, but still a potential problem as you pointed.
This is an existing problem with current implementation.
To be fair, I would rather advocate to not use a ledger whatsoever, as it can introduce many edge cases, as we are implementing in #2037.

But for the current implementation, do you have any safer suggestion?

felipebergamin: > But for the current implementation, do you have any safer suggestion?

I'm not sure, I agree that would be better to not use the ledger, but do we need to prune the ledger table together with hardware_status? Can't we remove only the hardware_status and keep the ledger row in this command?

alanpeixinho: Good point, it might be safer to not remove from the processed_listing_items ledger here.
I am dropping the prune from the unused_hardware_status command, and we can do two things later.
1. Add an option to prune old processed_listing_items in the global prune_db (#1969) command.
2. Move towards deprecating the ledger completely.

felipebergamin (approved):

PR #1991, opened 2026-07-14
fix/repo url malformed

alanpeixinho:
### What it is

Defensive check for malformed url on git repository url submissions.
Altough it is not an observed scenario, we are now making sure a malformed input cannot break the ingester.

### How to test

1. Start a local database
2. In the submissions folder include invalid git_repository_url on the json submissions.
3. Make sure it cannot cause a crash.

### DIFF
--- a/backend/kernelCI_app/tests/unitTests/commands/treeproof_test.py
+++ b/backend/kernelCI_app/tests/unitTests/commands/treeproof_test.py
@@ -1,0 +1,85 @@
+from unittest import TestCase
+from unittest.mock import MagicMock, patch
+
+from kernelCI_app.management.commands.treeproof import Command
+
+
+class TestDefineTreeName(TestCase):
+    def setUp(self):
+        self.command = Command()
+
+    def test_https_org_repo(self):
+        url = "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
+        self.assertEqual(self.command._define_tree_name(url), "torvalds-linux")
+
+    def test_trailing_slash(self):
+        url = "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/"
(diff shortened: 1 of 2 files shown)


felipebergamin (approved):

backend/kernelCI_app/tests/unitTests/commands/treeproof_test.py
```diff
@@ -0,0 +1,85 @@
...
+        self.assertEqual(self.command._define_tree_name(url), "linux")
+
+    def test_unmatched_returns_none(self):
+        for url in (
+            "",
+            "   ",
+            "https://example.com/",
+            "https://example.com",
+            "linux.git",
+            "git@github.com:org/repo.git",
```
mentonin, line 46: The previous code accepted URLs without scheme, are we ok with changing behavior?

mentonin (approved):

backend/kernelCI_app/tests/unitTests/commands/treeproof_test.py
alanpeixinho: Good catch.

But it seems that the scheme is enforced on kcidb.
https://github.com/kernelci/kcidb-io/blob/main/kcidb_io/schema/v05_03.py
And the available data also seems to be composed solely of urls with scheme.

Do you think it might be safer to keep previous behavior more strictly here?

mentonin: No, I am okay with gracefully dropping data that does not adhere to kcidb-schema

alanpeixinho: Nice. But I will add some comment in the commit message regardless, just to make clear we are changing existing behavior here.

PR #1992, opened 2026-07-15
chore: add checks for schema changes

mentonin:
## Summary
This PR adds a pre-commit hook to check that `backend/schema.yml` is up-to-date with `python manage.py spectacular`.

It also adds all pre-commit checks to the CI workflow, replacing duplicated linting/format specifications.

## Related Issues
- Closes #1910

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,23 @@ env:
   CI: true
 
 jobs:
+  pre-commit-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: 3.13
+      - name: Cache Pre-Commit Hooks
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: ~/.cache/pre-commit
+          key: pre-commit-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('.pre-commit-config.yaml') }}
+      - name: Setup pre-commit
+        run: pip install pre-commit
+      - name: Run pre-commit hooks
+        run: pre-commit run --all-files --show-diff-on-failure
+
   setup-node:
     if: github.event.pull_request.draft != true
     runs-on: ubuntu-latest
@@ -42,34 +59,6 @@ jobs:
       - name: Installing dependencies with pnpm
         if: steps.cache-node_modules.outputs.cache-hit != 'true'
         run: pnpm install --frozen-lockfile
-        working-directory: ./dashboard
-
-  lint-js:
-    if: github.event.pull_request.draft != true
-    needs: setup-node
-    runs-on: ubuntu-latest
-    timeout-minutes: 5
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Fetch node_modules
-        uses: actions/cache@v4
-        id: fetch-node_modules
-        with:
-          key: node_modules-${{ runner.os }}-${{ hashFiles('./dashboard/pnpm-lock.yaml') }}
-          path: ./dashboard/node_modules
-
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
-        with:
-          version: 9.15.2
-
-      - uses: actions/setup-node@v4
-        with:
-          node-version-file: './dashboard/.nvmrc'
-
-      - name: Run eslint
-        run: pnpm lint-staged
         working-directory: ./dashboard
 
   build-front:
@@ -104,7 +93,7 @@ jobs:
         run: pnpm build
         working-directory: ./dashboard
 
-  lint-and-unit-test-django:
+  unit-test-django:
     if: github.event.pull_request.draft != true
     runs-on: ubuntu-latest
     timeout-minutes: 3
@@ -113,14 +102,6 @@ jobs:
 
       - name: Setup Backend
         uses: ./.github/actions/setup-backend
-
-      - name: Check Lint
-        run: poetry run ruff check .
-        working-directory: ./backend
-
-      - name: Check Format
-        run: poetry run ruff format --check .
-        working-directory: ./backend
 
       - name: Run unit tests with coverage
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -18,6 +18,23 @@ env:
   CI: true
 
 jobs:
+  pre-commit-checks:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
```
alanpeixinho: wont we need to setup node (and pnpm) to be able to run the frontend linters?

.github/workflows/ci.yaml
```diff
@@ -44,34 +61,6 @@ jobs:
...
-
-      - name: Fetch node_modules
-        uses: actions/cache@v4
-        id: fetch-node_modules
-        with:
-          key: node_modules-${{ runner.os }}-${{ hashFiles('./dashboard/pnpm-lock.yaml') }}
-          path: ./dashboard/node_modules
-
-      - name: Install pnpm
-        uses: pnpm/action-setup@v4
```
alanpeixinho, line 63: might be needed on pre-commit run?

.github/workflows/ci.yaml
alanpeixinho: we are using the old task name here

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,23 @@ env:
   CI: true
 
 jobs:
+  pre-commit-checks:
```
alanpeixinho: should we filter to only run on ready PRs?
`if: github.event.pull_request.draft != true`

.github/workflows/ci.yaml
alanpeixinho: might be better to reuse setup-backend
`uses: ./.github/actions/setup-backend`

.github/workflows/ci.yaml
alanpeixinho: might be worth to include timeout just in case

.github/workflows/ci.yaml
mentonin: Nice catch. Pre-commit is able to setup isolated environments, including for node, but we are not using that for the frontend checks

mentonin: We don't need poetry and dependencies, as pre-commit should use its own environment for running tests

.github/workflows/ci.yaml
mentonin: I deliberately chose to not include that filter because pre-commit checks are supposed to be cheap/fast to run. Should I filter anyway?

alanpeixinho: No. Fair point.

alanpeixinho (approved):
LGTM

.pre-commit-config.yaml
```diff
@@ -74,3 +74,25 @@ repos:
...
+          if new_schema != old_schema:
+              print("Schema is not up-to-date. Run \"python manage.py spectacular --file schema.yml\" to regenerate it")
+              sys.exit(1)
+          '
+        files: 'backend/(.*\.py|schema.yml)'
+        pass_filenames: false
+        language: python
+        language_version: '3.12'
+        additional_dependencies: [poetry]
+        stages: [pre-commit]
```
tales-aparecida, line 98: Any reason to use python+subprocess instead of sh?

```suggestion
      - id: api-schema
        name: API schema is up-to-date
        entry: |
          bash -c '
          cd backend || exit 1
          poetry install --quiet
          NEW_SCHEMA=$(poetry run python manage.py spectacular)
          OLD_SCHEMA=$(cat schema.yml 2>/dev/null)
          
          if [ "$NEW_SCHEMA" != "$OLD_SCHEMA" ]; then
            echo "Schema is not up-to-date. Run \"python manage.py spectacular --file schema.yml\" to regenerate it"
            exit 1
          fi
          '
        files: '^backend/(.*\.py|schema\.yml)$'
        pass_filenames: false
        language: system
        stages: [pre-commit]
```

-- Assisted-by: Gemini

tales-aparecida: It also warned about `poetry install --quiet` causing commits to become slow, but I doubt it's a problem

tales-aparecida: Finally, it suggested using `--validate --fail-on-warn` instead of diffing the schema

tales-aparecida (approved):
we can improve it later if this new  precommit hook is really slow

.pre-commit-config.yaml
mentonin: > Any reason to use python+subprocess instead of sh?

using `language: python` uses an isolated, cached environment. This makes it harder (not impossible) for pre-commit checks to affect or be affected by each other or the runner environment.

<details>
<summary>If you want to check it:</summary>

Run `pre-commit run -v `test-system`; pre-commit run -v test-python` with these hooks:

As a nice bonus, a developer should only need pre-commit installed to run pre-commit checks that do not use `language: system|unsupported|unsupported-script`

```yaml
# .pre-commit-config.yaml
repos:
  - repo: local
    hooks:
      - id: test-system
        name: hook with system
        entry: sh -c 'cd backend && type python poetry && poetry env info'
        language: system
        pass_filenames: false

      - id: test-python
        name: hook with python
        entry: sh -c 'cd backend && type python poetry'
        language: python
        additional_dependencies: [poetry]
        pass_filenames: false
```
</details>

Using `sh` (or better `/bin/sh`) could be done (`bash` might not be available), and may be an improvement in readability for the future

mentonin: > `poetry install --quiet` causing commits to become slow

On a development machine, it is cached, so should only take a little while on the first run.

On CI, this PR also caches it, and we would need to `poetry-install` in the job in any case; the alternative would be installing only the subset of packages we need to call `python manage.py spectacular`, but that would add maintenance headache.

mentonin: > using `--validate --fail-on-warn` instead of diffing the schema

`python manage.py --validate` does something else entirely (and does not update the schema). It might be useful to include it later, though. If we use it, `--fail-on-warn` is good, but we currently have 8 warnings, so we would need to fix them first

mentonin: Sorry for taking a while to answer

tales-aparecida: https://drf-spectacular.readthedocs.io/en/latest/client_generation.html

> For generating clients with CI, we highly recommend using ./manage.py spectacular --file schema.yaml --validate --fail-on-warn to catch potential problems early on.

mentonin: I investigated a little further, and I think I found the reason `--validate` was not working for me: our schema currently does not pass validation. I'll open an issue

PR #1995, opened 2026-07-16
feat/issues list cache

### LINKED ISSUE #1615: Add cache to issue listing
The issue listing page doesn't seem to be changing too much in the near future, but its query doesn't have any kind of cache, so it should have.

alanpeixinho:
### What it is

  * Add redis cache to top level issues listing.

### How to test

  * Start backend locally with `CACHE_TIMEOUT>0` setting
  * call /api/issue endpoint
  * Make sure that subsequent calls (during the cache window time) fetch values from redis cache.

Closes #1615

### DIFF
--- a/backend/kernelCI_app/tests/unitTests/queries/issues_queries_test.py
+++ b/backend/kernelCI_app/tests/unitTests/queries/issues_queries_test.py
@@ -86,7 +86,27 @@ class TestGetIssueListingData:
     @patch("kernelCI_app.queries.issues.set_query_cache")
     @patch("kernelCI_app.queries.issues.dict_fetchall")
     @patch("kernelCI_app.queries.issues.connection")
-    def test_get_issue_listing_data_success(
+    def test_get_issue_listing_data_cache_hit(
+        self, mock_connection, mock_dict_fetchall, mock_set_cache, mock_get_cache
+    ):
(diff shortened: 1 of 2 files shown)


mentonin (approved):

felipebergamin (approved):

backend/kernelCI_app/tests/unitTests/queries/issues_queries_test.py
```diff
@@ -95,6 +100,7 @@ def test_get_issue_listing_data_success(self, mock_connection, mock_dict_fetchal
         )
 
         assert result == expected_result
+        mock_set_cache.assert_called_once()
```
felipebergamin: I think this is a nit, but maybe we could add a test to a cache hit.

also, I'd change the test case. `test_get_issue_listing_data_success` does not make it clear how it expected cache to behave.

Maybe we could have `test_get_issue_listing_data_cache_hit` and `test_get_issue_listing_data_cache_miss`

alanpeixinho: Done — split into `test_get_issue_listing_data_cache_hit` and `test_get_issue_listing_data_cache_miss`.

PR #1996, opened 2026-07-16
fix/crash test hardware deselected tree

### LINKED ISSUE #1482: Opening the log sheet of a desselected boot crashes hardware details
In hardware details, if there are two trees A and B selected, **both with boots**, deselect tree A and quickly open the log sheet of a build from A; when the future data loads, hardware details will crash (because the build that you selected is not an option anymore).

This doesn't seem to happen on builds or tests tab, and also doesn't happen if A has boots but B does not. If you wait for the loading of tree B, you obviously won't be able to select boots from tree A either. It's a very specific bug.

alanpeixinho:
### What it is

Fixes the crash where opening the log sheet of a boot whose tree was just deselected broke hardware details.

The log-sheet tables (BootsTable, BuildsTable, IndividualTestsTable) tracked the open log by row index. When a tree was deselected, the reloaded data dropped/reordered rows, so the stored index pointed past the list (sortedItems[currentLog].id → crash) or at the wrong item.

The tables now track the open log by the item's stable id and derive the row index from the live list each render. When the reloaded data no longer contains that id, the sheet closes on its own. This removes the fragile positional indexing entirely (crash impossible by construction) and also fixes the sheet reopening by itself when the tree was reselected.

### How to test

  1. Open a hardware details page and select two trees, A and B, both with boots.
  2. Go to the Boots tab, deselect tree A, and quickly open the log sheet of a boot from A before B finishes loading.
  3. Expected: no crash. Once the filtered data reloads, the sheet closes automatically.
  4. Reselect tree A — the log sheet should stay closed (it must only open on an explicit click).
  5. Repeat on the Builds and Tests tabs to confirm consistent behavior.

Closes #1482

### DIFF
--- a/dashboard/src/components/BootsTable/BootsTable.tsx
+++ b/dashboard/src/components/BootsTable/BootsTable.tsx
@@ -305,10 +305,29 @@ export function BootsTable({
     [modelRows],
   );
 
-  const [currentLog, setLog] = useState<number | undefined>(undefined);
+  const [currentLogId, setLog] = useState<string | undefined>(undefined);
+
+  const currentLog = useMemo(() => {
+    const index = sortedItems.findIndex(item => item.id === currentLogId);
+    return index === -1 ? undefined : index;
+  }, [sortedItems, currentLogId]);
+
+  const activeLogId = currentLog !== undefined ? currentLogId ?? '' : '';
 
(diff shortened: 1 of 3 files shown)


felipebergamin (approved):
looking good. Tested locally, the fix works

PR #1998, opened 2026-07-17
feat/vcs info

### LINKED ISSUE #1611: Add git tag and commit to sidebar
We have had problems where the staging is not up to date with the current main commit, and since some of the changes are backend-only, it may be hard to check if they are really out-of-sync. Adding the git tag (release) and hash would make it easier to check for that.

alanpeixinho:
### What it is

Shows the deployed version (`git describe --tags --always`, uses tag if present, otherwise uses short commit) at the bottom of the sidebar, so we can tell which build is running even when drift is backend-only.

#### Dev
* Start frontend with `pnpm dev`
* Sidebar version should match `git describe --tags --always --dirty`.
* Clicking on the version should redirect you to the repository (in a specific version, or to the root if not a published version).

#### Docker
* Use `DASHBOARD_VERSION=$(git describe --tags --always) docker compose build dashboard`.
* Sidebar version should match DASHBOARD_VERSION.
* Clicking on the version should redirect you to the repository (in a specific version, or to the root if not a published version).

Closes #1611

### DIFF
--- a/.github/workflows/deploy-containers.yaml
+++ b/.github/workflows/deploy-containers.yaml
@@ -57,10 +57,16 @@ jobs:
       digest: ${{ steps.build.outputs.digest }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
       - name: Compute image prefix
         id: meta
         run: echo "prefix=${REGISTRY}/${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
+
+      - name: Compute dashboard version
+        id: dashboard_version
+        run: echo "dashboard_version=$(git describe --tags --always)" >> "$GITHUB_OUTPUT"
 
       - name: Log in to GHCR
         uses: docker/login-action@v3
@@ -79,6 +85,8 @@ jobs:
           context: .
           file: ./dashboard/Dockerfile
           push: true
+          build-args: |
+            DASHBOARD_VERSION=${{ steps.dashboard_version.outputs.dashboard_version }}
           tags: |
             ${{ steps.meta.outputs.prefix }}/dashboard-frontend:latest
             ${{ steps.meta.outputs.prefix }}/dashboard-frontend:${{ github.sha }}
--- a/.github/workflows/deploy-staging.yaml
+++ b/.github/workflows/deploy-staging.yaml
@@ -9,11 +9,16 @@ jobs:
     outputs:
       has_new_migrations: ${{ steps.check.outputs.has_new_migrations }}
       migration_list: ${{ steps.check.outputs.migration_list }}
+      dashboard_version: ${{ steps.dashboard_version.outputs.dashboard_version }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
+      - name: Compute dashboard version
+        id: dashboard_version
+        run: echo "dashboard_version=$(git describe --tags --always)" >> "$GITHUB_OUTPUT"
 
       - name: Check for new Django migrations
         id: check
@@ -68,6 +73,7 @@ jobs:
             cp ~/.env-staging dashboard-staging/.env &&
             cd dashboard-staging &&
             git checkout ${GITHUB_SHA} &&
+            export DASHBOARD_VERSION=${DASHBOARD_VERSION} &&
             docker compose pull --ignore-buildable --policy=always &&
             docker compose build --no-cache &&
             docker compose up -d --remove-orphans
@@ -76,6 +82,7 @@ jobs:
           SSH_USER: ${{ secrets.STAGING_USER }}
           SSH_HOST: ${{ secrets.STAGING_HOST }}
           SSH_KEY: ${{ secrets.STAGING_KEY }}
+          DASHBOARD_VERSION: ${{ needs.check-migrations.outputs.dashboard_version }}
 
   cleanup:
     needs: deploy-staging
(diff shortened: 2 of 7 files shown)


.github/workflows/deploy-containers.yaml
```diff
@@ -79,6 +85,8 @@ jobs:
           context: .
           file: ./dashboard/Dockerfile
           push: true
+          build-args: |
+            VITE_APP_VERSION=${{ steps.version.outputs.value }}
```
felipebergamin: just a nit: your commit message says VITE_VERSION but actually is VITE_APP_VERSION

.github/workflows/deploy-containers.yaml
```diff
@@ -57,11 +57,17 @@ jobs:
...
+        with:
+          fetch-depth: 0
 
       - name: Compute image prefix
         id: meta
         run: echo "prefix=${REGISTRY}/${GITHUB_REPOSITORY,,}" >> "$GITHUB_OUTPUT"
 
+      - name: Compute app version
+        id: version
+        run: echo "value=$(git describe --tags --always)" >> "$GITHUB_OUTPUT"
```
mentonin: Can we use a descriptive name here?

docker-compose.yml
```diff
@@ -114,6 +114,8 @@ services:
         build:
             context: .
             dockerfile: ./dashboard/Dockerfile
+            args:
+                VITE_APP_VERSION: ${APP_VERSION:-}
```
mentonin: Why mix APP_VERSION and VITE_APP_VERSION?

alanpeixinho: The `VITE_` prefix is required by VITE.
But, on our own variables, I would rather keep them agnostic (if we ever change vite).

.github/workflows/deploy-containers.yaml
alanpeixinho: Absolutelly. Do you have any suggestion?

.github/workflows/deploy-containers.yaml
alanpeixinho: Fixed

docker-compose.yml
mentonin: In that case, I would prefer `FRONTEND_VERSION` or `DASHBOARD_VERSION`

.github/workflows/deploy-containers.yaml
mentonin: > Do you have any suggestion?

`version` or `repo_version` or `app_version` or `APP_VERSION` (or equivalent if you decide to change `APP_VERSION`)

docker-compose.yml
mentonin: Also, use `APP_VERSION` as the container ARG/ENV, and set `VITE_APP_VERSION` inside the container, so we only interact with "our" variable when building (e.g. on the deploy-containers workflow)

felipebergamin (approved):

docker-compose.yml
alanpeixinho: Done

alanpeixinho: > The `VITE_` prefix is required by VITE. But, on our own variables, I would rather keep them agnostic (if we ever change vite).

some correction here: we can change the naming here, so I went with DASHBOARD_VERSION for the app, we only need to keep the VITE_ prefix for the inner env variable.

.github/workflows/deploy-containers.yaml
alanpeixinho: Done

mentonin (changes requested):
LGTM, just needs to rebase to solve conflicts

docker-compose.yml
alanpeixinho: Also changed to VITE_DASHBOARD_VERSION to keep consistent.

.github/workflows/deploy-staging.yaml
```diff
@@ -68,36 +67,17 @@ jobs:
...
             docker compose build --no-cache &&
-            docker compose up -d --remove-orphans
+            docker compose up -d
           "
         env:
           SSH_USER: ${{ secrets.STAGING_USER }}
           SSH_HOST: ${{ secrets.STAGING_HOST }}
           SSH_KEY: ${{ secrets.STAGING_KEY }}
 
-  cleanup:
```
MarceloRobert, line 80: Why remove the cleanup step? is the `prune` unnecessary? Either way I don't understand how this is related to the PR subject

alanpeixinho: It was accidental, just a rebase against outdated version. I am updating it now.

mentonin (approved):

MarceloRobert (approved):
LGTM

DEPLOYMENT.md
```diff
@@ -205,6 +205,28 @@ docker compose -f docker-compose-next.yml pull
 docker compose -f docker-compose-next.yml up -d
 ```
 
+### Tagging a release
+
+Every production deployment must be preceded by a release tag.
+The dashboard displays its version (`git describe --tags`) at the bottom of the
+side menu, so an untagged deployment shows only a commit hash, making it hard to
```
mentonin: technically, it will show `release/<old release>-N-g<sha>`, not only a commit sha

mentonin (approved):
Looks good, only a few nits:
Commit message still using VITE_APP_VERSION and VITE_VERSION

DEPLOYMENT.md
alanpeixinho: Fair point. Fixed.

PR #2016, opened 2026-07-21
proxy: add JavaScript cookie challenge for dumb bots

nuclearcat:
Production observations show many unsophisticated bots stressing the dashboard with random requests. Add a lightweight JavaScript and cookie hurdle before serving frontend or API requests from Safari, Android, and Chromium-style user agents.

Cross-check the claimed browser platform/vendor, reject WebDriver, require the exact verification cookie value, and safely return browsers to the original URI. Preserve local HTTP support while marking the cookie Secure on browser-facing HTTPS connections.

Legitimate programmatic API clients using their normal python-requests, Wget, or curl user agents bypass the browser gate. Clients that claim to be browsers must complete the same challenge before reaching /api.

This is intentionally a speed bump for dumb bots rather than an authentication boundary

### DIFF
--- a/proxy/etc/nginx/templates/default.conf.template
+++ b/proxy/etc/nginx/templates/default.conf.template
@@ -1,12 +1,118 @@
+# This is deliberately a lightweight challenge for unsophisticated bots, not
+# an authentication mechanism. Browser UAs commonly containing "Safari"
+# include Safari itself and Chromium-based browsers; Android browsers are
+# included separately because not all of them advertise Safari.
+map $http_user_agent $dashboard_browser_ua {
+    default 0;
+    ~*(Safari|Android) 1;
+}
+
+map $cookie_kernelci_dashboard_verified $dashboard_cookie_verified {
+    default 0;
+    "1" 1;
+}
+
+map "$dashboard_browser_ua:$dashboard_cookie_verified" $dashboard_cookie_gate {
+    default 0;
+    "1:0" 1;
+}
+
 server {
     location /api {
+        if ($dashboard_cookie_gate) {
+            add_header Cache-Control "no-store" always;
+            return 302 /__dashboard_cookie_check?return=$request_uri;
+        }
         proxy_pass ${PROXY_TARGET};
         proxy_connect_timeout 240s;
         proxy_read_timeout 240s;
         proxy_send_timeout 240s;
         send_timeout 240s;
     }
+
+    # Keep the challenge outside `location /` so it cannot redirect to itself.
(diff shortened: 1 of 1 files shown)


proxy/etc/nginx/templates/default.conf.template
```diff
@@ -1,12 +1,118 @@
...
+    }
+
+    var marker = "?return=";
+    var target = location.search.slice(marker.length);
+    if (location.search.slice(0, marker.length) !== marker ||
+        target.charAt(0) !== "/" || target.charAt(1) === "/" ||
+        target.charAt(1) === "\\") {
+        target = "/";
+    }
+    location.replace(target);
```
tales-aparecida, line 103: Gemini suggests to replace string parsing for `?return=` in JavaScript with native `URL()` handling to guarantee absolute path safety.

```suggestion
    var params = new URLSearchParams(location.search);
    var target = params.get("return") || "/";

    // Enforce a strict relative path to prevent open redirects (e.g., //evil.com, /\evil.com)
    if (!target.startsWith("/") || target.startsWith("//") || target.startsWith("/\\")) {
        target = "/";
    }

    location.replace(target);
```

alanpeixinho: In the solution proposed by Gemini, we are possibly dropping extra query parameters other than `return`.
I think its fine to check via startsWith (cleaner), but using URLSearchParams might bring problematic. Unless we only have `return` as a single query parameter.

proxy/etc/nginx/templates/default.conf.template
```diff
@@ -1,12 +1,118 @@
...
+(function () {
+    var ua = navigator.userAgent;
+    var uaData = navigator.userAgentData;
+    var claimsAndroid = /Android/i.test(ua);
+    var claimsAppleMobile = /(iPhone|iPad|iPod)/i.test(ua);
+    var claimsChromium = /(Chrome|Chromium|CriOS|Edg|OPR)/i.test(ua);
+    var claimsSafari = /Safari/i.test(ua) && !claimsChromium;
+    var platformLooksValid = false;
+
+    if (claimsAndroid) {
```
alanpeixinho, line 61: Cant we endup loosing some legitimate accesses here?
Mostly, people who use hardened browsers or less conventional ones?

nuclearcat: Only those who dont support cookie at all (which are integral part of browsers, even console based), but i am afraid it will happen with Anubis as well.

PR #2017, opened 2026-07-21
proxy: add Anubis bot protection

nuclearcat:
Route public dashboard traffic through an Anubis sidecar before it reaches the existing NGINX API and static-file handlers. Pin the stable v1.25.0 image and wait for its health check before starting the proxy.\n\nImport Anubis's default policy after an explicit allow rule for curl, wget, python-requests, and kci-dev user agents so documented API clients bypass browser challenges. Document the deployment and secure-cookie override.

### DIFF
--- a/docker-compose-next.yml
+++ b/docker-compose-next.yml
@@ -68,12 +68,36 @@ services:
         volumes:
             - static-data:/data/static
 
+    anubis:
+        image: ${ANUBIS_IMAGE:-ghcr.io/techarohq/anubis:v1.25.0}
+        restart: always
+        networks:
+            - public
+        environment:
+            BIND: ":8923"
+            METRICS_BIND: ":9090"
+            TARGET: "http://proxy:8080"
+            POLICY_FNAME: "/data/cfg/botPolicy.yaml"
+            COOKIE_SECURE: ${ANUBIS_COOKIE_SECURE:-true}
+        volumes:
+            - ./anubis/botPolicy.yaml:/data/cfg/botPolicy.yaml:ro
+        healthcheck:
+            test: ["CMD", "anubis", "--healthcheck"]
+            interval: 5s
+            timeout: 30s
+            retries: 5
+            start_period: 500ms
+
     proxy:
         image: ${IMAGE_REGISTRY:-ghcr.io}/${IMAGE_OWNER:-kernelci}/${IMAGE_REPOSITORY:-dashboard}/dashboard-proxy:${IMAGE_TAG:-latest}
         restart: always
         depends_on:
-            - backend
-            - dashboard
+            backend:
+                condition: service_started
+            dashboard:
+                condition: service_started
+            anubis:
+                condition: service_healthy
         networks:
             - public
         volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -118,13 +118,37 @@ services:
         volumes:
             - static-data:/data/static
 
+    anubis:
+        image: ${ANUBIS_IMAGE:-ghcr.io/techarohq/anubis:v1.25.0}
+        restart: always
+        networks:
+            - public
+        environment:
+            BIND: ":8923"
+            METRICS_BIND: ":9090"
+            TARGET: "http://proxy:8080"
+            POLICY_FNAME: "/data/cfg/botPolicy.yaml"
+            COOKIE_SECURE: ${ANUBIS_COOKIE_SECURE:-false}
+        volumes:
+            - ./anubis/botPolicy.yaml:/data/cfg/botPolicy.yaml:ro
+        healthcheck:
+            test: ["CMD", "anubis", "--healthcheck"]
+            interval: 5s
+            timeout: 30s
+            retries: 5
+            start_period: 500ms
+
     proxy:
         build: ./proxy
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -118,13 +118,37 @@ services:
...
+        environment:
+            BIND: ":8923"
+            METRICS_BIND: ":9090"
+            TARGET: "http://proxy:8080"
+            POLICY_FNAME: "/data/cfg/botPolicy.yaml"
+            COOKIE_SECURE: ${ANUBIS_COOKIE_SECURE:-false}
+        volumes:
+            - ./anubis/botPolicy.yaml:/data/cfg/botPolicy.yaml:ro
+        healthcheck:
+            test: ["CMD", "anubis", "--healthcheck"]
```
bhcopeland, line 135: Anubis does give you a nice metrics URL too, which we can wire into grafana in a later date (see how many bots/honeytraps it catches)

docker-compose-next.yml
```diff
@@ -68,12 +68,36 @@ services:
...
             - static-data:/data/static
 
+    anubis:
+        image: ${ANUBIS_IMAGE:-ghcr.io/techarohq/anubis:v1.25.0}
+        restart: always
+        networks:
+            - public
+        environment:
+            BIND: ":8923"
+            METRICS_BIND: ":9090"
```
mentonin, line 78: Do we need to set these to their default values?

docker-compose-next.yml
```diff
@@ -68,12 +68,36 @@ services:
         volumes:
             - static-data:/data/static
 
+    anubis:
+        image: ${ANUBIS_IMAGE:-ghcr.io/techarohq/anubis:v1.25.0}
+        restart: always
+        networks:
+            - public
+        environment:
```
mentonin, line 76: The [docs](https://anubis.techaro.lol/docs/admin/installation) make me think we must set `COOKIE_DOMAIN` (or `COOKIE_DYNAMIC_DOMAIN`).

docker-compose-next.yml
mentonin, line 1: Since we are using Redis already, would it make sense to set it as our storage backend?

docker-compose-next.yml
```diff
@@ -68,12 +68,36 @@ services:
...
+            BIND: ":8923"
+            METRICS_BIND: ":9090"
+            TARGET: "http://proxy:8080"
+            POLICY_FNAME: "/data/cfg/botPolicy.yaml"
+            COOKIE_SECURE: ${ANUBIS_COOKIE_SECURE:-true}
```
mentonin, line 81: `COOKIE_SAME_SITE` and `COOKIE_HTTP_ONLY` also seem relevant here

proxy/etc/nginx/templates/default.conf.template
```diff
@@ -1,4 +1,34 @@
+upstream anubis {
+    server anubis:8923;
+    keepalive 32;
+}
+
+server {
+    listen 80 default_server;
```
mentonin, line 7: If we are redirecting though port 80, how will `COOKIE_SECURE` work here?

docker-compose-next.yml
nuclearcat: Unset domain is an intentional, working mode: Anubis emits Set-Cookie without a Domain attribute, so the browser creates a host-only cookie for the public hostname. This commit also preserves that hostname with proxy_set_header Host $host (in section for anubis upstream).

docker-compose-next.yml
nuclearcat: Not really, but kept them so we dont look into Anubis documentation each time to guess what are defaults. I can remove them.

docker-compose-next.yml
nuclearcat: Existing Redis no very suitable: no persistent volume, it is on private network (Anubis currently joins only public). Using it would add coupling and network/config changes without durable persistence.

docker-compose-next.yml
nuclearcat: COOKIE_HTTP_ONLY is not supported by pinned v1.25.0, but we can upgrade it (i just pinned version i tested before).
COOKIE_SAME_SITE=Lax makes sense for this deployment, so I’ll set it explicitly.

proxy/etc/nginx/templates/default.conf.template
nuclearcat: From the user’s perspective, the connection is HTTPS because TLS is terminated at NGINX. COOKIE_SECURE is enforced by the user’s browser, so the internal HTTP connection between NGINX and Anubis does not matter.

docker-compose-next.yml
mentonin: I didn't see `COOKIE_HTTP_ONLY` was in pre-release; we can look into it when it is out

alanpeixinho (approved):

felipebergamin (approved):

docker-compose-next.yml
alanpeixinho: @nuclearcat Do we still want to make any changes here (adding COOKIE_SAME_SITE=Lax)? Or are we ready to merge?

nuclearcat: I suggest to try as-is, then we can make adjustments

PR #2019, opened 2026-07-22
feat/tables default ordering

alanpeixinho:
### What it is
  * Hardware sorting no longer clears to an unsorted state, so Platform stays `asc` or `desc`.
  * TableHeader now uses TanStack’s getToggleSortingHandler(), so table sorting options (including `enableSortingRemoval`) are respected everywhere.
  * Removes redundant page-level sorts on hardware and tree listing; trees still use pinned-first ordering when unsorted.

#### Hardware

* Go to /hardware page
* Platform shows asc order; rows using asc order by platform.
* Click Platform
* Platform shows desc order; rows using desc order by platform.
* Click Platform
* Platform shows again asc (not unsorted).
* Click another column (e.g. Builds) → sorts; cycle asc/desc/clear as usual.

#### Trees
* Go to /tree page
* Unsorted keeps pinned trees first (mainline/master, next/master, stable/*);
* Click Tree
* Column sort overrides; clear restores pinned-first.

Closes #2004

### DIFF
--- a/dashboard/src/components/TreeListingPage/treeTableUtils.tsx
+++ b/dashboard/src/components/TreeListingPage/treeTableUtils.tsx
@@ -88,7 +88,7 @@ export const commonTreeTableColumns: ColumnDef<TreeListingItem>[] = [
 export const sortTreesWithPinnedFirst = <T extends TreeListingItem>(
   treeTableRows: T[],
 ): T[] => {
-  return treeTableRows.sort((a, b) => {
+  return [...treeTableRows].sort((a, b) => {
     const aKey = makeTreeIdentifierKey({
       treeName: valueOrEmpty(a.tree_name),
       gitRepositoryBranch: valueOrEmpty(a.git_repository_branch),
--- a/dashboard/src/pages/Hardware/HardwareTable.tsx
+++ b/dashboard/src/pages/Hardware/HardwareTable.tsx
@@ -397,16 +397,14 @@ export function HardwareTable({
   const { listingSize, intervalInDays } = useSearch({ strict: false });
   const navigate = useNavigate({ from: navigateFrom });
 
-  const [sorting, setSorting] = useState<SortingState>([]);
+  const [sorting, setSorting] = useState<SortingState>([
+    { id: 'platform', desc: false },
+  ]);
   const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
   const { pagination, paginationUpdater } = usePaginationState(
     'hardwareListing',
     listingSize,
   );
-
-  const data = useMemo(() => {
-    return treeTableRows;
-  }, [treeTableRows]);
 
   const columns = useMemo(
     () =>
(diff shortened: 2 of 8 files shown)


mentonin (approved):
This PR looks great. I wonder if you can add the changes to our other sortable tables as well?
Builds/Boots/Tests/Tests(grouped) tables under `/tree/mainline/<tree_name>/<git_branch>/<git_commit_hash>`, `/hardware/<platform_id>`, `/build/<build_id>`, `/issue/<issue_id>`,  are the ones I can recall

dashboard/src/components/TreeListingPage/treeTableUtils.tsx
```diff
@@ -88,7 +88,7 @@ export const commonTreeTableColumns: ColumnDef<TreeListingItem>[] = [
 export const sortTreesWithPinnedFirst = <T extends TreeListingItem>(
   treeTableRows: T[],
 ): T[] => {
-  return treeTableRows.sort((a, b) => {
+  return [...treeTableRows].sort((a, b) => {
```
MarceloRobert, line 91: Why do you have to spread it if it is already an array (as per `treeTableRows: T[]` on old line 89)?

dashboard/src/pages/Hardware/HardwareTable.tsx
```diff
@@ -397,26 +397,25 @@ export function HardwareTable({
...
+  ]);
   const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
   const { pagination, paginationUpdater } = usePaginationState(
     'hardwareListing',
     listingSize,
   );
 
-  const data = useMemo(() => {
-    return treeTableRows;
-  }, [treeTableRows]);
```
MarceloRobert, line 409: Yeah... 😅

dashboard/src/components/TreeListingPage/treeTableUtils.tsx
alanpeixinho: Might have been overzealous here. It was mostly because javascript sort in done inplace, and I wanted to avoid side effects here.

felipebergamin: I think you can use [.toSorted()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) if you don't want to change the original array

felipebergamin (approved):

dashboard/src/components/TreeListingPage/treeTableUtils.tsx
alanpeixinho: Nice hint. But unfortunately, for what I checked, toSorted is only available at ES2023. And we are using ES2020.
We could take a look on the impact of bumping ES version, so we can access a wider library.

mentonin (approved):
Approved. We should look into bringing the unremovable sort to IssueTable and HardwareDetailsHeaderTable later

PR #2026, opened 2026-07-27
Fix: tests table height

MarceloRobert:
## Summary
Replaces static height with a max limit on individual tests table, allowing tables with less data to be shorter

## Changes
- Replaces static table height with max-h
- Replaces static px with tailwind's size

## How to test
- Go to a page that has the individual tests table (treeDetails/hardwareDetails tests tab, buildDetails, issueDetails) and has only a few tests.
- Check that the table doesn't have whitespace at the bottom but still allows for scrolling on other pages when it has a lot of tests.
- Compare with the original

## Visual Examples

Before / After:

image

image

Test with staging environment: https://staging.dashboard.kernelci.org/issue/redhat%3Aissue_2504?iv=1774001989&o=redhat

Closes #2025

### DIFF
--- a/dashboard/src/components/TestsTable/TestsTable.tsx
+++ b/dashboard/src/components/TestsTable/TestsTable.tsx
@@ -260,7 +260,7 @@ export function TestsTable({
         onSearchChange={onSearchChange}
         currentPathFilter={currentPathFilter}
       />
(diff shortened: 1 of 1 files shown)


mentonin (approved):

alanpeixinho (approved):
LGTM

PR #2029, opened 2026-07-28
feat: sync table sort state to the URL

mentonin:
## Summary

- Persist sortable table state in the URL via a shared `useSortingState` hook and root `tableSort` search param, so sort order is shareable and bookmarkable.
- Encode a single active table as flat `s=…`; use keyed params (`s|b=…`, `s|t=…`) only on pages with multiple tracked tables (issue details).
- Clear `tableSort` when switching tree/hardware details tabs so inactive tables do not keep stale sort.

## Test plan

- [ ] Sort tree, hardware, and issue listing tables; confirm the URL updates and reload/share restores the same sort
- [ ] Reset to default sort and confirm the `s` / `tableSort` param is omitted
- [ ] On tree and hardware details, sort a tab, switch tabs, and confirm sort resets for the newly active table
- [ ] On issue details, sort builds and tests independently and confirm keyed URL params (`s|b`, `s|t`)
- [ ] Confirm unit tests: `useSortingState.test.ts`, `search.test.ts`

## Related issues

- Closes #2005
- Impacts A/B comparison - #1951 et al.
- Impacts #2006

### DIFF
--- a/dashboard/src/hooks/useSortingState.ts
+++ b/dashboard/src/hooks/useSortingState.ts
@@ -1,0 +1,141 @@
+import { useCallback, useMemo } from 'react';
+
+import type { OnChangeFn, SortingState } from '@tanstack/react-table';
+import { useNavigate, useSearch } from '@tanstack/react-router';
+
+import { TABLE_SORT_UNSORTED, type TableSortSearch } from '@/types/general';
+import { EMPTY_OBJECT } from '@/utils/constants/general';
+
+const EMPTY_SORTING: SortingState = [];
+
+const sortingEquals = (a: SortingState, b: SortingState): boolean => {
+  if (a.length !== b.length) {
+    return false;
+  }
+  return a.every(
+    (sort, index) => sort.id === b[index]?.id && sort.desc === b[index]?.desc,
+  );
+};
+
+export const sortingToSearchValue = (sorting: SortingState): string => {
+  if (sorting.length === 0) {
+    return TABLE_SORT_UNSORTED;
+  }
+  const sort = sorting[0];
+  return sort.desc ? `-${sort.id}` : sort.id;
+};
+
+export const searchValueToSorting = (
+  value: string | undefined,
+  defaultSorting: SortingState,
+): SortingState => {
+  if (value === undefined) {
+    return defaultSorting;
+  }
+  if (value === TABLE_SORT_UNSORTED) {
+    return [];
+  }
+  if (value.startsWith('-')) {
+    return [{ id: value.slice(1), desc: true }];
+  }
+  return [{ id: value, desc: false }];
+};
+
+const toUrlValue = (
+  sorting: SortingState,
+  defaultSorting: SortingState,
+): string | undefined => {
+  if (sortingEquals(sorting, defaultSorting)) {
+    return undefined;
+  }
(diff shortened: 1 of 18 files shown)


dashboard/src/hooks/useSortingState.ts
```diff
@@ -0,0 +1,139 @@
...
+  const sortValue = getTableSortValue(tableSort, sortKey);
+
+  const sorting = useMemo(
+    () => searchValueToSorting(sortValue, defaultSorting),
+    [defaultSorting, sortValue],
+  );
+
+  const sortingUpdater: OnChangeFn<SortingState> = useCallback(
+    updater => {
+      navigate({
```
alanpeixinho, line 114: prefer to use replace instead of push here, or we might end up hurting navigation stack

dashboard/src/pages/Hardware/HardwareTable.tsx
alanpeixinho, line 384: This seems to be happening only on hardware listing. But if we select a sorting strategy, and later click in a row, the sorting parameter seems to be carried in the url.

alanpeixinho (commented):
Pretty nice

dashboard/src/pages/Hardware/HardwareTable.tsx
mentonin: I think the new commit fixes this everywhere now

dashboard/src/hooks/useSortingState.ts
```diff
@@ -0,0 +1,140 @@
...
+};
+
+type UseSortingStateOptions = {
+  defaultSorting?: SortingState;
+  /** Set only when multiple tables on the same page need distinct sort params. */
+  sortKey?: string;
+};
+
+export const useSortingState = (
+  options: UseSortingStateOptions = {},
```
felipebergamin:
```suggestion
const EMPTY_OBJECT = {};

export const useSortingState = (
  options: UseSortingStateOptions = EMPTY_OBJECT,
```

dashboard/src/hooks/useSortingState.ts
```diff
@@ -0,0 +1,140 @@
...
+  const navigate = useNavigate();
+
+  const sortValue = getTableSortValue(tableSort, sortKey);
+
+  const sorting = useMemo(
+    () => searchValueToSorting(sortValue, defaultSorting),
+    [defaultSorting, sortValue],
+  );
+
+  const sortingUpdater: OnChangeFn<SortingState> = useCallback(
```
felipebergamin: just to follow the convention I'd name it `onUpdateSorting` or `handleSortingUpdate`

dashboard/src/types/general.ts
```diff
@@ -140,6 +140,17 @@ export const zListingSize = z
   .optional(z.number().min(1).catch(DEFAULT_LISTING_ITEMS))
   .default(DEFAULT_LISTING_ITEMS);
 
+/** Sentinel for an explicit unsorted state when the table default is not unsorted. */
+export const TABLE_SORT_UNSORTED = 'none';
```
felipebergamin:
```suggestion
export const TABLE_SORT_UNSORTED = 'none' as const;
```

alanpeixinho (approved):
Nice work

felipebergamin (approved):

PR #2030, opened 2026-07-28
feat: unify hierarchical test table grouping

mentonin:
## Summary

- Replace the nested individual-tests subtable with a single TanStack hierarchy of group and leaf rows, so expanded tests share one header and column layout.
- Omit single-item path groups (hoisting path segments) and add Expand all / Collapse all / Disable grouping controls.
- Keep the table virtualized and use O(1) leaf-index lookups for log-sheet navigation.

## Comparison screenshots

taken from the tests tab in `/tree/stable/linux-5.10.y/065a677fad98698de04279ba2cb152a472ab8b1f`.
Changes to observe:

- Grouping behavior
- Grouped columns' data
  - repeats leaf if unique
  - shows muted unique count if not
  - shows range for dates
- Leafs (actual tests) shown as rows instead of nested tables
  - particularly on information density, clutter and clearness
- Grouping buttons on the top right of the table
- Sorting and searching are not shown on screenshots, but I suggest running this branch locally and feeling how it behaves - they are quite significant changes

<details>
    <summary> Collapsed (production) </summary>
    Screenshot From 2026-08-04 18-31-21
</details>

<details>
    <summary> Collapsed (this pr) </summary>
    Screenshot From 2026-08-04 18-35-20
</details>

<details>
    <summary> Expanded (production) </summary>
    Screenshot From 2026-08-04 18-33-27
</details>

<details>
    <summary> Expanded (this pr) </summary>
    Screenshot From 2026-08-04 18-35-31
</details>

<details>
    <summary> Ungrouped (this PR only) </summary>
    Screenshot From 2026-08-04 18-36-12
</details>

## Test plan

- [ ] On Tree / Hardware / Issue / Build test tabs, expand and collapse path groups; confirm leaf columns (status, date, duration, details) render correctly.
- [ ] Use Expand all / Collapse all / Disable grouping / Enable grouping; confirm flat mode shows full paths and grouping restores.
- [ ] Open a leaf log sheet and walk next/previous across leaves spanning multiple groups.
- [ ] Smoke-test a large result set (expand all or ungrouped) for scroll performance and sticky header behavior.
- [ ] Run `collapseTestsTree` / unified-tree unit tests.

## Related issues

- #2003 
- Scrolling expanded tables got jumpy because of #2008
- #1951 et al.
- Closes #2028


dashboard/src/components/Table/TableGroupingControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onModeChange: (mode: TableGroupingMode) => void;
+  onExpandAll: () => void;
+  onCollapseAll: () => void;
+}
+
+type GroupingButton = {
+  key: string;
+  labelId: MessagesKey;
+  onClick: () => void;
+  isSelected: boolean;
```
alanpeixinho: this is always false. Do we need it?

dashboard/src/components/TestsTable/buildUnifiedTestsTree.ts
```diff
@@ -0,0 +1,114 @@
...
+  };
+}
+
+/**
+ * Converts a collapsed path tree into unified group/leaf rows.
+ *
+ * Groups wrapping a single item are omitted: a node with one leaf child
+ * (after recursion) is hoisted to the parent with the path segment prepended.
+ */
+export function buildUnifiedTestsTree(nodes: TPathTests[]): UnifiedTestRow[] {
```
alanpeixinho, line 63: I have not profiled this properly, but the operation of grouping and ungrouping seems to be a little bit slow for huge tables.
Maybe we could just use a toggle/expand all only?

dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
```diff
@@ -47,43 +43,130 @@ const PathCell = ({
...
+    return null;
+  }
+  return value as ReactNode;
+}
+
+/**
+ * Adapts leaf/individual-test column defs for the unified group+leaf table:
+ * path/status get group-aware cells; other columns are blank on group rows.
+ */
+export function adaptColumnsForUnifiedTable(
```
alanpeixinho, line 310: I dont know if we need a full adapter for a table used in a single place.
Also, if we go for the adapter route, we might need to remove the columns property.

dashboard/src/components/Table/TableGroupingControls.tsx
```diff
@@ -0,0 +1,98 @@
...
+ */
+export function TableGroupingControls({
+  mode,
+  onModeChange,
+  onExpandAll,
+  onCollapseAll,
+}: ITableGroupingControls): JSX.Element {
+  const isGrouped = mode === 'grouped';
+
+  const buttons = useMemo((): GroupingButton[] => {
```
alanpeixinho: nit: Don't know if we need memoization here. We are just creating a simple array of objects here.

dashboard/src/components/TestsTable/buildUnifiedTestsTree.ts
mentonin: Fair point. I chose to pre-compute both states instead, as the ungrouped state is needed for ungrouped sorting

dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
mentonin: I tried to uncouple it from TestsTable specifically to lay the groundwork for grouping in builds/boots (#2003), and maybe elsewhere too. I removed the leaked coupling here on the new commit

dashboard/src/components/TestsTable/TestsTable.tsx
```diff
@@ -94,29 +137,40 @@ export function TestsTable({
...
 
+  const groupedData = useMemo(
+    () => buildUnifiedTestsTree(filteredTree),
+    [filteredTree],
+  );
+  const flatData = useMemo(
+    () => flattenTestsToLeafRows(filteredTree),
+    [filteredTree],
+  );
+  const data = groupingMode === 'ungrouped' ? flatData : groupedData;
```
felipebergamin, line 156: do we need to keep both on them in memory?
if so I think we can at least save an useMemo since they have the same deps


```suggestion
  const [groupedData, flatData]  = useMemo(
    () => [
      buildUnifiedTestsTree(filteredTree),
      flattenTestsToLeafRows(filteredTree),
    ],
    [filteredTree],
  );
  const data = groupingMode === 'ungrouped' ? flatData : groupedData;
```

PR #2036, opened 2026-07-30
feat: add Table column visibility, order, and resize options

### LINKED ISSUE #420: Table columns should have fixed/consistent size
The columns of any table are currently flexible, meaning that depending on the content of the rows, the columns may have varying sizes. This is especially noticeable when looking at tables with pagination (such as in the Builds Tab of Tree Details page, or the main page, Tree Listing). Having a fixed length for columns would prevent drastic changes in the UI.

mentonin:
## Summary

- Partial fix for #2008: column visibility, reorder, and resize - no persistence of any kind.
- Fixes #420 content-independent widths (no jump on sort/pagination)
- Shared `useLayoutTable` entry point drives Tree, Hardware, Issues, Builds, Boots, and hardware tree picker tables.
- Interior resize handles redistribute width between neighbors (table width fixed); outer edges can grow/shrink the table. Manual resize ignores `maxWidth`.
- Status column minWidths sized for Build (3-digit), Boot (4-digit), and Test (5-digit) counts.
- Tests / IndividualTests tables intentionally left out (pending #2030 grouping rework). Persistence and share/export URL are follow-ups.

## Test plan

- [ ] Tree / Hardware / Issues listing: Columns menu toggles and reorders; row-header columns 
stay leftmost and visible
- [ ] Drag interior handles: neighboring columns trade width; table width unchanged
- [ ] Drag outer (first left / last right) handles: table width changes; horizontal scroll when needed
- [ ] Reset marker clears manual width for that column
- [ ] Sort / paginate: column widths do not jump with content
- [ ] Build/Boot/Test status badges fit 3/4/5-digit counts at default mins without clipping

## Related Issues

- Part of #2008
- Closes #420
- Avoids touching tables reworked by #2030


dashboard/src/components/BootsTable/BootsTable.tsx
felipebergamin, line 1: Both BootsTable.tsx and BuildsTable.tsx have the details column, which is the only column that allows navigation to the details page. Clicking on the other cells only opens the right drawer.

I'm wondering if it's okay to allow the user to hide the only way they have to navigate.

felipebergamin (commented):
Nice one.

I'm not sure about how it is looking on large screens.

Comparing your branch to prod, the table is now narrower, leaving extra whitespace around it.

Image

Image

dashboard/src/components/BootsTable/BootsTable.tsx
alanpeixinho: I believe it might be safer to allow only for between cells resize, but leave the outter table size untouched.

MarceloRobert: > I'm wondering if it's okay to allow the user to hide the only way they have to navigate.

The user can also navigate by opening the right drawer and clicking on "show more details" at the botton of the drawer

dashboard/src/components/Table/ColumnResizeHandle.tsx
```diff
@@ -0,0 +1,121 @@
...
+ */
+export const ColumnResizeHandle = ({
+  edge,
+  leftColumn,
+  rightColumn,
+  onSizingChange,
+}: ColumnResizeHandleProps): JSX.Element => {
+  const intl = useIntl();
+
+  const onMouseDown = useCallback(
```
alanpeixinho, line 47: Does not seem to be working on phone touch screen.

dashboard/src/hooks/useLayoutTable.tsx
```diff
@@ -0,0 +1,294 @@
...
+          }
+          onColumnSizesChange={onColumnSizesChange}
+        >
+          {header.isPlaceholder
+            ? null
+            : flexRender(header.column.columnDef.header, header.getContext())}
+        </LayoutTableHead>
+      );
+    });
+    // sorting / extraHeaderDeps: force header re-render when table UI state
```
alanpeixinho, line 269: do we really need memoization here?
We should have only a few columns, it might be safe to just recompute.

PR #2042, opened 2026-08-03
fix: add timeouts to all production requests calls

tales-aparecida:
Without explicit timeouts, any requests.post() or requests.get() call can block indefinitely when a remote server hangs. This caused the ingester on db.kernelci.org to stall for 14 hours on a hung HTTP upload, leaving 543k submission files unprocessed.

Fixes: https://github.com/kernelci/dashboard/issues/2041

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/backend/kernelCI_app/constants/general.py
+++ b/backend/kernelCI_app/constants/general.py
@@ -11,3 +11,8 @@ SCHEMA_VERSION_ENV_FILE = "kernelCI/envconfig/schema-version.yaml"
 
 PRODUCTION_HOST = "https://dashboard.kernelci.org"
 STAGING_HOST = "https://staging.dashboard.kernelci.org"
+
+# Timeouts in seconds
+REQUESTS_TIMEOUT_UPLOAD_IN_SECONDS = 30
+REQUESTS_TIMEOUT_WEBHOOK_IN_SECONDS = 10
+REQUESTS_TIMEOUT_FETCH_IN_SECONDS = 30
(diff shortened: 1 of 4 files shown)


backend/kernelCI_app/constants/general.py
```diff
@@ -11,3 +11,7 @@
 
 PRODUCTION_HOST = "https://dashboard.kernelci.org"
 STAGING_HOST = "https://staging.dashboard.kernelci.org"
+
+REQUESTS_TIMEOUT_UPLOAD = 30
+REQUESTS_TIMEOUT_WEBHOOK = 10
+REQUESTS_TIMEOUT_FETCH = 30
```
felipebergamin: nit: I think it is worthy to make explicit the time unit used with a comment or the variable name to avoid confusion 

```suggestion
# Timeouts in seconds
REQUESTS_TIMEOUT_UPLOAD_IN_SECONDS = 30
REQUESTS_TIMEOUT_WEBHOOK_IN_SECONDS = 10
REQUESTS_TIMEOUT_FETCH_IN_SECONDS = 30
```

felipebergamin (approved):

mentonin (commented):
We need to handle exception `requests.ConnectTimeout` - probably by logging, maybe also retrying a limited number of times

Also, we might want to follow the advice [on `requests` docs](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts): pass `timeout=(connection_timeout, read_timeout)` and make `connection_timeout` slightly more than a multiple of 3

PR #2043, opened 2026-08-03
fix(ingester): break main loop when all workers exit unexpectedly

tales-aparecida:
The main loop waited for the queue to drain (while not
process_queue.empty()), which blocks forever when a worker hangs and
never consumes its poison pill. Now the loop checks whether any worker
is still alive after each progress report and breaks with an error log
if all workers have exited while items remain in the queue.

After joining, non-zero exit codes are logged and a Prometheus counter
(kcidb_ingester_worker_failures) is incremented with reason="exception"
or reason="signal". The reason label is more actionable for alerting
(e.g. rate(worker_failures{reason="signal"}) > 0 catches OOM kills)
while the exact exit code in the log line provides detail for
debugging.

Also adds docs/ingester.md documenting the parallel ingestion
architecture, worker-queue protocol, and error handling.

Fixes: https://github.com/kernelci/dashboard/issues/2041

Depends on https://github.com/kernelci/dashboard/pull/2042

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>
Signed-off-by: Tales da Aparecida <tales.aparecida@redhat.com>

### DIFF
--- a/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
+++ b/backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
@@ -71,6 +71,11 @@ TESTS_COUNTER = Counter(
 INCIDENTS_COUNTER = Counter(
     "kcidb_incidents", "Number of incidents ingested", ["ingester", "origin"]
 )
+WORKER_FAILURES_COUNTER = Counter(
+    "kcidb_ingester_worker_failures",
+    "Number of ingester worker processes that exited abnormally",
+    ["ingester", "reason"],
+)
 
 
 def standardize_tree_names(
@@ -569,10 +574,25 @@ def ingest_submissions_parallel(  # noqa: C901 - orchestrator with IO + multipro
                     process_queue.qsize(),
                 )
                 last_progress = time.time()
+            if not any(w.is_alive() for w in writers):
+                if not process_queue.empty():
+                    logger.error("All workers exited while queue still has items")
+                break
             time.sleep(1)
 
         for writer in writers:
             writer.join()
+            if writer.exitcode:
+                reason = "signal" if writer.exitcode < 0 else "exception"
+                logger.error(
+                    "Worker %s exited with code %s (%s)",
+                    writer.pid,
+                    writer.exitcode,
+                    reason,
+                )
+                WORKER_FAILURES_COUNTER.labels(
+                    ingester=INGESTER_GRAFANA_LABEL, reason=reason
+                ).inc()
     except KeyboardInterrupt:
         out("\nKeyboardInterrupt: terminating workers...")
         for writer in writers:
--- a/backend/kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py
+++ b/backend/kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py
@@ -2,7 +2,10 @@ from unittest.mock import MagicMock, call, mock_open, patch
 
 import pytest
 
-from kernelCI_app.constants.ingester import AUTOMATIC_LAB_FIELD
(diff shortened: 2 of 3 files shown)


backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -569,10 +574,24 @@ def ingest_submissions_parallel(  # noqa: C901 - orchestrator with IO + multipro
                     process_queue.qsize(),
                 )
                 last_progress = time.time()
+            if not any(w.is_alive() for w in writers):
```
mentonin, line 577: This will incorrectly log an error if workers finish between the loop check and this line

backend/kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py
```diff
@@ -870,3 +870,76 @@ def test_ingest_submissions_parallel_success(
...
+    def test_ingest_submissions_parallel_worker_failure(
+        self,
+        mock_getsize,
+        mock_value,
+        mock_queue_cls,
+        mock_process,
+        mock_logger,
+        mock_failures_counter,
+    ):
+        """Test that partial worker failure breaks the loop and logs errors."""
```
mentonin: This tests for complete worker failure, not partial. Actual partial worker failure tests would be nice to have

backend/kernelCI_app/tests/unitTests/commands/monitorSubmissions/kcidbng_ingester_test.py
```diff
@@ -870,3 +870,76 @@ def test_ingest_submissions_parallel_success(
...
+
+        mock_logger.error.assert_any_call(
+            "Worker %s exited with code %s (%s)", 1001, 1, "exception"
+        )
+        mock_logger.error.assert_any_call(
+            "Worker %s exited with code %s (%s)", 1002, -9, "signal"
+        )
+
+        # Prometheus counter incremented for both failures
+        assert mock_failures_counter.labels.call_count == 2
```
mentonin: This checks that every failure is being labeled. We should also check that failures are being labeled correctly

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
```diff
@@ -569,10 +574,24 @@ def ingest_submissions_parallel(  # noqa: C901 - orchestrator with IO + multipro
...
+                logger.error("All workers exited while queue still has items")
+                break
             time.sleep(1)
 
         for writer in writers:
             writer.join()
```
mentonin, line 584: We might want (in a separate PR) to use `writer.join(timeout=...)` for more defensive programming related to hung ingester

mentonin (commented):
commit message says alive checks run "after each progress report", but they run every loop instead

Also, commit message claims this fixes #2041 and deals with hanging, but this fix does neither - it deals with all workers dying (not hanging) before the queue is dealt with, and hanging workers still block forever (#2041 only fixes workers hanging on http requests). We can go ahead with the merge, but live workers hanging might come back as an issue later

backend/kernelCI_app/management/commands/helpers/kcidbng_ingester.py
tales-aparecida: join() only runs after the while loop exits, which now happens in two cases: (1) the queue drained (workers consumed their poison pills and died gracefully), or (2) queue still has items, but all workers are already dead (the is_alive guard broke the loop). I believe dead workers can't block join(), so I think a timeout would be a no-op. But I can be wrong

mentonin (approved):

felipebergamin (approved):

PR #2050, opened 2026-08-11
feat(issue-details): per-tree incidents and first good checkout

alanpeixinho:
## What it is

Per-tree incidents on the issue details page (`/issue/$issueId`): first seen, last seen, and first good checkout (earliest later checkout on that tree with no incident for the issue).

Backend adds `group_by=tree` on seen queries and a single LATERAL first-good-checkout query. Wired only on `/api/issue/{id}` so extras/listing stay unchanged.

Closes #1957

## How to test

1. Open an issue details page with incidents on multiple trees (e.g. one that previously showed only global first/last).
2. Confirm **Incidents per Tree** appears before the global first/last sections.
3. Expand a tree: check first/last links and dates; if recovered, **First Good Checkout**; if not, **Still failing**.
4. Spot-check an issue still failing on a tree and one that recovered.
5. Confirm issue listing / extras endpoints still work and do not gain the expensive per-tree work.

per-tree-incidents

### DIFF
--- a/backend/kernelCI_app/queries/issues.py
+++ b/backend/kernelCI_app/queries/issues.py
@@ -240,57 +240,104 @@ def get_test_issues(*, test_id: str) -> list[dict]:
 
 
 def get_issue_seen_data(
-    *, issue_id_list: list[str], mode: Literal["first", "last"] = "first"
+    *,
+    issue_id_list: list[str],
+    mode: Literal["first", "last"] = "first",
+    group_by: Literal["issue", "tree"] = "issue",
 ) -> list[dict]:
     """
     Retrieves the incident and checkout data of either the first or last
     incident of a list of issues through a list of `issue_id`s.
 
     :param mode: Either 'first' to get oldest incidents or 'last' to get the newest ones.
+    :param group_by: 'issue' returns one row per issue; 'tree' returns one row per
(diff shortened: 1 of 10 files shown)


backend/kernelCI_app/queries/issues.py
```diff
@@ -240,57 +240,104 @@ def get_test_issues(*, test_id: str) -> list[dict]:
...
-                incidents IC
-            LEFT JOIN tests T ON IC.test_id = T.id
-            LEFT JOIN builds B ON (
-                IC.build_id = B.id
-                OR T.build_id = B.id
-            )
-            LEFT JOIN checkouts C ON B.checkout_id = C.id
-            JOIN target_incident TI ON IC.id = TI.id
-        """
+            """
```
felipebergamin, line 305: ranking by issue_version before _timestamp, so “last seen” is the highest/lowest version, not the latest failure. Is that right?

alanpeixinho: I am keeping the behavior of the global "last seen" here, prioritizing latest version of the issue, and then using timestamp.
For what I see issue versioning is mostly used here as a way to make immutable data updates.

felipebergamin (approved):

PR #2053, opened 2026-08-14
fix(tests-table): stabilize columns and improve path/lab display

### LINKED ISSUE #420: Table columns should have fixed/consistent size

alanpeixinho:
## What it is

Closes #420 — tests table columns jumping/overlapping on sort, pagination, and scroll.

- `table-fixed` + path width from longest path; horizontal scroll when table is wider than viewport
- 12px tree indent
- Group lab summaries like hardware: one name when uniform, `(n)` when mixed; missing → `Unknown`
- Fix group cell rendering (`row.getValue` override) and linked-cell truncation (`min-w-0`)

Scoped to tests table (+ hardware tests path column). Other listing tables unchanged.

## How to test

1. Open **Tests** tab (Tree Details or Hardware Details) with enough rows to sort and scroll.
2. Sort columns and scroll vertically — column widths stay stable, no overlap.
3. Expand/collapse grouped rows — 12px indent per level.
4. Grouped **Lab** — single lab shows name; multiple shows `(n)`; leaf rows show name or `Unknown`.

### DIFF
--- a/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
+++ b/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
@@ -30,8 +30,36 @@ import { UNKNOWN_STRING } from '@/utils/constants/backend';
 import { getDateSortKey } from './groupSummaries';
 import type { GroupFieldSummary, UnifiedTestRow } from './types';
 
-const INDENT_WIDTH = 20;
+const INDENT_WIDTH = 12;
+export const PATH_TREE_INDENT = INDENT_WIDTH;
 const MUTED_VALUE_CLASS = 'text-gray-500';
+
+export const PathWithPrefixEllipsis = ({
+  value,
+}: {
+  value: string;
+}): JSX.Element => (
+  <div
+    className="w-full min-w-0 overflow-hidden text-ellipsis whitespace-nowrap"
+    dir="rtl"
+    style={{ textAlign: 'left' }}
+  >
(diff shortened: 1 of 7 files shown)


dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
```diff
@@ -45,32 +73,30 @@ const PathCell = ({
...
-        <TooltipTrigger>
-          {isLeaf ? (
-            <div
-              className="max-w-80 overflow-hidden text-nowrap text-ellipsis"
-              style={{ direction: 'rtl', textAlign: 'left' }}
-            >
-              <span style={{ direction: 'ltr', unicodeBidi: 'embed' }}>
+        <TooltipTrigger asChild>
+          <div className="min-w-0 flex-1 overflow-hidden">
+            {isLeaf ? (
```
felipebergamin, line 96: I'm not sure, but I think this <div> won't be focusable via keyboard. This might be annoying for our users.

alanpeixinho: Good point, I had not considered keyboard focus here.

alanpeixinho:
Closed as these changes were introduced on #2055

PR #2058, opened 2026-08-19
fix: bust persisted query cache on deploy

### LINKED ISSUE #2048: Listing page crashes with "Cannot read properties of undefined (reading 'map')"
## Description
On `/tree`, the app can render the route error boundary ("Something went wrong! / Show Error") with the error `Cannot read properties of undefined (reading 'map')`.

The error originates in `TopBar` → `OriginSelect` (`dashboard/src/components/TopBar/TopBar.tsx`). `pageOrigins` is derived from the `origins` query and only guards `!originData`, not the case where `originData` exists without `checkout_origins`/`test_origins`:

```ts
const pageOrigins = useMemo(() => {
  if (!originData) return [];
  return isHardwarePath ? originData.test_origins : originData.checkout_origins;
}, [originData, isHardwarePath]);
// ...
pageOrigins.map(...)
```

The `origins` query (`useOrigins`) uses `staleTime: 2h` and is persisted via `PersistQueryClientProvider`. When the persisted cache holds an `origins` entry whose shape lacks those fields, `pageOrigins` is `undefined` and `.map` throws.

## Reproduction
Dev server uses `sessionStorage` (prod uses `localStorage`).

1. Open `/tree`.
2. In the browser console, seed an `origins` cache entry with a mismatched shape and reload:

```js
sessionStorage.setItem('REACT_QUERY_OFFLINE_CACHE', JSON.stringify({
  buster: '', timestamp: Date.now(),
  clientState: { mutations: [], queries: [{
    queryKey: ['origins'], queryHash: '["origins"]',
    state: { data: ['maestro'], dataUpdatedAt: Date.now(), dataUpdateCount: 1,
      error: null, errorUpdateCount: 0, errorUpdatedAt: 0, fetchFailureCount: 0,
      fetchFailureReason: null, fetchMeta: null, isInvalidated: false,
      status: 'success', fetchStatus: 'idle' } }] },
}));
location.reload();
```

3. `/tree` shows "Something went wrong! / Show Error"; expanding shows `Cannot read properties of undefined (reading 'map')`.

Recover: `sessionStorage.clear()` (or `localStorage.clear()`) then reload.

alanpeixinho:
## What it is
Fixes `/tree` crashing with `Cannot read properties of undefined (reading 'map')` when a persisted React Query cache holds an `origins` payload without `checkout_origins` / `test_origins` (#2048).
- `OriginSelect` treats missing origin lists as `[]` so `.map` cannot throw.
- Persist cache is versioned with `VITE_CACHE_BUSTER` (dashboard version, or `build-<timestamp>` if empty) so a new deploy drops stale blobs.
## How to test
Dev persist uses `sessionStorage` (prod: `localStorage`). Recover with `sessionStorage.clear()` then reload.
1. Open `/tree` and confirm the origin dropdown works.
2. Seed a mismatched origins shape and reload:
```js
sessionStorage.setItem('REACT_QUERY_OFFLINE_CACHE', JSON.stringify({
buster: '', timestamp: Date.now(),
clientState: { mutations: [], queries: [{
  queryKey: ['origins'], queryHash: '["origins"]',
  state: { data: ['maestro'], dataUpdatedAt: Date.now(), status: 'success', fetchStatus: 'idle' },
}]},
}));
location.reload();
```
Expect: no “Something went wrong!”; origins refetch; dropdown populated.
3. Reload without editing the cache: same `buster` → hydrates, no crash.

Closes #2048

### DIFF
--- a/dashboard/src/components/TopBar/TopBar.tsx
+++ b/dashboard/src/components/TopBar/TopBar.tsx
@@ -40,9 +40,11 @@ const OriginSelect = ({
     if (!originData) {
       return [];
     }
-    return isHardwarePath
-      ? originData.test_origins
-      : originData.checkout_origins;
+    return (
+      (isHardwarePath
+        ? originData.test_origins
+        : originData.checkout_origins) ?? []
+    );
   }, [originData, isHardwarePath]);
 
   const selectItems = useMemo(() => {
--- a/dashboard/src/main.tsx
+++ b/dashboard/src/main.tsx
@@ -90,7 +90,10 @@ ReactDOM.createRoot(document.getElementById('root')!).render(
   <React.StrictMode>
     <PersistQueryClientProvider
       client={queryClient}
-      persistOptions={{ persister }}
+      persistOptions={{
+        persister,
+        buster: import.meta.env.VITE_DASHBOARD_VERSION,
+      }}
     >
       <IntlProvider
         messages={currentMessages}
--- a/dashboard/vite.config.ts
+++ b/dashboard/vite.config.ts
@@ -19,14 +19,23 @@ const dashboardVersion =
   })();
 
 // https://vitejs.dev/config/
-export default defineConfig({
-  plugins: [tailwindcss(), tsconfigPaths(), react(), TanStackRouterVite()],
-  define: {
-    'import.meta.env.VITE_DASHBOARD_VERSION': JSON.stringify(dashboardVersion),
-  },
-  resolve: {
-    alias: {
-      '@': path.resolve(__dirname, './src'),
+export default defineConfig(({ command }) => {
+  if (command === 'build' && !dashboardVersion) {
+    throw new Error(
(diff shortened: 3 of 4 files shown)


alanpeixinho:
Leaving here so it is documented, we had some discussion on calls about solving this by using local memory for cache or sessionStorage. However there are a few drawbacks there.
* local memory: clears cache at every refresh of the page, being somewhat aggressive when invalidating cache.
* session storage: refresh would keep the cache, but it would not solve the bug. If a deploy was to change any response schema, the user would still face the same error, requiring them to close and reopen the page on the browser. It means even using sessionStorage, busting stale cache would still be needed.
Considering this I believe it would be better to keep using localStorage (as it is right now), and just include the stale cache bust.
cc: @felipebergamin

felipebergamin:
I made a comment about the sessionStorage then I saw your comment about it. Ok

felipebergamin (approved):

PR #2064, opened 2026-08-21
perf/hardware summary perf

alanpeixinho:
### What
  Faster hardware details `/summary` for the default 5-day window. Same response.
  The tests-branch `OR` on platform identity is split into a `UNION` so each arm can use its index. The builds branch uses `array_position` instead of `@>` so Postgres does not rebuild the compatible GIN
  bitmap per build. Tree heads are probed with a `LATERAL` join on `builds.build_id`. The unfiltered summary query is skipped when the filtered SQL already matches head commits. Frontend timestamps are coerced
  to numbers so a URL rewrite does not refetch.
### How to test
  Confirm behavior is unchanged:
  1. Open a hardware details page (e.g. `bcm2711-rpi-4-b`, maestro, last 5 days) with no filters. Summary counts, trees, compatibles, and filter options match `main`.
  2. Apply status / duration and other filter. Counts follow the filter


backend/kernelCI_app/queries/hardware.py
felipebergamin, line 1: this is probably intentional, but just checking. The builds are still joining tests, so they are not really being optimized like boots/tests, right?

felipebergamin, line 1: nit: For this kind of change, it would be nice to have a before-and-after comparison to highlight the improvement. Just a nit though — no need to do it now.

alanpeixinho: Good catch. I am including this on builds as well.

alanpeixinho: I agree. I am including a small local benchmark on the PR.

alanpeixinho:
Local DB: **9.6M tests**, **44k builds**, **48k checkouts**.

Window: **2026-08-21 .. 2026-08-26** (5 days). Timed with `EXPLAIN ANALYZE` on 2026-08-28.

Example platforms: **`bcm2711-rpi-4-b` / maestro** (timed below; ~52k platform tests in window). Also checked **`imx8mp-evk`**, **`qemu-arm64`**, **`kubernetes` / maestro**, **`qemu-arm64` / linaro_pull_labs** — same counts before/after.

## Changed queries

| Query | Before | After |
|---|---:|---:|
| `get_hardware_trees_head_commits` (LATERAL probe on `builds.build_id`) | 9.3 s | 1.7 s |
| `get_hardware_details_summary` (`platform_tests` CTE + both arms) | 3.8 s | 1.9 s |

Both on `bcm2711-rpi-4-b` / maestro, no duration filters, all tree heads.

cc: @felipebergamin

felipebergamin (approved):

PR #2037, opened 2026-07-31
[WIP] feat(hardware): add daily hardware aggregates recomputed from raw

alanpeixinho:
### What it is                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  Adds daily hardware aggregate tables (hardware_daily_builds, hardware_daily_tests) built from raw checkouts / builds / tests.                                                                                      
                                                                                                                                                                                                                     
  * recompute_hardware_daily: idempotent per-day rebuild (DELETE + INSERT). Uses temp tables so the raw join runs once per day. Advisory locks avoid concurrent collisions. If raw is pruned but aggregates          
    exist, keeps the stale rows.                                                                                                                                                                                     
  * Cron: hourly --days=1 (today), every 6h --days=7 — no 60-day batch.                                                                                                                                              
  * prune_db: can prune aggregates by checkout_day independently of raw.                                                                                                                                             
  * update_db / seed_test_data: snapshot/restore and seed support for the new tables.                                                                                                                                
  * hardware_status: unchanged — still written and read as before.                                                                                          

  Closes #2034.   

### How to test                                                                                                                                                                                                                                                                                                                                                                                                                        

#### migrate                                                                                                                                                                                                          
  Run local database and apply migration                                                                                                                                                                                                                                                                                                  

#### manual recompute                                                                                                                                                                                                 
  `python manage.py recompute_hardware_daily --day=2025-07-18`                                                                                                                                                         
  `python manage.py recompute_hardware_daily --days=7`                                                                                                                                                                 

#### verify rows                                                                                                                                                                                                      
  Check that the aggregate counts for builds, boots and tests match the sum of the raw tables (builds and tests).

#### prune dry-run (aggregates only)
  `python manage.py prune_db --older-than "30 days" \
    --tables hardware_daily_builds,hardware_daily_tests --dry-run`

PR #2070, opened 2026-08-25
feat/hardware improved filters

alanpeixinho:
Depends on: #2037

### What it is

Replaces the hardware listing's single origin selector with five independent filters, querying the daily aggregate tables from #2037.

#### Backend
  • /api/hardware/ reads hardware_daily_* with per-side filters: checkoutOrigin, buildOrigin, testOrigin, buildLab, testLab (comma-separated)
  • Build and test counters filtered independently; a narrowed side decides which platforms appear
  • New GET /api/hardware/filters/ — independent option lists, no cascading
  • By-revision listing uses same daily path; deprecated origin aliases checkout + build origin

#### Frontend
  • Filter drawer on hardware listing; TopBar origin hidden there
  • URL via diffFilter; defaults: checkout + build origin = maestro

Part of #1989.

Closes #2035.

### What to test

  Prerequisite: #2037 merged and aggregates computed (recompute_hardware_daily --days=15)

  • /api/hardware/: maestro defaults, each filter independently, empty param clears constraint, deprecated origin, commitsList
  • /api/hardware/filters/: five lists, no cascading; pull-labs and virtual labs present
  • By-revision + selectors work with side filters
  • UI: drawer, URL params, unchecking origin shows all origins, revision selector unchanged


dashboard/src/api/hardware.ts
```diff
@@ -1,168 +1,154 @@
...
+      paramKey,
+      selectedFilterValues(diffFilter[sectionKey]),
+    ]),
+  );
 
-type HardwareListingByRevisionApiResponse = {
-  hardware: HardwareListingByRevisionApiItem[];
-};
+export const buildOriginForSelectors = (diffFilter: TFilter): string =>
+  selectedFilterValues(diffFilter.buildOrigin) || DEFAULT_ORIGIN;
```
felipebergamin: Unchecking every build origin makes the table query all origins and the tree/revision dropdown stay on maestro. The listing and the selector are then different datasets. Use the same empty-means-all rule as the selectors API (or disable the selector until an origin is chosen).

felipebergamin (commented):
Took a proper read through this. The query side is nice, and the tests around the independent build/test filters gave me a good idea of what to expect. A few notes below — the details page origin is the only one I'd really want sorted before this goes in.

dashboard/src/api/hardware.ts
alanpeixinho: Nice catch, the frontend was incorrectly feeding the revision selector. Fixed it. We now have the same behavior as the table.

PR #2072, opened 2026-08-27
fix(tests-table): keyboard access for path tooltips and links

### LINKED ISSUE #2071: Tests table: path tooltips and row links are not keyboard-accessible
On the tests table, keyboard users cannot reliably reach or see focus for some interactive elements:
1. Path tooltips — Path cells use non-focusable div elements as tooltip triggers. Collapsed group paths and leaf paths are skipped when tabbing, so tooltips are only usable with a mouse.
2. Row links — Table cells use overflow-hidden, which clips the default focus ring on full-cell links. Focus is hard to see when navigating with Tab.

alanpeixinho:
### What it is

Fixes keyboard accessibility in the tests table path column and linked table cells.

* Replaces non-focusable div tooltip triggers with button elements so Tab reaches collapsed groups and leaf paths
* Insets cell link focus rings so they stay visible inside overflow-hidden table cells

### How to test

1. Open a page with the tests table (tree details or hardware details).
2. Tab through the table and confirm path tooltip triggers receive focus on both group and leaf rows.
3. Confirm tooltips can be opened from the keyboard on focused path cells.
4. Tab to row links and confirm the focus ring is visible and not clipped by the cell.
5. Confirm mouse behavior is unchanged (row expand/collapse, link navigation, tooltips).

Closes #2071

### DIFF
--- a/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
+++ b/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
@@ -88,7 +88,6 @@ const PathCell = ({
       {!isExpandable && row.depth > 0 && <span className="mr-2 w-4 shrink-0" />}
       <Tooltip>
         <TooltipTrigger asChild>
-          {/* Activation bubbles to the row's onClick, which toggles groups. */}
           <button
             type="button"
             className="min-w-0 flex-1 overflow-hidden text-left"
--- a/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
+++ b/dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
@@ -53,9 +53,9 @@ export const PathWithPrefixEllipsis = ({
 export const PathWithTooltip = ({ value }: { value: string }): JSX.Element => (
   <Tooltip>
     <TooltipTrigger asChild>
-      <div className="min-w-0 overflow-hidden">
+      <button type="button" className="min-w-0 overflow-hidden text-left">
         <PathWithPrefixEllipsis value={value} />
-      </div>
+      </button>
     </TooltipTrigger>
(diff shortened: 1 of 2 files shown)


dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
```diff
@@ -88,15 +88,19 @@ const PathCell = ({
       {!isExpandable && row.depth > 0 && <span className="mr-2 w-4 shrink-0" />}
       <Tooltip>
         <TooltipTrigger asChild>
-          <div className="min-w-0 flex-1 overflow-hidden">
+          {/* Activation bubbles to the row's onClick, which toggles groups. */}
```
felipebergamin: I don't think we need to keep the comments

dashboard/src/components/ui/table.tsx
```diff
@@ -114,7 +114,15 @@ const TableCellWithLink = React.forwardRef<
     className={cn('align-middle [&:has([role=checkbox])]:pr-0', className)}
     {...props}
   >
-    <Link className={cn('flex min-w-0 flex-1 p-4', linkClassName)} {...linkProps}>
+    <Link
+      // The link fills the cell, so an outset focus ring is clipped by cells
+      // that hide overflow.
```
felipebergamin: ditto

dashboard/src/components/TestsTable/DefaultTestsColumns.tsx
alanpeixinho: Done

dashboard/src/components/ui/table.tsx
alanpeixinho: Done

felipebergamin (approved):

